PR #1, opened 2016-02-17
Initial commit

justinfagnani:


justinfagnani:
@rictic for Typescript stuff

src/commands/command.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export interface Command {
+  name: string;
+  description: string;
+  args: ArgDescriptor[];
+  run(options: {[name: string]: string}): Promise<any>;
+}
+
+// TODO(justinfagnani): how do I re-export an interface?
+export interface ArgDescriptor extends ArgDescriptor {
```
garlicnation: Why do you need to re-export this?

samccone:
oh hey @justinfagnani is there a design doc for this tool? I would love to see the goals here

src/commands/command.ts
justinfagnani: hmm... it's not anymore. I was pulling it in with `Command` into polytool and help. I might again.

justinfagnani:
@samccone shared a doc w/ you.

samccone:
a quick note, as much as I like typescript and writing in it for personal projects, if we go down this path we are basically excluding a large number of possible contributors (same argument as coffeescript stuff)

justinfagnani:
This is way different than Coffeescript. Coffeescript is an entirely different syntax, with little tangible benefit, and that's having trouble keeping up with JavaScript changes. Typescript is a relatively small delta from JS, but with a huge benefit from type checking, and polyfilling features like async/await, and they've shown a big willingness to change to be more compatible with JavaScript. Many more projects are moving to Typescript than I have seen move to Coffeescript. We're moving all of our node tools to TS. </:coffee:-:scroll:-rant>

samccone:
sure, just a thought and something we will need to defend because people will ask :)

robdodson:
I can haz doc too?

rictic:
Another big advantage for typescript over similar stuff like coffeescript is that the produced javascript is just the original code without the types. So it's super easy to do JS <-> TS interaction

samccone:
types are good, typescript is good, just like other options, I only wanted to point out that writing in another language for a tool targeting the JS community has tradeoffs, something to be aware of and something to be prepared for. 

+1 for whatever is the most productive and helps us ship something great

samccone:
+1 to landing this :shipit:

justinfagnani:
@garlicnation ping

src/polytool.ts
```diff
@@ -0,0 +1,48 @@
...
+  commands : Map<String, Command> = new Map();
+  cli : commandLineCommands.CLI;
+
+  constructor() {
+    this.addCommand(new HelpCommand(this.commands));
+    this.addCommand(new ServeCommand());
+  }
+
+  addCommand(command) {
+    this.commands.set(command.name, command);
```
garlicnation: What is the intended behavior if a user calls addCommand with the same name twice?

src/polytool.ts
```diff
@@ -0,0 +1,48 @@
...
+import {Command} from './commands/command';
+
+export class Polytool {
+
+  commandDescriptors = [];
+  commands : Map<String, Command> = new Map();
+  cli : commandLineCommands.CLI;
+
+  constructor() {
+    this.addCommand(new HelpCommand(this.commands));
```
garlicnation: Nit: I'd rather see this in bin/, or a static method that provides a polytool configured with a default set of commands. It feel better to me for programmatic consumers to be able to add commands piecemeal.

garlicnation:
LGTM with a couple of nits.

addyosmani:
Can we please share the doc with the rest of the team? Only found out about this repo today.. :(

src/polytool.ts
justinfagnani: It'll overwrite. It's not a user though, it's us. Let's not do that :)

src/polytool.ts
justinfagnani: Future work? Not targeting programatic consumers at the moment.

PR #2, opened 2016-02-18
Add test command

justinfagnani:

PR #8, opened 2016-02-23
Add test command

justinfagnani:
Fixes #2

### DIFF
--- a/src/commands/test.ts
+++ b/src/commands/test.ts
@@ -1,0 +1,134 @@
+import {Command} from './command';
+import * as wct from 'web-component-tester';
+// import * as process from 'process';
+
+export class TestCommand implements Command {
+  name = 'test';
+
+  description = 'Runs web-component-tester';
+
+  args = [
+    {
+      name: 'persistent',
+      alias: 'p',
+      description: 'Keep browsers active (refresh to rerun tests)',
+      type: Boolean,
+    },
+    {
+      name: 'root',
+      description: 'The root directory to serve tests from',
+      type: String,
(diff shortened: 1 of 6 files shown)


src/commands/test.ts
```diff
@@ -0,0 +1,134 @@
...
+      type: String,
+    },
+    {
+      name: 'visibility',
+      description: "Set job visibility to 'public', 'public restricted', 'share', 'team' or 'private'",
+      type: String,
+    },
+  ];
+
+  run(options): Promise<any> {
```
samccone: this will either be an Error or null so we could do like Error|null right?

justinfagnani: The error type is not part of Promise's type parameters. I can change `any` to `void` though.

src/commands/test.ts
```diff
@@ -0,0 +1,134 @@
+import {Command} from './command';
+import * as wct from 'web-component-tester';
+// import * as process from 'process';
+
+export class TestCommand implements Command {
+  name = 'test';
+
+  description = 'Runs web-component-tester';
+
+  args = [
```
garlicnation: Is there any way to get this from wct, or are we only going to try that for the newer batch of command-line-arg based tools?

garlicnation:
LGTM

src/commands/test.ts
justinfagnani: Yeah, this is somewhat temporary, and obviously not very maintainable. I either want to standardize on command-line-args style declarations, or add some mapping functions.

PR #14, opened 2016-04-14
Initial build pipeline

justinfagnani:
This adds a generic build pipeline to polytool.

 * Builds can be extended by the project gulpfile by exporting a `transforms` property.
 * HTML files are split similar to crisper, but style should also be split (WIP), and they are recombined by default. script and style tags are not moved like in cripser.
 * Vulcanize reads files from the gulp stream unlike default gulp-vulcanize. This means that transformed files are passed through vulcanize. I'm uncertain how well that actually works with normal gulp-vulcanize, as it might depend on the ordering and completeness of other transform steps and whether they've written to disk yet.
* entrypoints and source globs can be specified, but there's only support for a single entrypoint right now. This needs to be expanded to multiple entrypoints, and sharding.

Future work:
  * ES6 module support: transform `<script type=module>` to an HTML import that loads an HTML file with non-module script tag.
  * polystyle support: auto-wrap external CSS in a dom-module
  * service worker / manifest type things
  * optionally run packages through the pipeline. Right now there's an escape hatch in the loader that vulcanize uses to read from disk, but piping packages through the compiler might be preferable, especially with npm and packages that have esnext fields.

The code is a bit of a mess, but I wanted to get this up for interested parties.

@robdodson @addyosmani @garlicnation

### DIFF
--- a/src/build/html-project.ts
+++ b/src/build/html-project.ts
@@ -1,0 +1,236 @@
+/**
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+import * as dom5 from 'dom5';
+import * as path from 'path';
+import {Transform} from 'stream';
+import File = require('vinyl');
+
+const pred = dom5.predicates;
+
+const extensionsForType = {
+  'text/ecmascript-6': 'js',
+  'application/javascript': 'js',
+  'text/javascript': 'js',
+  'application/x-typescript': 'ts',
+  'text/x-typescript': 'ts',
+};
+
+/**
+ * Splits and rejoins inline scripts and styles from HTML files.
+ *
+ * Use `HtmlProject.prototype.split` and `HtmlProject.prototype.rejoin` to
+ * surround processing steps that operate on the extracted resources.
+ * HtmlProject works well with gulp-if to process files based on filename.
+ */
+export class HtmlProject {
+
+  _splitFiles : Map<string, SplitFile> = new Map();
+  _parts : Map<string, SplitFile> = new Map();
+  split = new Splitter(this);
+  rejoin = new Rejoiner(this);
+
+  isSplitFile(parentPath: string): boolean {
+    return this._splitFiles.has(parentPath);
+  }
+
+  getSplitFile(parentPath: string): SplitFile {
+    let splitFile = this._splitFiles.get(parentPath);
+    if (!splitFile) {
+      splitFile = new SplitFile(parentPath);
+      this._splitFiles.set(parentPath, splitFile);
+    }
+    return splitFile;
+  }
+
+  addSplitPath(parentPath: string, childPath: string): void {
+    let splitFile = this.getSplitFile(parentPath);
+    splitFile.addPartPath(childPath);
+    this._parts.set(childPath, splitFile);
+  }
+
+  getParentFile(childPath: string): SplitFile {
+    return this._parts.get(childPath);
+  }
+
+}
+
+
+/**
+ * Represents a file that is split into multiple files.
+ */
+class SplitFile {
+  path : string;
+  parts : Map<string, string> = new Map();
+  outstandingPartCount = 0;
+  vinylFile : File = null;
+
+  constructor(path: string) {
+    this.path = path;
+  }
+
+  addPartPath(path: string): void {
+    this.parts.set(path, null);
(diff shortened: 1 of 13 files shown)


abdonrd:
@justinfagnani this will kill `polybuild` package? :)

justinfagnani:
Eventually yes, but not before `polytool build` is a complete replacement. We have too many command line tools as it is, consolidation will be good.

abdonrd:
Makes sense! Thanks @justinfagnani! :)

justinfagnani:
@garlicnation this is in better shape to review now

src/build/stream-resolver.ts
```diff
@@ -0,0 +1,148 @@
...
+
+export interface Config {
+  entrypoint?: string;
+  /**
+   * Hostname to match for absolute urls.
+   * Matches "/" by default
+   */
+  host?: string;
+  /**
+   * Prefix directory for components in url. Defaults to "/".
```
garlicnation: I'd like to see a default config instance that StreamResolver references, since it's really StreamResolver that has the defaults, not the Config interface.

src/build/stream-resolver.ts
```diff
@@ -0,0 +1,148 @@
...
+    } else {
+      deferred.resolve(content);
+    }
+  });
+}
+
+/**
+ * Returns true if `patha` is a sibling or aunt of `pathb`.
+ */
+function isSiblingOrAunt(patha:string, pathb:string) {
```
garlicnation: These should ideally get out into a shared library.

src/build/html-project.ts
```diff
@@ -0,0 +1,232 @@
...
+  );
+
+  _project : HtmlProject;
+
+  constructor(project) {
+    super({objectMode: true});
+    this._project = project;
+  }
+
+  _transform(file, encoding, callback) {
```
garlicnation: What type is callback?

src/build/html-project.ts
```diff
@@ -0,0 +1,232 @@
...
+        let newFile = new File({
+          cwd: file.cwd,
+          base: file.base,
+          path: file.path,
+          contents: new Buffer(splitContents),
+        });
+        callback(null, newFile);
+      } catch (e) {
+        console.log('error', e);
+        throw e;
```
garlicnation: Should this be: callback(e, null)

src/build/html-project.ts
```diff
@@ -0,0 +1,232 @@
...
+      }
+    } else {
+      callback(null, file);
+    }
+  }
+}
+
+class Rejoiner extends Transform {
```
garlicnation: Docs

src/build/html-project.ts
```diff
@@ -0,0 +1,232 @@
...
+ * HtmlProject works well with gulp-if to process files based on filename.
+ */
+export class HtmlProject {
+
+  _splitFiles : Map<string, SplitFile> = new Map();
+  _parts : Map<string, SplitFile> = new Map();
+  split = new Splitter(this);
+  rejoin = new Rejoiner(this);
+
+  isSplitFile(parentPath) {
```
garlicnation: Explicit types?

garlicnation:
Mostly looks good - I'd mainly like to see the implicitly define type become explicit .

src/build/stream-resolver.ts
justinfagnani: A default `Config` instance won't help in the cases where a config is provided, but with no `host`. `StreamResolver` could merge config objects, but I'm not sure how much benefit that it. Unless I'm missing what you mean here.

src/build/stream-resolver.ts
justinfagnani: Yep... I still have a FileLoader library under development.

src/build/html-project.ts
justinfagnani: It's just `Function` in the declaration for `Transformer`. Added.

src/build/html-project.ts
justinfagnani: done

src/build/html-project.ts
justinfagnani: done

src/build/html-project.ts
justinfagnani: done

justinfagnani:
PTAL

robdodson:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

src/build/html-project.ts
```diff
@@ -0,0 +1,237 @@
...
+        let scriptTags = dom5.queryAll(doc, Splitter.isInlineScript);
+        let styleTags = dom5.queryAll(doc, pred.hasTagName('style'));
+
+        let scripts = [];
+        let styles = [];
+
+        for (let i = 0; i < scriptTags.length; i++) {
+          let scriptTag = scriptTags[i];
+          let source = dom5.getTextContent(scriptTag);
+          // let typeAtribute = dom5.getAttribute(scriptTag, 'type');
```
garlicnation: remove commented code

garlicnation:
LGTM w/ nit

src/build/html-project.ts
justinfagnani: done

PR #20, opened 2016-04-24
Use flags defined in polyserve

justinfagnani:


garlicnation:
LGTM.

Does polyserve still work with `require`?

justinfagnani:
Yes, typescript is compiling to commonjs

PR #24, opened 2016-04-27
Fix deps after running depcheck

justinfagnani:


justinfagnani:
submitting TBR (to be reviewed)

PR #30, opened 2016-04-27
add basic testing to check dependencies

### LINKED ISSUE #25: Add depcheck to tests
To avoid publishing without proper dependencies installing.

FredKSchott:
Fixes #25
/cc @justinfagnani

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ node_js:
   - "5"
 sudo: false
 before_script:
-  - npm install -g typescript typings
   - npm install
   - npm run init
-script:
   - npm run build
+script:
+  - npm run test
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -3,8 +3,8 @@ node_js:
   - "5"
 sudo: false
 before_script:
-  - npm install -g typescript typings
```
FredKSchott: no longer necessary after 577ce994ed1499b1a0a8590382271824e128847d

src/commands/init.ts
```diff
@@ -23,7 +24,7 @@ export class InitCommand implements Command {
   run(options): Promise<any> {
     return new Promise((resolve, reject) => {
       let env = new YeomanEnvironment();
-      env.register(require.resolve('generator-polymer-init'), 'polymer-init:app');
+      env.registerStub(polymerGenerator, 'polymer-init:app');
```
FredKSchott: `require.resolve()` was giving us trouble with `depcheck`, since it technically was never required in our application. Instead of asking Yeoman to load the generator for us, we can require it ourselves.

justinfagnani:
LGTM!

PR #36, opened 2016-04-29
New polytool init template

justinfagnani:


src/commands/init.ts
```diff
@@ -19,13 +22,22 @@ export class InitCommand implements Command {
...
+    },
+  ];
 
   run(options): Promise<any> {
     return new Promise((resolve, reject) => {
+      let templateName = options['name'] || 'polymer-init';
       let env = new YeomanEnvironment();
       env.registerStub(polymerGenerator, 'polymer-init:app');
-      env.run('polymer-init', {}, () => resolve());
+      env.registerStub(PolykartGenerator, 'polykart:app');
```
FredKSchott: You mentioned that the line above just broke for you, rebase and use the `env.register()` method instead.

src/templates/polykart.ts
```diff
@@ -0,0 +1,118 @@
...
+        console.error(error);
+        done();
+      });
+  }
+
+  _getGitHubToken() {
+    try {
+      return fs.readFileSync('token', 'utf8').trim();
+    } catch (e) {
+      console.error(`
```
FredKSchott: nit: do you want a newline as the first char in your console log?

src/templates/polykart.ts
```diff
@@ -0,0 +1,118 @@
...
+You need to create a github token and place it in a file named 'token'.
+The token only needs the 'public repos' permission.
+Generate a token here: https://github.com/settings/tokens
+This restriction will be removed soon!
```
FredKSchott: This is actually temporary right? Otherwise an env variable would a better choice here.

src/templates/polykart.ts
```diff
@@ -0,0 +1,118 @@
...
+        .pipe(tar.extract(process.cwd(), {
+          ignore: (_, header) => {
+            let splitPath = header.name.split(path.sep);
+            // ignore the top directory in the tarfile to unpack directly to
+            // the cwd
+            return splitPath.length < 2 || splitPath[1] === '';
+          },
+          map: (header) => {
+            let unprefixed = header.name.split(path.sep).slice(1).join(path.sep).trim();
+            // console.log('unpacking', unprefixed);
```
FredKSchott: safe to remove?

FredKSchott:
Comments added, overall looks good except for the register vs. registerStub issue. And `polykart.ts` could use some cleanup.

src/templates/polykart.ts
justinfagnani: done

src/templates/polykart.ts
justinfagnani: It's really, truly temporary

src/templates/polykart.ts
justinfagnani: temporary, do not care

src/commands/init.ts
justinfagnani: registerStub works here because the generator is part of this package

justinfagnani:
Thanks Fred! Added a test... going to merge now.

PR #37, opened 2016-05-02
Fix init command using the wrong template path.

justinfagnani:


justinfagnani:
@FredKSchott I had to roll back this line to get init to work again. I'm not sure how registerStub is supposed to work...

FredKSchott:
arg! I'll look into rolling this back in a way that doesn't break depchecks.

FredKSchott:
@justinfagnani fixed!

PR #38, opened 2016-05-03
Fix handling of absolute URLs in StreamResolver

justinfagnani:


package.json
```diff
@@ -16,6 +16,7 @@
   "author": "The Polymer Project Authors",
   "license": "BSD-3-Clause",
   "dependencies": {
+    "chai": "^3.5.0",
```
azakus: This should be a devDependency right?

justinfagnani: looks like you already fixed

justinfagnani:
@azakus added a test. PTAL

azakus:
LGTM

PR #45, opened 2016-05-05
Factor out Github release download into its own file

azakus:


azakus:
@justinfagnani PTAL

src/templates/polykart.ts
```diff
@@ -101,7 +54,7 @@ export function getGenerator(options?) {
         return githubToken;
```
garlicnation: I'd like you to move this function to github and use it as a default value for Github()

azakus:
LGTM from @garlicnation

PR #46, opened 2016-05-05
Add `build --env reyserve` and `serve --env reyserve`

garlicnation:


abdonrd:
Why double quotes?

src/polytool.ts
```diff
@@ -17,6 +17,21 @@ import {ServeCommand} from './commands/serve';
 import {TestCommand} from './commands/test';
 import {Command} from './commands/command';
 
+import {ReyServe} from './environments/reyserve';
+
+const globalArguments = [{
+  name: 'env',
+  type: function(value) {
```
azakus: Make this return a type of Target

src/target/target.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+import {ServerOptions} from 'polyserve/lib/start_server';
+
+export interface TargetImpl {
+  serve?(options: ServerOptions): Promise<any>;
+}
+
+export interface Target {
```
azakus: Name seems generic

src/commands/serve.ts
```diff
@@ -28,6 +28,15 @@ export class ServeCommand implements Command {
       browser: options.browser,
       componentDir: options['component-dir'],
       packageName: options['package-name'],
+    };
+
+    if (options.env) {
```
azakus: Drop a comment that this is for supporting "unimplemented" backends.

src/environments/reyserve.ts
```diff
@@ -0,0 +1,50 @@
...
+import {ServerOptions} from 'polyserve/lib/start_server';
+import {Target, TargetImpl} from '../target/target';
+import * as temp from 'temp';
+import * as fs from 'fs-extra';
+import * as path from 'path';
+import * as child_process from 'child_process';
+
+temp.track();
+
+const REYSERVE_SOURCES = '/Users/ajo/workspace/reyserve';
```
azakus: Use the `Github` class :)

src/commands/serve.ts
```diff
@@ -28,6 +28,15 @@ export class ServeCommand implements Command {
...
+      console.log(options.env.impl);
+      if (!options.env.impl.serve) {
+        throw new Error(`Serve unimplemented in environment ${options.env.name}`);
```
justinfagnani: do we want to require that all environments implement all hooks?

src/environments/reyserve.ts
```diff
@@ -0,0 +1,50 @@
...
+const REYSERVE_FILES = ['app.yaml', 'reyserve.py', 'http2push.py'];
+
+export class ReyServe implements TargetImpl {
```
justinfagnani: why does this implement `TargetImpl`. `TargetImpl` sounds like a class...

src/environments/reyserve.ts
```diff
@@ -0,0 +1,50 @@
...
+  serve(options: ServerOptions): Promise<any> {
+    return new Promise((resolve, reject) => {
+    // Make temp dir
```
justinfagnani: can't we reuse the build dir, rather than make a temp dir/

justinfagnani: to expand: I think most of the work here should be in `ReyServe.build`, and serve should make sure build has been called (maybe we call it for every serve for now), and then just call dev_appserver.py

src/target/target.ts
justinfagnani: Let's go with Environment

azakus:
LGTM

src/commands/serve.ts
garlicnation: I'd rather each environment be able to implement as few hooks as possible and work with the default build/project/etc whenever possible.

garlicnation:
@abdonrd TSLint wanted it. I've reverted them.

src/commands/init.ts
```diff
@@ -13,8 +13,7 @@ import {ArgDescriptor} from 'command-line-args';
 
 import {PolykartGenerator} from '../templates/polykart';
 
-// not ES modules compatible
-const YeomanEnvironment = require('yeoman-environment');
+import *  as YeomanEnvironment from 'yeoman-environment';
```
garlicnation: @FredKSchott Does this look OK to you?

src/commands/serve.ts
justinfagnani: that's why I commented... here you're requiring the environment to implement `serve()`

src/commands/init.ts
justinfagnani: this should be fine to satisfy depcheck. To tell, run `npm run test`

src/commands/serve.ts
```diff
@@ -28,6 +29,12 @@ export class ServeCommand implements Command {
       browser: options.browser,
       componentDir: options['component-dir'],
       packageName: options['package-name'],
+    };
+
+    const env: Environment = options.env;
+
+    if (env) {
```
justinfagnani: maybe `if (env && env.serve)` ?

src/environment/environment.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+import {ServerOptions} from 'polyserve/lib/start_server';
+
+export class HookNotImplemented extends Error {
+  constructor(name: string, hook: string){
+    super(`Hook "${hook}" not implemented in environment "${name}"`);
+  }
+}
+
+/** This only exists for to throw errors. */
```
justinfagnani: "for to"? :)

src/environment/environment.ts
```diff
@@ -0,0 +1,32 @@
...
+export abstract class AbstractEnvironment implements Environment {
+  build(options: ServerOptions): Promise<any> {
+    throw new HookNotImplemented(this.constructor.name, 'build');
```
justinfagnani: like the other comments... not sure we need to require implementations

src/environments/environments.ts
```diff
@@ -0,0 +1,37 @@
...
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+import {Environment} from '../environment/environment';
+
+import {ReyServe} from './reyserve';
+
+interface EnvironmentTable {
```
justinfagnani: this could just be `Map<string, Environment>`. I'd prefer Maps rather than objects for string-keyed lookups.

src/environments/environments.ts
```diff
@@ -0,0 +1,37 @@
...
+  [name: string]: {new(): Environment};
+}
+
+const Environments: EnvironmentTable =([ReyServe]).reduce(
```
justinfagnani: `reduce` might be a little overkill here. IMO, it's easier to see what's going on with a for..of loop, or simply:

```
environments.set('reyserve', ReyServe);
```

justinfagnani: lower-case `environments`. No need to yell that something is const :)

src/environments/environments.ts
```diff
@@ -0,0 +1,37 @@
...
+      throw new Error(`Two environments have the name: ${name}`);
+    }
+    environments[name] = environment;
+    return environments;
+  },
+  <EnvironmentTable>{}
+);
+
+/**
+  * Builds an environment with the given name.
```
justinfagnani: indented one space too far

src/environments/environments.ts
```diff
@@ -0,0 +1,37 @@
...
+  */
+export function buildEnvironment(name: string):Environment {
```
justinfagnani: style: need space after last `:`

src/environments/environments.ts
justinfagnani: style: space after `=`

src/environments/reyserve.ts
```diff
@@ -0,0 +1,60 @@
...
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+import {ServerOptions} from 'polyserve/lib/start_server';
+import {AbstractEnvironment} from '../environment/environment';
```
justinfagnani: style: put project internal imports after external imports separated by a newline. Sort by import name

src/environments/reyserve.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+export class ReyServe extends AbstractEnvironment {
+  build() {
+
+  }
+  serve(options: ServerOptions): Promise<any> {
+    return new Promise((resolve, reject) => {
+    // Make temp dir
+      const servingDir = temp.mkdirSync('reyserve');
```
justinfagnani: I still think we don't need the temp dir. I'd move most of this into `build()`

src/build/sw-precache.ts
```diff
@@ -131,6 +131,9 @@ export class SWPreCacheTransform extends Transform {
         options.configFileName
       );
     }
+    this.on('end', () => {
+      console.log("precacheEnded");
```
justinfagnani: remove?

src/commands/build.ts
```diff
@@ -44,16 +44,25 @@ export class BuildCommand implements Command {
       defaultValue: 'sw-precache-config.js',
       description: 'Path to an sw-precache configuration to be ' +
         'used for service worker generation.'
+    },
+    {
+      name: 'root',
+      description: 'The root directory to find sources and place build. ' +
+          'Defaults to $CWD'
```
justinfagnani: "$CWD" -> "current directory"

src/environments/environments.ts
```diff
@@ -0,0 +1,26 @@
...
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+import {Environment} from '../environment/environment';
+
+import {ReyServe} from './reyserve';
+
+class EnvironmentMap extends Map<String, {new(): Environment}> {
```
justinfagnani: Why a class? This could just be `Map<string, EnvironmentConstructor>`

justinfagnani: `String` should be lowercased

src/environments/environments.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+const environments = new EnvironmentMap()
```
justinfagnani: I'm not sure this warrants it's own file. It'll be easy to see what's going on right in polytool.ts

src/environments/reyserve.ts
```diff
@@ -0,0 +1,60 @@
...
+import * as fs from 'fs-extra';
+import * as path from 'path';
+import * as child_process from 'child_process';
+
+temp.track();
+
+const REYSERVE_FILES = ['app.yaml', 'reyserve.py', 'http2push.py'];
+
+export class ReyServe implements Environment {
+  build(opts: BuildOptions): Promise<any> {
```
justinfagnani: style nit: add some newlines around methods

src/environments/reyserve.ts
```diff
@@ -0,0 +1,60 @@
...
+    const bundled = path.join(process.cwd(), 'build/bundled');
+    const unbundled = path.join(process.cwd(), 'build/unbundled');
+    const reyserve = temp.mkdirSync('reyserve');
```
justinfagnani: naming: `reyserve` -> `reyserveFiles`?

src/environments/reyserve.ts
```diff
@@ -0,0 +1,60 @@
...
+      // TODO(garlicnation): One server serves bundled and unbundled.
+      REYSERVE_FILES.forEach((file) => {
+        fs.copySync(path.join(reyserve, file), path.join(bundled, file));
+        fs.copySync(path.join(reyserve, file), path.join(unbundled, file));
+      });
+    });
+  }
+  serve(options: ServerOptions): Promise<any> {
+    return new Promise((resolve, reject) => {
+      console.log('Spawning dev server');
```
justinfagnani: 'Launching reyserve dev server'?

justinfagnani:
LGTM

Would love to see a test for build and serve deferring to environments soon.

src/environments/reyserve.ts
garlicnation: Done

src/environments/reyserve.ts
garlicnation: Done

justinfagnani:
@garlicnation is this ready to go in?

PR #49, opened 2016-05-06
Generate service worker from config in repo

azakus:


azakus:
@justinfagnani PTAL

src/build/build.ts
```diff
@@ -95,23 +95,25 @@ export function build(options?: BuildOptions): Promise<any> {
...
+      depsResolve = resolve;
+    });
+
     let bundledPhase = new ForkedVinylStream(allFiles)
       .pipe(bundler.bundle)
+      .on('end', () => {
+        let depsSet = bundler.streamResolver.requestedUrls;
+        depsResolve(Array.from(depsSet));
+      })
+      .pipe(new SWPreCacheTransform({root, main, buildRoot: 'build/bundled', deps: depsPromise, serviceWorkerName}))
```
justinfagnani: long line

src/build/sw-precache.ts
```diff
@@ -10,40 +10,161 @@
...
+   */
+  buildRoot: string;
+  /**
+   * Name of the output service worker file.
+   */
+  serviceWorkerName: string;
+  /**
+   * Promise that returns the list of files to be cached by the service worker.
+   *
+   * If not given, all files streamed to SWPreCacheTransform will be put into the precache list, except for the file matching `configFileName`.
```
justinfagnani: long line

justinfagnani: I'm not sure about that default behavior. The way we source files here that would be _everything_ including demos, tests, READMEs, etc. Since we won't not specify the files, would it be simpler to drop this?

src/build/sw-precache.ts
```diff
@@ -10,40 +10,161 @@
...
+    }
+    if (options.configFileName) {
+      this.fullConfigFilePath = path.resolve(
+        options.root,
+        options.configFileName
+      );
+    }
+  }
+
+  _transform(file: File, encoding: string, callback:(error?, data?) => void): void {
```
justinfagnani: style: ass space after `callback:`

src/build/sw-precache.ts
```diff
@@ -10,40 +10,161 @@
...
+    if (file.path === this.fullConfigFilePath) {
+      this.swConfig = JSON.parse(file.contents.toString());
+      callback(null, null);
```
justinfagnani: we write almost every other file to the output, even if it might not be that useful, should we just write the config also?

src/build/sw-precache.ts
```diff
@@ -10,40 +10,161 @@
...
+    .then((config) => {
+      let file = new File({
+        path: path.resolve(
+          this.options.buildRoot,
+          this.options.serviceWorkerName
+        ),
+        contents: new Buffer(config)
+      });
+      this.push(file);
+      super.end();
```
justinfagnani: is it cool to defer `end()` like this? I suspect it is, but deferring a super call is a pretty tricky thing to do generally...

justinfagnani:
LGTM, modulo tests and maybe not needed the cache everything fallback behavior. Let's get it submitted though, so we can rebase and continue.

PR #50, opened 2016-05-06
fix build by providing options to vulcanize correctly

azakus:


azakus:
PTAL @garlicnation, @justinfagnani

src/build/optimize.ts
```diff
@@ -26,31 +26,30 @@ export interface OptimizeOptions {
...
-    removeComments?: boolean;
+    [property: string]: any;
   };
   /**
    * Enable CSS minification
    */
-  css?: {
-    stripWhitespace?: boolean;
+  css?: boolean | {
+    [property: string]: any;
```
justinfagnani: Why the union type? DOesn't look like you're using the object variant.

justinfagnani:
LGTM

src/build/optimize.ts
azakus: The properties in there aren't even used, so I'm not 100% sure what the interface should be anymore. That should probably not be in _this_ PR though.

PR #51, opened 2016-05-07
Fix busted streams

azakus:
StreamLogger should call the callback

ForkedVinylStream should push `null` instead of emitting end on source
`end` event, as that may miss chunks sitting in the internal buffer.
Pushing `null` will let the buffer play out until it hits the null, when
the `end` event will emit itself.


azakus:
PTAL @justinfagnani

PR #54, opened 2016-05-09
Add flag for service worker precache config file

azakus:
Clean up deps pipeline by adding precache config file after optimization
pass with `passthrough` mode for vinyl-fs.src

### DIFF
--- a/src/commands/build.ts
+++ b/src/commands/build.ts
@@ -24,6 +24,7 @@ export class BuildCommand implements Command {
       name: 'main',
       description: 'The main HTML file',
       defaultOption: true,
+      defaultValue: 'index.html'
     },
     {
       name: 'shell',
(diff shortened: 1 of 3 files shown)


src/commands/build.ts
```diff
@@ -35,7 +36,14 @@ export class BuildCommand implements Command {
     },
     {
       name: 'sources',
+      multiple: true,
       description: 'The sources file to build',
+    },
+    {
+      name: 'sw-precache-config',
+      defaultValue: 'sw-precache-config.js',
+      description: 'Path to an service worker precache configuration to be ' +
```
addyosmani: s/service worker/sw-precache (given that this is specific to the tool rather than service workers in general)

addyosmani:
LGTM minus one small comment

PR #61, opened 2016-05-09
Removes unused minimatch dependency, fixes broken tests

FredKSchott:
We're now using `minimatch-all`


FredKSchott:
@justinfagnani updated to remove actual import calls

justinfagnani:
LGTM

justinfagnani:
The minimatch typing was still needed by it's dependents. `typings` does not install transitive ambient dependencies due to problems in definitely typed.

PR #65, opened 2016-05-09
Support --help & --version flags

FredKSchott:
Previously, both `--help/-h` & `--version/-v` flags would throw errors if they were used. To implement these across commands, the global `polymer-cli` now watches for and handles them.

CR @justinfagnani

### DIFF
--- a/src/polymer-cli.ts
+++ b/src/polymer-cli.ts
@@ -42,6 +42,21 @@ export class PolymerCli {
   }
 
   run(args) {
+    // If the "--version"/"-v" flag is ever present, just print
+    // the current version. Useful for globally installed CLIs.
+    if (args.indexOf('--version') > -1 || args.indexOf('-v') > -1) {
+      console.log(require('../package.json').version);
+      return;
+    }
+
+    // Alias the "--help"/"-h" flag to run the help command for the given
+    // command. This gets us around some limitations in our arguments
+    // parsers and keeps us from having to implement additional help logic
+    // inside of each command.
+    if (args.indexOf('--help') > -1 || args.indexOf('-h') > -1) {
+      args = ['help', args[0]];
+    }
+
     this.cli = commandLineCommands(this.commandDescriptors);
     let cliCommand = this.cli.parse(args);
     let command = this.commands.get(cliCommand.name || 'help');
--- a/test/commands/cli_test.js
+++ b/test/commands/cli_test.js
@@ -1,0 +1,72 @@
+/**
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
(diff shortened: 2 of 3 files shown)


src/polymer-cli.ts
```diff
@@ -42,6 +43,21 @@ export class PolymerCli {
   }
 
   run(args) {
+    // If the "--version"/"-v" flag is ever present, just print
+    // the current version. Useful for globally installed CLIs.
+    if (args.indexOf('--version') > -1 || args.indexOf('-v') > -1) {
+      console.log(packageJSON.version);
```
justinfagnani: might be nicer to require package.json here

src/polymer-cli.ts
```diff
@@ -42,6 +43,21 @@ export class PolymerCli {
   }
 
   run(args) {
+    // If the "--version"/"-v" flag is ever present, just print
```
justinfagnani: It's slightly unsatisfying to manually scan when we have the args parser, but not a huge deal. Maybe we could use a separate args definition and first pass for globals. No change needed now though.

test/commands/cli_test.js
```diff
@@ -0,0 +1,72 @@
...
+
+const packageJSON = require('../../package.json');
+const assert = require('chai').assert;
+const PolymerCli = require('../../lib/polymer-cli').PolymerCli;
+const sinon = require('sinon');
+
+suite('The general CLI', () => {
+
+  test('displays general help when no command is called', () => {
+    let polytool = new PolymerCli();
```
justinfagnani: `polytool` -> `cli` here and throughout

src/polymer-cli.ts
FredKSchott: I'm a big fan of keeping all of your `require` statements at the top of the file, it's a helpful pattern for when you need to hunt down missing/unused dependencies, circular dependencies in larger applications, etc.

src/polymer-cli.ts
FredKSchott: I know, me too. It's mostly to work around limitations in our arguments parser & architecture. If we didn't do this, we'd have to add `--help` support to each command individually (otherwise the parser would throw an error [a few lines below here](https://github.com/Polymer/polymer-cli/pull/65/files#diff-81f7c55176652a78582250c2c66fbd3aR62)). Then we'd need logic inside of each command to bounce us out and into the help command. 

In the end this seemed like the cleanest choice until we can reevaluate our args parser and/or need to move between commands.

test/commands/cli_test.js
FredKSchott: doh, good catch.

src/polymer-cli.ts
justinfagnani: Yes, but in this case it's really `require` abuse to load a file. It's not a library import, it's a data file, which can't have dependencies, is not listed in package.json, etc. It doesn't need to be loaded at startup at all.

test/commands/cli_test.js
FredKSchott: done!

justinfagnani:
FYI: https://github.com/75lb/command-line-commands/issues/3

src/polymer-cli.ts
FredKSchott: that's a fair point, especially in a CLI where startup time is such an important factor. Lets move it.

FredKSchott: done!

justinfagnani:
LGTM!

PR #70, opened 2016-05-10
For bundle build, only precache entrypoints and shared bundle

azakus:


justinfagnani:
I think this is it... we can merge and get feedback from use.

justinfagnani:
LGTM

Issue #63, opened 2016-05-09
Docs: might add heads-up in README to upgrade your `node` version if you're getting an error when you run


tjsavage:

justinfagnani:
Yes, we need to publish supported versions.

Our target is current LTS, so 4.4.4.

FredKSchott:
It's common to define which node.js versions you support in your
package.json.

FredKSchott:
@justinfagnani & @tjsavage we now throw a warning when installing on a <v4 version, that says polymer-cli needs node v4.0 or higher. Is this enough or do we still want something in the README?

tjsavage:
Perfect. Might just add a line in the README too just to potentially save people the step of Install > Error > Update

PR #73, opened 2016-05-10
Update package.json engines and travis version testing

FredKSchott:
Being explicit in both our package.json & travis testing is enough to handle #63 IMO. Now ff anyone tries to install the CLI with an earlier version of Node they'll get a warning at install time.


abdonrd:
LGTM!

PR #74, opened 2016-05-10
Bundle shared imports into the shell

### LINKED ISSUE #67: build: bundle shared resources with app shell
This removes a round trip to load shared-bundle.html

justinfagnani:
Fixes #67

### DIFF
--- a/src/build/bundle.ts
+++ b/src/build/bundle.ts
@@ -8,6 +8,7 @@
  * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
  */
 
+import * as dom5 from 'dom5';
 import * as gulpif from 'gulp-if';
 import * as path from 'path';
 import {Transform} from 'stream';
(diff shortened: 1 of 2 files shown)


justinfagnani:
@azakus PTAL

src/build/bundle.ts
```diff
@@ -176,6 +190,45 @@ export class Bundler {
     });
   }
 
+  _addSharedImportsToShell(bundles: Map<string, string[]>): string {
+    console.assert(this.shell != null);
+    let shellDeps = bundles.get(this.shell)
+        .slice()
```
azakus: `slice is unnecessary`

src/build/bundle.ts
```diff
@@ -176,6 +190,45 @@ export class Bundler {
...
+    let shellDepsSet = new Set(shellDeps);
+    let file = this.streamResolver._files.get(this.shell);
+    console.assert(file != null);
+    let contents = file.contents.toString();
+    let doc = dom5.parse(contents);
+    let imports = dom5.queryAll(doc, dom5.predicates.AND(
+      dom5.predicates.hasTagName('link'),
+      dom5.predicates.hasAttrValue('rel', 'import')
+    ));
+    for (let _import of imports) {
```
azakus: Comment on why you rip the imports out of the doc if they are in the shared deps bundle

src/build/bundle.ts
```diff
@@ -176,6 +190,45 @@ export class Bundler {
...
+      if (shellDepsSet.has(dom5.getAttribute(_import, 'href'))) {
+        dom5.remove(_import);
+      }
+    }
+    let head = dom5.query(doc, dom5.predicates.hasTagName('head'));
+    if (!head) {
```
azakus: parse5 always generates at least `<html><head></head><body></body></html>`, so this case is always false.

azakus:
LGTM

src/build/bundle.ts
justinfagnani: done

src/build/bundle.ts
justinfagnani: done

src/build/bundle.ts
justinfagnani: done

PR #76, opened 2016-05-10
Fix precaching to be consistent

### LINKED ISSUE #75: sw-precache does not include all entry points
It looks like this happens only when generating a fresh build and not when a previous build already exists.

azakus:
sw-precache reads directly from disk to generate precache manifests, so we
must wait for the streams to be complete before calling
generateServiceWorker.

Read the config file once outside of the stream, let sw-precache write to
disk

Clone config so that bundled and unbundled don't write into each other

Add tests for service worker generation

Also handle uncaught promises in build

Fixes #75


package.json
```diff
@@ -54,6 +55,7 @@
     "eslint": "^2.9.0",
     "gulp-mocha": "^2.2.0",
     "sinon": "^1.17.4",
+    "tmp": "0.0.28",
```
justinfagnani: what's the difference between this an `temp`? @garlicnation is adding `temp` in #46

azakus:
@justinfagnani PTAL

src/build/build.ts
```diff
@@ -45,8 +46,13 @@ process.on('uncaughtException', (err) => {
   console.error(err.stack);
 });
 
+process.on('unhandledRejection', (err, p) => {
+  console.log(`Promise rejection: ${err}`);
```
justinfagnani: maybe add `console.error(err.stack)`

src/build/build.ts
```diff
@@ -45,8 +46,13 @@ process.on('uncaughtException', (err) => {
   console.error(err.stack);
 });
 
+process.on('unhandledRejection', (err, p) => {
```
justinfagnani: I like longer names, like `error` and `promise`. We only have to type them once, then read them foreverrrr...

src/build/build.ts
```diff
@@ -108,50 +114,49 @@ export function build(options?: BuildOptions): Promise<any> {
...
+      let unbundledDeps = Array.from(bundler.streamResolver.requestedUrls);
+
+      let bundledDeps = Array.from(bundler.entrypointFiles.keys());
+      bundledDeps.push(bundler.sharedBundleUrl);
+
+      parsePreCacheConfig(swPrecacheConfig).then((swConfig) => {
+        let unbundledConfig = clone(swConfig);
+        let bundledConfig = clone(swConfig);
+        Promise.all([
+          generateServiceWorker({
```
justinfagnani: how about making a local wrapper on `generateServiceWorker` that takes care of the common params?

justinfagnani:
big LGTM

PR #82, opened 2016-05-11
Optimize command imports to only load when command is run

### LINKED ISSUE #79: CLI is slowwwwwww
Lets see if we can get startup time down at all.

```
$ polymer help
real	0m2.256s
user	0m1.612s
sys	0m0.217s
```
```
fschott-macbookpro:polytool-test fschott$ time polymer --version
real	0m2.215s
user	0m1.575s
sys	0m0.209s
```
(Note that these two commands are essentially no-ops)

FredKSchott:
Fixes #79

#### Old startup time:
```
$ time polymer --version
real	0m2.667s
user	0m1.625s
sys	0m0.299s
```

#### New startup time:
```
$ time polymer --version
real	0m0.671s
user	0m0.511s
sys	0m0.073s
```

Each dependency that has been optimized to load at run-time was previously adding at least 300ms a piece to the overall load time. `PolykartGenerator` was adding 1 second!. The remaining ~670ms time is mostly taken up by loading the dependencies of `lint` & `serve`, which cannot be easily optimized because their commands rely on the arguments defined within their dependencies. We'll need to revisit this pattern if we want to optimize further.

### DIFF
--- a/src/commands/build.ts
+++ b/src/commands/build.ts
@@ -12,7 +12,6 @@ import {ArgDescriptor} from 'command-line-args';
 import {CLI} from 'command-line-commands';
 
 import {Command} from './command';
-import {build} from '../build/build';
 
 export class BuildCommand implements Command {
   name = 'build';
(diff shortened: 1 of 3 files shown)


justinfagnani:
🎊

justinfagnani:
LGTM! please rebase :)

PR #90, opened 2016-05-11
only log broken sw-precache config if it exists

### LINKED ISSUE #89: Don't show error on no sw-precache-config.js
We should check for file existence, rather than try/catch.

azakus:
Fixes #89


azakus:
TBR @justinfagnani

PR #99, opened 2016-05-12
Add a default config

azakus:
All commands will share this config object for common information


azakus:
PTAL @justinfagnani

src/config.ts
```diff
@@ -0,0 +1,40 @@
...
+import * as path from 'path';
+
+export interface ConfigOptions {
+  root?: string;
+  main?: string;
+  shell?: string;
+  entrypoints?: string[];
+}
+
+export class Config {
```
justinfagnani: To be a little more clear at he use sites of this type, let's call it something like `ProjectConfig`.

justinfagnani:
LGTM

PR #103, opened 2016-05-12
Read from 'polymer.json' for project config

azakus:
Flags will override

### DIFF
--- a/src/project-config.ts
+++ b/src/project-config.ts
@@ -47,9 +47,7 @@ export class ProjectConfig {
       // config file is default, options will override
       let fromFile = ProjectConfig.fromConfigFile(configFile);
       if (fromFile) {
-        for (let option in options) {
-          fromFile[option] = options[option];
-        }
+        Object.assign(fromFile, options);
         options = fromFile;
       }
     }
--- a/src/project-config.ts
+++ b/src/project-config.ts
@@ -24,8 +24,35 @@ export class ProjectConfig {
   shell: string;
   entrypoints: string[];
 
-  constructor(options: ProjectConfigOptions) {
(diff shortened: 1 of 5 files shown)


azakus:
@justinfagnani PTAL

src/project-config.ts
```diff
@@ -24,8 +24,35 @@ export class ProjectConfig {
   shell: string;
   entrypoints: string[];
 
-  constructor(options: ProjectConfigOptions) {
+  static fromConfigFile(filepath: string): ProjectConfig {
+    try {
+      let configContent = fs.readFileSync(filepath, 'utf-8');
```
justinfagnani: is there really no way to check a file besides catching an exception? `statSync`?

src/project-config.ts
```diff
@@ -24,8 +24,35 @@ export class ProjectConfig {
...
+    this.init(configFile, options);
+  }
+
+  init(configFile?: string, options?: ProjectConfigOptions) {
     options = options || {};
+    if (configFile) {
+      // config file is default, options will override
+      let fromFile = ProjectConfig.fromConfigFile(configFile);
+      if (fromFile) {
+        for (let option in options) {
```
justinfagnani: `Object.keys(options)` so you don't get things from the proto chain.

justinfagnani: Actually, just use `Object.assign`

justinfagnani:
LGTM, with nits.

src/project-config.ts
azakus: Nope. `fs.exists` is super deprecated.

src/project-config.ts
azakus: I was worried about Node v4 support, but let's let Travis tell us.

PR #104, opened 2016-05-13
Add logging framework, add logging to main & init command

### LINKED ISSUE #95: Logging
For our CLI logging, here are the log levels & flags I’m planning to move forward with. Please comment with feedback, if any.

**Levels:**
`error` (error occurred, we're probably going to exit now)
`warn` (something odd or sub-optimal happened. We'll keep going but check this out)
`info` (normal output for the user)
`debug` (granular, for debugging purposes)

**CLI Behavior & Flags:**
default behavior: output error,warn,info
`--verbose`: default + debug
`--quiet`: error only

**Design:**
1. polymer-cli creates a logger for each command
2. passes that down and through the chosen command via:
  1. `build`: Passed in as `log` option
  1. `help`: command fully implemented in CLI, no passing required
  1. `init`: Passed in as `log` option. Additionally, the yeoman CLI uses the `debug` library for its verbose logging. We'll want to turn this on as well to log all relevant yeoman events as well (usually done via `DEBUG=yeoman*`env variable).
  1. `lint`: Passed in as `log` option
  1. `serve`: Passed in as `log` option
  1. `test`: Need to do some small work here to refactor WCT to be run as a library. It is currently being called through a method specifically designed for the command line, which means no passing of complex objects like loggers. Once this is done, it can be passed in as a `log` option.

*For all "Passed in as log option":* each library will take a logging instance as an option value. If none is provided, it will instantiate its own logger with the default behavior. Later, each package can be extended to support it's own `--verbose` & `--quiet` options.

FredKSchott:
Fixes #95

This PR adds basic logging to the Main CLI runner, and the init command. Once this is merged we can logging to each of the other commands as needed.

We've built a simple logger to be shared across polymer CLIs & libraries. See https://github.com/Polymer/plylog/blob/master/index.js for the logic there & please open an issue if you have any comments there.

/cc @justinfagnani @azakus

### DIFF
--- a/custom_typings/plylog.d.ts
+++ b/custom_typings/plylog.d.ts
@@ -1,0 +1,15 @@
+declare module 'plylog' {
+
+  class PolymerLogger {
+    constructor(options: any)
+    setLevel(newLevel: string)
+    error(message: string, metadata?: any)
+    warn(message: string, metadata?: any)
+    info(message: string, metadata?: any)
+    debug(message: string, metadata?: any)
+  }
+
+  export function setVerbose();
+  export function setQuiet();
+  export function getLogger(name: string): PolymerLogger;
+}
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -12,6 +12,7 @@ import {Command} from './command';
 import {ArgDescriptor} from 'command-line-args';
 import {PolykartGenerator} from '../templates/polykart';
 
+import * as logging from 'plylog';
 import *  as YeomanEnvironment from 'yeoman-environment';
 
 // NOTE(fschott) 05-02-2015: Yeoman needs to load our generator in a non-standard way via require.resolve.
@@ -30,7 +31,7 @@ export class InitCommand implements Command {
       description: 'The template name',
       type: String,
       defaultOption: true,
-    },
+    }
   ];
 
   run(options, config): Promise<any> {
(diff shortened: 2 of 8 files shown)


justinfagnani:
Needs a rebase

custom_typings/plylog.d.ts
```diff
@@ -0,0 +1,18 @@
+declare module 'plylog' {
+
+	class PolymerLogger {
```
justinfagnani: indent 2 spaces

src/commands/init.ts
```diff
@@ -31,6 +32,18 @@ export class InitCommand implements Command {
       type: String,
       defaultOption: true,
     },
+    {
+      name: 'verbose',
```
justinfagnani: the should be global flags

custom_typings/plylog.d.ts
FredKSchott: arg! sublime be crazy

src/commands/init.ts
FredKSchott: done!

FredKSchott:
Shoot, I didn't realize I killed tests by making these changes. Nothing actually broke with behavior, but the way that these tests were run is no longer valid.

@justinfagnani If this is blocking you tonight and you feel comfortable merging, feel free. Either way I'll be able to fix this first thing tomorrow morning.

FredKSchott:
@justinfagnani done! tests passing

justinfagnani:
Ok, #108 is in. Rebase! :)

FredKSchott:
done! That was a silly, 1-line rebase

src/commands/init.ts
```diff
@@ -30,7 +31,7 @@ export class InitCommand implements Command {
       description: 'The template name',
       type: String,
       defaultOption: true,
-    },
+    }
```
justinfagnani: we need to set up the linter, but I prefer trailing commands on all multiline arrays and object literals

PR #113, opened 2016-05-13
Fix duplicate argument definitions.

### LINKED ISSUE #114: `serve` and `lint` are broken

justinfagnani:
Fixes #114 

Existing tests now catch duplicate argument definitions across all commands because the detection is done in the CLI constructor.

### DIFF
--- a/src/commands/lint.ts
+++ b/src/commands/lint.ts
@@ -17,7 +17,39 @@ export class LintCommand implements Command {
 
   description = 'Lints the project';
 
-  args = polylint.argumentDefinitions;
+  args = [
+    {
+      name: "policy",
+      type: String,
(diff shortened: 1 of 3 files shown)


src/commands/lint.ts
```diff
@@ -17,7 +17,39 @@ export class LintCommand implements Command {
...
+      type: String,
+      defaultValue: "bower.json",
+      description: (
+        "If inputs are specified, look for `config-field` in this JSON file."
+      )
+    },
+    {
+      name: "config-field",
+      type: String,
+      defaultValue: "main",
```
azakus: This should be `entrypoint` now right?

azakus:
LGTM

PR #123, opened 2016-05-13
Add prompt to choose init template

justinfagnani:
We'll now be able to install new generators with `npm i -g`


src/commands/init.ts
```diff
@@ -28,20 +29,87 @@ export class InitCommand implements Command {
...
-      env.register(require.resolve('generator-polymer-init'), 'polymer-init:app');
-      env.registerStub(PolykartGenerator, 'polykart:app');
+      env.register(
+          require.resolve('generator-polymer-init'), 'polymer-init-basic:app');
+      env.registerStub(PolykartGenerator, 'polymer-init-shop:app');
+
+      env.lookup(() => {
+        let generators = env.getGeneratorsMeta();
+
+        let runGenerator = (generatorName: string, templateName?: string) => {
```
FredKSchott: nit: Are we really not doing named functions in this case? They'd fit this use better imo

src/commands/init.ts
```diff
@@ -28,20 +29,87 @@ export class InitCommand implements Command {
...
+
+        if (options.name) {
+          let generatorName = `polymer-init-${options.name}:app`;
+          runGenerator(generatorName, options.name);
+        } else {
+          let polymerInitGenerators = Object.keys(generators)
+              .filter((k) => k.startsWith('polymer-init')
+                  && k !== 'polymer-init:app');
+          let displayNames = polymerInitGenerators.map((n) => {
+          });
```
FredKSchott: old code?

src/commands/init.ts
justinfagnani: I basically default to arrow functions for everything exception functions which need to be able to have `this bound. That way I don't accidentally forget to `.bind()`, and the presence of `function` means I should pay attention because `this` is going to be interesting.

src/commands/init.ts
```diff
@@ -28,20 +29,87 @@ export class InitCommand implements Command {
...
+          let choices = polymerInitGenerators.map((generatorName: string) => {
```
FredKSchott: Can we combine this with the logic in L72-74 to just iterate over all generators directly (instead of pulling out the names above and then having to get back the generator here).

src/commands/init.ts
```diff
@@ -28,20 +29,87 @@ export class InitCommand implements Command {
...
+            let generator = generators[generatorName];
+            let description = 'no description';
+            let name = getDisplayName(generatorName);
+
+            if (name === 'basic') {
+              description = "built-in tempalte for basic apps and elements";
+            } else if (generator.resolved && generator.resolved !== 'unknown') {
+              console.log('generator.resolved', generator.resolved);
```
FredKSchott: logger.info :)

src/commands/init.ts
```diff
@@ -28,20 +29,87 @@ export class InitCommand implements Command {
...
+              // keeping this so things work when it does
+              short: description,
+            };
+          });
+          inquirer.prompt([{
+            type: 'list',
+            name: 'generatorName',
+            message: 'Choose a template to initialize this folder with',
+            choices: choices,
+          }]).then((answers) => {
```
FredKSchott: this is awesome

src/commands/init.ts
justinfagnani: removed

src/commands/init.ts
justinfagnani: sure, but one's a filter and one's a map, so it's about the same either way

src/commands/init.ts
FredKSchott: Got it, I don't feel the same but I understand your reasoning :)

src/commands/init.ts
justinfagnani: removed

src/commands/init.ts
```diff
@@ -28,20 +29,85 @@ export class InitCommand implements Command {
...
+        }
+
+        if (options.name) {
+          let generatorName = `polymer-init-${options.name}:app`;
+          runGenerator(generatorName, options.name);
+        } else {
+          let polymerInitGenerators = Object.keys(generators)
+              .filter((k) => k.startsWith('polymer-init')
+                  && k !== 'polymer-init:app');
+          let choices = polymerInitGenerators.map((generatorName: string) => {
```
FredKSchott: Can we combine this with the logic in L72-74 to just iterate over all generators directly (instead of pulling out the names above and then having to get back the generator here).

src/commands/init.ts
FredKSchott: Ah your right, for some reason I thought we could get away with only one run through. Feel free to do whichever you're feeling

FredKSchott:
LGTM!

PR #125, opened 2016-05-13
Move all command logging to use new logger

FredKSchott:
Everything old is using the new logger. All thats left to move over is within the smaller packages that polymer-cli pulls in.

### DIFF
--- a/custom_typings/plylog.d.ts
+++ b/custom_typings/plylog.d.ts
@@ -3,10 +3,10 @@ declare module 'plylog' {
   class PolymerLogger {
     constructor(options: any)
     setLevel(newLevel: string)
-    error(message: string, metadata?: any)
-    warn(message: string, metadata?: any)
-    info(message: string, metadata?: any)
-    debug(message: string, metadata?: any)
+    error(...data: any[])
+    warn(...data: any[])
+    info(...data: any[])
+    debug(...data: any[])
   }
 
   export function setVerbose();
--- a/custom_typings/plylog.d.ts
+++ b/custom_typings/plylog.d.ts
@@ -3,10 +3,10 @@ declare module 'plylog' {
   class PolymerLogger {
     constructor(options: any)
     setLevel(newLevel: string)
-    error(...data: any[])
-    warn(...data: any[])
-    info(...data: any[])
-    debug(...data: any[])
+    error(message: string, metadata?: any)
+    warn(message: string, metadata?: any)
+    info(message: string, metadata?: any)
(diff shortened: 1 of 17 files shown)


custom_typings/plylog.d.ts
```diff
@@ -3,10 +3,10 @@ declare module 'plylog' {
   class PolymerLogger {
     constructor(options: any)
     setLevel(newLevel: string)
-    error(message: string, metadata?: any)
-    warn(message: string, metadata?: any)
-    info(message: string, metadata?: any)
-    debug(message: string, metadata?: any)
+    error(...data: any[])
```
justinfagnani: this is pretty weak typing, does this describe the actual API in plylog?

package.json
```diff
@@ -40,7 +40,7 @@
     "merge-stream": "^1.0.0",
     "minimatch-all": "^1.0.2",
     "multipipe": "^0.3.0",
-    "plylog": "^0.3.0",
+    "plylog": "^0.4.0",
```
justinfagnani: What's in 0.4? It's not tagged on github. We need to do a review there too

src/build/analyzer.ts
```diff
@@ -213,7 +215,7 @@ class StreamResolver implements Resolver {
       if (file) {
         deferred.resolve(file.contents.toString());
       } else {
-        console.log('No file found for', filepath);
+        logger.info('No file found for', filepath);
```
justinfagnani: regarding the log method APIs, I'm fine with only the first argument being the message, now that JS has string interpolation.

justinfagnani:
This needs some rebasing now, sorry for moving the target on you! :)

custom_typings/plylog.d.ts
FredKSchott: as discussed, the winston logger is very lax. We'll keep the stronger argument definitions for our own sake.

package.json
FredKSchott: agreed, I've just been working on that in isolation up to this point. I'll invite you to a review of that library after this.

FredKSchott: oh right. v0.4 included work to share verbosity/level state across loggers. I had originally hoped we could get away with setting the state early and then having all loggers properly load that state on creation. But with module loading taking place before user input is read, that was unrealistic. In v0.4 changing the level now affects all loggers, instead of just yet-to-be-created loggers.

FredKSchott:
@justinfagnani rebased and responded. Should now be ready for another look / merge

justinfagnani:
LGTM

PR #129, opened 2016-05-15
Make sure unbundled precache deps list is complete

### LINKED ISSUE #128: Unbundled build is missing fragments and shell in sw-precache

azakus:
Include all the fragments and shell, just as bundled build does

Fixes #128 Unbundled build is missing fragments and shell in sw-precache


justinfagnani:
LGTM, but we really need some tests here. Merging to unblock @kevinpschaaf

PR #134, opened 2016-05-16
Let commands override properties of global argument definitions.

### LINKED ISSUE #133: polymer serve 'path' does not serve from path
Instead this serves from the current directory.

justinfagnani:
Fixes #133


azakus:
LGTM

PR #137, opened 2016-05-16
Make dependency stuffing optional

### LINKED ISSUE #138: Inserting additional links (prefetch and import) should be optional

azakus:
New flag `--insert-dependency-links`

Fixes #138

### DIFF
--- a/src/commands/build.ts
+++ b/src/commands/build.ts
@@ -29,6 +29,13 @@ export class BuildCommand implements Command {
       defaultValue: 'sw-precache-config.js',
       description: 'Path to an sw-precache configuration to be ' +
         'used for service worker generation.'
+    },
+    {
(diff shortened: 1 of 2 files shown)


src/commands/build.ts
```diff
@@ -29,6 +29,13 @@ export class BuildCommand implements Command {
       defaultValue: 'sw-precache-config.js',
       description: 'Path to an sw-precache configuration to be ' +
         'used for service worker generation.'
+    },
+    {
+      name: 'insert-dependency-links',
```
garlicnation: Default value?

justinfagnani:
LGTM!  (w/ AJ's comment addressed)

src/commands/build.ts
azakus: should be default false because it has type boolean

PR #142, opened 2016-05-16
New element and application templates

justinfagnani:


azakus:
PTAL @FredKSchott for yeoman stuff.

package.json
```diff
@@ -64,6 +64,8 @@
     "depcheck": "^0.6.3",
     "eslint": "^2.9.0",
     "gulp-mocha": "^2.2.0",
+    "gulp-typescript": "^2.13.4",
+    "gulp-typings": "^1.3.6",
     "sinon": "^1.17.4",
     "temp": "^0.8.3",
     "typescript": "^1.8.10",
```
FredKSchott: get rid of typescript & typings now that we're all in on gulp?

src/init/application/application.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  prompting() {
+    var _this = this;
+
+    var prompts = [
+      {
+        name: 'name',
+        type: 'input',
+        message: `Application name (${this.appname})`,
```
FredKSchott: IIRC yeoman already displays the default value next to the message in parenthesis, so including it here will result in a double-print.

src/init/application/application.ts
```diff
@@ -0,0 +1,100 @@
...
+        default() {
```
FredKSchott: nit: function is unnecessary here. if possible all defaults should be strings, or functions.

src/init/application/application.ts
FredKSchott: yup just confirmed

FredKSchott:
Really excited to see this land! 
btw comments apply to both generators

package.json
justinfagnani: done

src/init/application/application.ts
justinfagnani: done

justinfagnani:
@FredKSchott PTAL

FredKSchott:
LGTM!

PR #150, opened 2016-05-17
Fixups for github hosted templates

justinfagnani:


src/commands/init.ts
```diff
@@ -49,22 +49,23 @@ export class InitCommand implements Command {
...
       env.registerStub(ElementGenerator, 'polymer-init-element:app');
       env.registerStub(ApplicationGenerator, 'polymer-init-application:app');
       let shopGenerator = createGithubGenerator({
-        owner: 'PolymerLabs',
-        repo: 'polykart',
+        owner: 'Polymer',
+        repo: 'shop',
       });
       env.registerStub(shopGenerator, 'polymer-init-shop:app');
       let prplGenerator = createGithubGenerator({
```
azakus: appDrawerTemplateGenerator?

src/init/github.ts
```diff
@@ -59,7 +59,9 @@ export function createGithubGenerator(githubOptions: GithubGeneratorOptions) {
     }
 
     install() {
-      this.installDependencies();
+      this.installDependencies({
```
azakus: Should this check for `package.json`

azakus:
LGTM

src/init/github.ts
justinfagnani: eventually, yes. especially when a template needs it.

src/commands/init.ts
justinfagnani: done

PR #151, opened 2016-05-17
Move lint task into gulp

FredKSchott:
/cc @justinfagnani

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -11,10 +11,12 @@
 const fs = require('fs-extra');
 const gulp = require('gulp');
 const mergeStream = require('merge-stream');
+const eslint = require('gulp-eslint');
 const mocha = require('gulp-mocha');
 const path = require('path');
 const typescript = require('gulp-typescript');
 const typings = require('gulp-typings');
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -33,6 +35,20 @@ gulp.task('clean', (done) => {
 
 gulp.task('build-all', ['clean', 'init', 'build']);
 
+gulp.task('lint', function (cb) {
+  exec('depcheck . --ignores "generator-polymer-init"', function (err, stdout, stderr) {
```
justinfagnani: add: TODO: import depcheck directly so it doesn't need to be a global install

justinfagnani: fyi: https://github.com/mcasimir/gulp-depcheck

justinfagnani:
LGTM

gulpfile.js
FredKSchott: re global install: whoops, I completely missed that implication. I think this is a step backwards then (npm was including the dev dependency in the path when run). i'll close this until it's ready to go again.

PR #155, opened 2016-05-17
Only analyze entrypoint if the build doesn’t have shell and fragments.

### LINKED ISSUE #154: "Ignoring duplicate element definition"
Seems to be causing double-loading in apps as well.

justinfagnani:
Fixes #154

### DIFF
--- a/src/build/analyzer.ts
+++ b/src/build/analyzer.ts
@@ -49,7 +49,7 @@ export class StreamAnalyzer extends Transform {
       this.allFragments.push(shell);
     }
 
-    if (entrypoint) {
+    if (entrypoint && !shell && fragments.length === 0) {
       this.allFragments.push(entrypoint);
     }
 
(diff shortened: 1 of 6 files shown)


src/build/analyzer.ts
```diff
@@ -49,7 +49,7 @@ export class StreamAnalyzer extends Transform {
       this.allFragments.push(shell);
     }
 
-    if (entrypoint) {
+    if (entrypoint && !shell && fragments.length === 0) {
```
garlicnation: This means that the entry point must absolutely only import shell? Are we enforcing that throughout our tooling?

justinfagnani: The entrypoint can possibly import other things, but 1) they shouldn't and 2) we won't analyze or bundle it if there's a shell. This is part of the app structure we're defining in the docs. We can lint for this.

azakus:
LGTM, needs a nice refactor after 1.0 :)

PR #159, opened 2016-05-17
Much updated README

justinfagnani:


README.md
```diff
@@ -60,3 +156,9 @@ To compile your own version of the Polymer CLI clone this repo and make sure you
 Then run the build script
 
     $ npm run build
+
+## Supported node.js versions
+
+Polymer CLI targets the LTS version (4.x) of Node.js. It should
+work with the current version (6.x) but this is not officially
```
azakus: Why are we not officially supporting 6?

azakus:
LGTM

README.md
FredKSchott: Agreed, if we support any version of 6.x then theoretically we should work on all other 6.x versions.  If that's not true, it's Node's fault :)

FredKSchott:
Such documentation, wow.
LGTM!

PR #173, opened 2016-05-19
Add a web app manifest to the application template

### LINKED ISSUE #166: Feature request: autogenerate Web App Manifest
_As a Polymer CLI user,
when I use `polymer init` to start a new app,
I want the Polymer CLI to generate a [Web App Manifest](https://www.w3.org/TR/appmanifest/),
so that the default scaffolds are one step closer to a true PWA._

Polymer CLI already has a lot of the information necessary to generate a Web App Manifest. It would be really cool if I got one for free when I ran `polymer init`, and if it were already included in the generated `index.html`.

justinfagnani:
Fixes #166

### DIFF
--- a/src/init/application/templates/manifest.json
+++ b/src/init/application/templates/manifest.json
@@ -1,0 +1,8 @@
+{
+  "name": "<%= name %>",
+  "short_name": "<%= name %>",
+<% if (description) { -%>  "description": "<%= description %>",
+<% } -%>
+  "start_url": "/",
+  "display": "standalone"
+}
--- a/src/init/application/templates/index.html
+++ b/src/init/application/templates/index.html
@@ -5,6 +5,7 @@
     <title><%= name %></title>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="manifest" href="/manifest.json">
     <script src="/bower_components/webcomponentsjs/webcomponents-lite.js"></script>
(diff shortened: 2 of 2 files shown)


justinfagnani:
@cdata this is about all the information we have available right now. Things like theme color and icons will likely have to be in a more advanced template, like app-drawer.

src/init/application/templates/manifest.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "<%= name %>",
+  "short_name": "<%= name %>",
+  "description": "<%= description %>",
+  "start_url": "/",
+  "display": "standalone",
```
justinfagnani: I'm not really sure that "standalone" is the best option...

justinfagnani:
@garlicnation ping

src/init/application/templates/manifest.json
justinfagnani: Turns out PWAs need "standalone"

src/init/application/templates/manifest.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "<%= name %>",
+  "short_name": "<%= name %>",
+  "description": "<%= description %>",
```
FredKSchott: need an if statement to make sure description isn't empty string when not provided

justinfagnani: done

FredKSchott:
👍 LGTM

Issue #176, opened 2016-05-19
polymer init fails with third-party generators installed


ThatJoeMoore:
When a custom init template is installed, `polymer init` fails with the following error:

```
error [TypeError: iterator is not a function]
TypeError: iterator is not a function
    at FindUp._find (/usr/local/lib/node_modules/polymer-cli/node_modules/findup/index.js:60:3)
    at new FindUp (/usr/local/lib/node_modules/polymer-cli/node_modules/findup/index.js:53:8)
    at FindUp (/usr/local/lib/node_modules/polymer-cli/node_modules/findup/index.js:15:12)
    at module.exports (/usr/local/lib/node_modules/polymer-cli/node_modules/findup/index.js:10:10)
    at /usr/local/lib/node_modules/polymer-cli/lib/commands/init.js:94:44
    at Array.map (native)
    at /usr/local/lib/node_modules/polymer-cli/lib/commands/init.js:86:57
    at Environment.resolver.lookup (/usr/local/lib/node_modules/polymer-cli/node_modules/yeoman-environment/lib/resolver.js:50:12)
    at /usr/local/lib/node_modules/polymer-cli/lib/commands/init.js:58:17
    at new Promise (/usr/local/lib/node_modules/polymer-cli/node_modules/hydrolysis/node_modules/babel-polyfill/node_modules/core-js/modules/es6.promise.js:193:7)
```

This is due to an issue with the call to findup.

PR is #177.

ThatJoeMoore:
If anybody needs an example third-party template, you can `npm install -g byu-oit-appdev/polymer-yeoman#polymer-cli` for the current yeoman generator that I'm converting to an init template.

justinfagnani:
@ThatJoeMoore just to clarify, does this happen only with generators prefixed with `polymer-init`? If so, it's a bad bug, but maybe slightly less than critical :) I first thought that anyone with other Yeoman generators couldn't use `init`.

ThatJoeMoore:
Yes, it only affects generators prefixed with `polymer init`.  I don't think it's critical, as I doubt many people are writing custom generators yet. I just happened to be putting the finishing touches on one when this was announced and I just had to see if it all worked together (It all works wonderfully, BTW 👏).

PR #177, opened 2016-05-19
Fix findup dependency

ThatJoeMoore:
This resolves issue #176.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "command-line-commands": "^0.1.2",
     "css-slam": "^1.1.0",
     "dom5": "^1.3.1",
-    "findup": "^0.1.5",
+    "findup-sync": "^0.4.1",
     "fs-extra": "^0.30.0",
     "generator-polymer-init": "^0.0.3",
     "github": "^0.2.4",
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -38,7 +38,7 @@ export class InitCommand implements Command {
     // Defer dependency loading until this specific command is run
     const createGithubGenerator =
         require('../init/github').createGithubGenerator;
-    const findup = require('findup');
+    const findup = require('findup-sync');
     const inquirer = require('inquirer');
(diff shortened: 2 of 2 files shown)


justinfagnani:
Thanks for the PR!!

Looks like you need a rebase. Sorry I got in a bunch of lint commits just a bit ago.

Also, is there any chance you could get a test to fail before the fix?

ThatJoeMoore:
Rebased; I'll hopefully have the test cases for you later tonight or tomorrow morning.

justinfagnani:
Thanks a lot @ThatJoeMoore, especially for tests. We're really trying to increase the test coverage since we've seen a lot of breaks in the gaps that we have.

ThatJoeMoore:
Hmmm....I may have to decompose this function a bit to make it easier to test.  I can hold off if you want, but otherwise, I'm going to take a crack at it and see what happens.  I could submit the refactor as a separate PR if you want, just to keep the concerns separate.

justinfagnani:
Sure. I can try this locally and merge to unblock you to.

ThatJoeMoore:
I've submitted #195 with tests for InitCommand.  Once that PR is merged, I'll rebase this one and add the proper tests (they're actually in #195, just commented out).

ThatJoeMoore:
Just found the actual source of the problem - findup was being brought in, but was being used like findup-sync. Changing that dependency fixes the problem better than the former solution did.

admwx7:
Just manually applied this fix and renamed the generator I've been working on and it's getting pulled into the init menu properly.  Was failing prior to adding in the fix.

:+1:

FredKSchott:
Hey @ThatJoeMoore, thanks again for your help with this and the init tests PR.

We're now using findup somewhere else in the codebase. Instead of pulling in both libraries or using only the derivative `findup-sync` library, we're going to keep findup and update the code to call the synchronous interface explicitly. 

I'm going to close this in favor of #244. I'd love your review if you're able to.

Issue #206, opened 2016-05-25
Creating an Element Error: this.opts.filename.replce is not a function


Tykin:
Trying to create a polymer element using the polymer client tool in windows. The creation step fails repeatedly. Using Node 4.4.0.

```

C:\Users\tykin\test\test-element>polymer init
? Which starter template would you like to use? polymer-init-element:app
info:    Running template element
? Element name test-element
? Brief description of the element None
error:   Uncaught exception: TypeError: this.opts.filename.replace is not a function
error:   TypeError: this.opts.filename.replace is not a function
    at Object.Template.generateSource (C:\Program Files\nodejs\node_modules\polymer-cli\node_modules\ejs\lib\ejs.js:528:68)
    at Object.Template.compile (C:\Program Files\nodejs\node_modules\polymer-cli\node_modules\ejs\lib\ejs.js:444:12)
    at Object.compile (C:\Program Files\nodejs\node_modules\polymer-cli\node_modules\ejs\lib\ejs.js:289:16)
    at handleCache (C:\Program Files\nodejs\node_modules\polymer-cli\node_modules\ejs\lib\ejs.js:148:16)
    at Object.exports.render (C:\Program Files\nodejs\node_modules\polymer-cli\node_modules\ejs\lib\ejs.js:316:10)
    at copy.process (C:\Program Files\nodejs\node_modules\polymer-cli\node_modules\mem-fs-editor\lib\actions\copy-tpl.js:14:18)
    at applyProcessingFunc (C:\Program Files\nodejs\node_modules\polymer-cli\node_modules\mem-fs-editor\lib\actions\copy.js:13:16)
    at EditionInterface.exports._copySingle (C:\Program Files\nodejs\node_modules\polymer-cli\node_modules\mem-fs-editor\lib\actions\copy.js:51:16)
    at EditionInterface.<anonymous> (C:\Program Files\nodejs\node_modules\polymer-cli\node_modules\mem-fs-editor\lib\actions\copy.js:38:10)
    at Array.forEach (native)
```

yordis:
@Tykin Actually even for application type I get this message as well

MarcosGalaviz:
I have the same issue :(

FredKSchott:
I can reproduce this on Windows as well, not on Linux/Mac.

We're actively tracking and fixing a few different Windows bugs that have popped up since launching. See #199 for our current work to get tests passing on Windows. We'll tackle this as soon as possible once that's merged.

[...]

PR #207, opened 2016-05-25
Update call to copyTpl() to not throw in latest ejs

FredKSchott:
This bug was introduced by a breaking change in `mem-fs-editor`/`ejs`, that caused a previously supported behavior to throw. While original behavior may be restored downstream, we can fix this problem for ourselves today with this PR. See https://github.com/Polymer/polymer-cli/issues/206#issuecomment-221736928 for more info.

/cc @justinfagnani

### DIFF
--- a/test/init/application_test.js
+++ b/test/init/application_test.js
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+'use strict';
+
+const yoAssert = require('yeoman-assert');
+const helpers = require('yeoman-test');
+const ApplicationGenerator
+  = require('../../lib/init/application/application').ApplicationGenerator;
+
+suite('init/application', () => {
+
+  test('creates expected files while ignoring filenames with dangling underscores', (done) => {
+
+    helpers
+      .run(ApplicationGenerator)
+      .on('end', (a) => {
+        yoAssert.file(['bower.json']);
+        yoAssert.noFile(['src/_element/_element.html']);
+        done();
+      });
+
+  });
+
+});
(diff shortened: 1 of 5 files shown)


yordis:
@justinfagnani please can you merge this ASAP? Right now I can't use the `cli` at all :(

emartinpi:
+1 please

kevjs1982:
@yordis , a quick fix (assuming a global install and Windows 10) is to open a command prompt

`cd \Users\USERNAME\AppData\Roaming\npm\node_modules\polymer-cli\node_modules\yeoman-generator\node_modules\mem-fs-editor\node_modules`

(Replacing USERNAME with your username)

`rename ejs ejs-broken`

`npm install ejs@2.4.1`

FredKSchott:
Getting this merged along with the fixes for Windows users are our top priorities right now, expect this to land on npm sometime later today.

@kevjs1982 smart fix! For Mac/UNIX use 'mv' instead of 'replace'.

yordis:
@kevjs1982 I am on Mac

FredKSchott:
This is affecting everyone on all systems who's installed/updated since
yesterday morning.

justinfagnani:
@FredKSchott please rebase on master so we can make sure the Windows CIs are green. Also, I'd love to some test, no matter how small, trigger the original failure. There's a test of the github generator to base of off, and it should be pretty trivial to just run with preset answers to prompts. See: http://yeoman.io/authoring/testing.html

FredKSchott:
@justinfagnani rebased and tests added. Lets see what appveyor has to say...

test/init/application_test.js
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+'use strict';
+
+const assert = require('chai').assert;
+const fs = require('fs');
+const path = require('path');
+const yoAssert = require('yeoman-assert');
+const helpers = require('yeoman-test');
+const ApplicationGenerator = require('../../lib/init/application/application.js').ApplicationGenerator;
```
justinfagnani: long line?

justinfagnani:
LGTM++

test/init/application_test.js
FredKSchott: ewww... what should this split look like?

```
const ApplicationGenerator 
  = require('../../lib/init/application/application').ApplicationGenerator;
```

justinfagnani:
Looks like lint errors. Make sure to run `gulp lint` or `gulp build-app && gulp test` before pushing :)

test/init/application_test.js
justinfagnani: that's fine

FredKSchott:
-_-
If we're failing on lint errors lets add linting to `gulp test`?

justinfagnani:
Maybe we could add a `test-lite` target? I want a test target that's a fast as possible for dev-time iteration

yordis:
@justinfagnani when are you gonna create a release for get this changes over npm?

MarcosGalaviz:
Hi thanks for your help but now the error is diferent :(

error

PR #211, opened 2016-05-26
Remove silently failing depcheck task from build-all task

FredKSchott:
This will allow polymer-cli to build properly until we can figure out why `depcheck` is silently failing.

@justinfagnani suggest merging and pushing out a new patch version asap


FredKSchott:
Superceded by https://github.com/Polymer/polymer-cli/pull/212

PR #212, opened 2016-05-26
Fix broken depcheck task, use promise interface instead of callback

FredKSchott:
This will allow polymer-cli to lint, build, and publish properly.

@justinfagnani suggest merging and pushing out a new patch version asap

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -64,22 +64,15 @@ gulp.task('eslint', () =>
     .pipe(eslint.format())
     .pipe(eslint.failAfterError()));
 
-gulp.task('depcheck', () => new Promise((resolve, reject) => {
-  depcheck(__dirname, {ignoreDirs: ['generator-polymer-init']}, (result) => {
-    let invalidFiles = Object.keys(result.invalidFiles) || [];
-    let invalidJsFiles = invalidFiles.filter((f) => f.endsWith('.js'));
-    if (invalidJsFiles.length > 0) {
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -69,14 +69,12 @@ gulp.task('depcheck', () => new Promise((resolve, reject) => {
...
     let invalidJsFiles = invalidFiles.filter((f) => f.endsWith('.js'));
     if (invalidJsFiles.length > 0) {
-      console.log('Invalid files:', result.invalidFiles);
-      reject(new Error('Invalid files'));
+      reject(new Error(`Invalid files: ${result.invalidFiles}`));
       return;
     }
 
     if (result.dependencies.length) {
-      console.log('Unused dependencies:', unused.dependencies);
```
FredKSchott: Gulp just silently failed here. It looks like Gulp is swallowing exceptions. I'll continue to dig...

gulpfile.js
```diff
@@ -64,22 +64,15 @@ gulp.task('eslint', () =>
     .pipe(eslint.format())
     .pipe(eslint.failAfterError()));
 
-gulp.task('depcheck', () => new Promise((resolve, reject) => {
-  depcheck(__dirname, {ignoreDirs: ['generator-polymer-init']}, (result) => {
```
FredKSchott: https://github.com/depcheck/depcheck/blob/master/src/index.js#L68
oh boy...

justinfagnani:
LGTM

PR #221, opened 2016-05-28
Remove unused generator-polymer-init dependency

FredKSchott:
element & application templates are now internal to polymer-cli.

I'll add a deprecation notice to the `generator-polymer-init` README as well.

/cc @garlicnation


justinfagnani:
LGTM

PR #229, opened 2016-05-31
Add demo and description for element template

honzalo:
Currently there is no demo reference in the element template.
Thus when creating a new element and serving it, there is no demo option, although the demo folder and demo template are created. It will be more consistent to include the demo reference (`@demo`) given that the demo template is provided.

### DIFF
--- a/src/init/element/templates/_element.html
+++ b/src/init/element/templates/_element.html
@@ -1,5 +1,11 @@
 <link rel="import" href="../polymer/polymer.html">
 
+<!--
+    <%= description %>
+    
+    @demo demo/index.html 
+-->
+
 <dom-module id="<%= name %>">
   <template>
     <style>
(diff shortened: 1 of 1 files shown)


chuckh:
LGTM

FredKSchott:
You're right, the polymer docs do mention properly documenting your element with tags like this https://www.polymer-project.org/1.0/docs/tools/documentation

Because this is auto-generated and the description field is sometimes left blank, I'm worried about this looking odd for most people when they don't include one:

```
<!--


@demo demo/index.html 
 -->
```

@honzalo Can you change the template to make it more flexible? Perhaps something like this?
```
<!--
<example-element-name></example-element-name>

@demo demo/index.html 
 -->
```
```
<!--
<example-element-name></example-element-name>
A description was provided, so we include it here.

@demo demo/index.html 
 -->
```

justinfagnani:
@honzalo thanks for the PR!

I agree with @FredKSchott and with that addressed, I also agree with @chuckh: LGTM :)

addyosmani:
+1 to @FredKSchott's comments and then LGTM :)

honzalo:
Updated PR as per @FredKSchott suggestion. Thank you.

PR #231, opened 2016-05-31
Refactor project-config to make polymer-cli testable /w configuration

FredKSchott:
I was running into problems testing different commands with different configurations, so I've refactored `polymer-cli` & `project-configuration` to read from an optional configuration object when given, instead of always reading directly from file.

#232 illustrates this usefulness, which I've split out from this work for clarity.

/cc @justinfagnani @garlicnation

### DIFF
--- a/test/config/project-config_test.js
+++ b/test/config/project-config_test.js
@@ -14,7 +14,16 @@ const path = require('path');
 const ProjectConfig = require('../../lib/project-config').ProjectConfig;
 
 suite('Project Config', () => {
-  test('sane defaults', () => {
+
+  test('reads options from config file', () => {
+    let defaultOptions = ProjectConfig.fromConfigFile(path.join(__dirname, 'polymer.json'));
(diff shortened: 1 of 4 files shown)


test/config/project-config_test.js
```diff
@@ -34,15 +51,9 @@ suite('Project Config', () => {
     assert.deepEqual(config.fragments, [path.resolve('foo.html')]);
   });
 
-  test('inits from config file', () => {
```
garlicnation: Where did this test go?

FredKSchott: It's equivalent (after refactor) is now 'reads options from config file' https://github.com/Polymer/polymer-cli/pull/231/files/156f4a52154e51904c973fcf4627e749e9a73c98#diff-062b6039a01f744d6481ace6873ad64eR18

garlicnation:
LGTM

PR #232, opened 2016-05-31
Add input support for lint command

### LINKED ISSUE #227: polymer lint fails on basic element project layout
Hi!

I just followed the steps in the [docs](https://www.polymer-project.org/1.0/docs/tools/polymer-cli#element) to create a new element project and when I try to lint the command fails in different ways.

Using the command [specified in the docs
](https://www.polymer-project.org/1.0/docs/tools/polymer-cli#lint)
```
$ polymer lint --input my-el.html
error:   Uncaught exception: UNKNOWN_OPTION: Unknown option: --input
error:   UNKNOWN_OPTION: Unknown option: --input
    at halt (/usr/local/lib/node_modules/polymer-cli/node_modules/command-line-args/lib/argv.js:81:15)
    at Argv.validate (/usr/local/lib/node_modules/polymer-cli/node_modules/command-line-args/lib/argv.js:72:7)
    at CommandLineArgs.parse (/usr/local/lib/node_modules/polymer-cli/node_modules/command-line-args/lib/command-line-args.js:57:10)
    at CommandLineCommands.parse (/usr/local/lib/node_modules/polymer-cli/node_modules/command-line-commands/lib/command-line-commands.js:78:28)
    at PolymerCli.run (/usr/local/lib/node_modules/polymer-cli/lib/polymer-cli.js:115:35)
    at /usr/local/lib/node_modules/polymer-cli/bin/polymer.js:22:7
    at /usr/local/lib/node_modules/polymer-cli/node_modules/resolve/lib/async.js:46:14
    at process (/usr/local/lib/node_modules/polymer-cli/node_modules/resolve/lib/async.js:173:43)
    at ondir (/usr/local/lib/node_modules/polymer-cli/node_modules/resolve/lib/async.js:188:17)
    at load (/usr/local/lib/node_modules/polymer-cli/node_modules/resolve/lib/async.js:69:43)
garcianavalon at eva in ~/Development/my-el
```

Trying different options based on [the README of this repo](https://github.com/Polymer/polymer-cli#lint)
(I tried creating a polymer.json file with content `{"entrypoint": "my-el.html"}` but the result is the same)
```
$ polymer lint my-el.html
ERROR finding /Users/garcianavalon/Development/my-el/polymer/polymer.html
Error: ENOENT: no such file or directory, open '/Users/garcianavalon/Development/my-el/polymer/polymer.html'
    at Error (native)
```
```
$ polymer lint
ERROR finding /Users/garcianavalon/Development/my-el/polymer/polymer.html
Error: ENOENT: no such file or directory, open '/Users/garcianavalon/Development/my-el/polymer/polymer.html'
    at Error (native)
```

Looks like the import is not working properly
`<link rel="import" href="../polymer/polymer.html">`

I'm using latest version
```
$ polymer --version
0.11.0
```
Any idea on what could be wrong? 
Thanks for any help!
cheers.

FredKSchott:
*NOTE: this is built off of the project-config work I just pushed into #231, so the "files changed" is currently a little noisy. See the individual commit for less noise.*

This adds support for linting of explicit files via the `input` default argument. This has been supported in our [polymer documentation](https://github.com/Polymer/docs2/blob/master/app/1.0/docs/tools/polymer-cli.md#lint-elements-lint) & README but was never actually supported by our code.

~~This also fixes an issue that seems to have been introduced with the Windows path fixes, by removing the leading slash that polylint was not expecting.~~ Pulling into its own PR

Fixes #227
/cc @justinfagnani @garlicnation

### DIFF
--- a/src/commands/lint.ts
+++ b/src/commands/lint.ts
@@ -21,6 +21,14 @@ export class LintCommand implements Command {
 
   args = [
     {
+      name: 'input',
+      type: String,
+      alias: 'i',
+      defaultOption: true,
+      multiple: true,
+      description: 'Files and/or folders to lint. Exclusive. Defaults to cwd.'
+    },
+    {
       name: 'policy',
       type: String,
       alias: 'p',
@@ -45,10 +53,21 @@ export class LintCommand implements Command {
       )
     },
     {
-      name: 'no-recursion',
+      name: 'follow-dependencies',
       type: Boolean,
       description: (
-        'Only report errors on specified input files, not from their dependencies.'
+        'Follow through and lint dependencies. This is default behavior ' +
+        'when linting your entire application via the entrypoint, shell, ' +
+        'and fragment arguments.'
+      )
+    },
+    {
+      name: 'no-follow-dependencies',
+      type: Boolean,
+      description: (
+        'Only lint the files provided, ignoring dependencies. This is ' +
+        'default behavior when linting a specific list of files provided ' +
+        'via the input argument.'
       )
     }
   ];
@@ -57,12 +76,12 @@ export class LintCommand implements Command {
     // Defer dependency loading until this specific command is run
     const polylint = require('polylint/lib/cli');
 
-    if (config.inputs.length === 0) {
+    let lintFiles: string[] = options.input
+      || config.inputs.map((i) => i.substring(config.root.length));
+    if (lintFiles.length === 0) {
+      logger.warn('No inputs specified. Please use the --input, --entrypoint, ' +
+        '--shell or --fragment flags');
       let argsCli = commandLineArgs(this.args);
-
-      logger.warn('No inputs specified. Please use the --entrypoint, --shell ' +
-          'or --fragment flags');
-
       console.info(argsCli.getUsage({
         title: `polymer ${this.name}`,
         description: this.description,
@@ -70,15 +89,24 @@ export class LintCommand implements Command {
       return Promise.resolve();
     }
 
+    // Default to false if input files are provided, otherwise default to true
+    let followDependencies = !options.input;
+    if (options['follow-dependencies']) {
+      followDependencies = true;
+    } else if (options['no-follow-dependencies']) {
+      followDependencies = false;
+    }
+
     return polylint.runWithOptions({
-      input: config.inputs.map((i) => i.substring(config.root.length)),
+      input: lintFiles,
       root: config.root,
       // TODO: read into config
       bowerdir: 'bower_components',
       policy: options.policy,
       'config-file': options['config-file'],
       'config-field': options['config-field'],
-      'no-recursion': options['no-recursion'],
+      // NOTE: `no-recursion` has the opposite behavior of `follow-dependencies`
+      'no-recursion': !followDependencies,
     }).then(() => null);
   }
 }
--- a/test/commands/lint_test.js
+++ b/test/commands/lint_test.js
@@ -20,9 +20,15 @@ const sinon = require('sinon');
 suite('lint', () => {
 
   let polylintCliStub;
+  let defaultTestConfig;
 
   setup(function() {
     polylintCliStub = sinon.stub(polylintCli, 'runWithOptions').returns(Promise.resolve());
+    defaultTestConfig = new ProjectConfig(null, {
+      entrypoint: 'index.html',
+      fragments: ['foo.html'],
+      shell: 'bar.html',
+    });
   });
 
   teardown(() => {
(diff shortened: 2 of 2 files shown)


src/commands/lint.ts
```diff
@@ -24,6 +24,14 @@ export class LintCommand implements Command {
 
   args = [
     {
+      name: 'input',
```
garlicnation: Why use this instead of concating shell/main/entrypoint?

FredKSchott: It seems to be a common usecase for people getting started, or debugging linting issues. Being able to do `polymer lint path/to/file.html` is mentioned in both https://github.com/Polymer/polymer-cli#lint & https://www.polymer-project.org/1.0/docs/tools/polymer-cli#lint.

@robdodson might be able to shed some user-land light on this

robdodson: Not sure I fully understand what this does but the issue I've been bumping into is wanting to lint my entire project but exclude certain directories (bower_components, for example)

FredKSchott:
Closing this until we can discuss what we want expected behavior to be. Also I think there might be a bug in my linting fix 🙊

justinfagnani:
I like 80% of this change: adding the --input flag. Per in-person discussion, I think it should imply `--no-recursion`. Also, `--no-recursion` should be renamed to `--lint-dependencies` or some similar default-true (except in the presents of `--input`) flag.

FredKSchott:
@justinfagnani PTAL, renamed "no-recursion" to "follow-dependencies"

Just realized that the reason we built "no-X" flags for "default on" behaviors is because their presence is the only thing that turns those behaviors off. (ex:no way to set follow-dependencies to "off" with only the  `--follow-dependencies` flag).

Any strong opinions on how we do this? I'd be fine with either defaulting to never following unless `--follow-dependencies` is present, or supporting both `--follow-dependencies` & `--no-follow-dependencies` flags (git and some others use this system)

justinfagnani:
@FredKSchott I asked around and it seems like `--no-follow-dependencies` might be preferred :/

src/commands/lint.ts
justinfagnani: @robdodson I think glob support would be a great follow on to this. By default, given the project entrypoints, we shouldn't be linting bower_components.

src/commands/lint.ts
```diff
@@ -48,37 +56,42 @@ export class LintCommand implements Command {
       )
     },
     {
-      name: 'no-recursion',
+      name: 'follow-dependencies',
       type: Boolean,
       description: (
-        'Only report errors on specified input files, not from their dependencies.'
+        'Follow through and lint dependencies. Defaults to true unless ' +
+        'an exclusive list of input files is provided.'
```
justinfagnani: mention the `--input` flag explicitly.

FredKSchott:
@justinfagnani just to clarify, you're saying you prefer --no-follow-dependencies and then always following dependencies by default, even for `--input`? Or do we just `no-follow` by default for input, and then not support following dependencies at all for input (that's my preference, given the circumstance).

justinfagnani:
@FredKSchott I prefer not following dependencies for `--input`, so... maybe both `--no-follow-dependencies` and `--follow-dependencies` for use with `--input`.

FredKSchott:
👍 sounds good

src/commands/lint.ts
FredKSchott: done

FredKSchott:
@justinfagnani ptal

justinfagnani:
@FredKSchott  can you rebase?

src/commands/lint.ts
```diff
@@ -48,37 +56,57 @@ export class LintCommand implements Command {
...
+        'Only lint the files provided, ignoring dependencies. This is ' +
+        'default behavior when linting a specific list of files provided ' +
+        'via the input argument.'
       )
     }
   ];
 
   run(options, config): Promise<any> {
-    if (config.inputs.length === 0) {
+    let lintFiles: string[] = options.input
```
justinfagnani: empty lists coerce to `true`. Do you want to check the length of input as well?

test/commands/lint_test.js
```diff
@@ -38,7 +38,83 @@ suite('lint', () => {
...
-    assert.deepEqual(polylintCliStub.firstCall.args[0].input, [`${path.sep}index.html`, `${path.sep}bar.html`, `${path.sep}foo.html`]);
+    assert.deepEqual(polylintCliStub.firstCall.args[0].input, [
+      `${path.sep}index.html`,
+      `${path.sep}bar.html`,
+      `${path.sep}foo.html`,
+    ]);
+  });
+
+  test('lints the given files when provided through the `input` argument', () => {
+    let testLintConfig = new ProjectConfig(null, {
```
justinfagnani: much copy paste of the config...

justinfagnani:
LGTM with a couple of very minor comments

test/commands/lint_test.js
FredKSchott: good catch, I'll pull it out

src/commands/lint.ts
FredKSchott: So the only time `options.input` will be an empty list is if you do `polymer lint --input`. It's a strange case but if that did happen I would expect it to lint zero files. Maybe throw an error or log a warning in that case?

PR #236, opened 2016-06-02
small renaming and adding of logging to build/bundle

FredKSchott:
This is in preparation for the Windows fix I've just completed. I'd
like to push these separately so that the windows fix can be as
concise as possible.

In this commit I added logging that helps detect the issue of missing
duplicate matches during bundling. I also renamed the path/osPath
variable names to be more intuitive (`path` now refers to the path
library).

This should have no effect on behavior.

/cc @garlicnation @justinfagnani


garlicnation:
LGTM

PR #237, opened 2016-06-02
Fix build bundling on windows

### LINKED ISSUE #222: Error in polymer build with the last release (v0.11.0)
With the last release (v0.11.0), `polymer build` doesn't work:

error

### LINKED ISSUE #227: polymer lint fails on basic element project layout

FredKSchott:
Fixes #222

Previously, we were relying on path.posix.normalize() to convert unix & windows paths to the posix path format, to be returned as a URL path. However, path.posix.normalize() doesn't convert windows paths to posix. Example:

```
> path.posix.normalize('/src/123.html')
'/src/123.html'
> path.posix.normalize('\\src\\123.html')
'\\src\\123.html // expected url format `/src/123/html`
```

Both @justinfagnani  and I have looked into this, and the conclusion is that Node does not support this conversion natively. 

This commit adds the [slash](https://www.npmjs.com/package/slash) library to handle this conversion instead. This is a simple, well-supported package used by babel and some other major projects.

Our original worry was that escaped back slashes would be incorrectly converted, but Windows will actually not allow any slashes in file names, escaped or otherwise. See:
\- http://stackoverflow.com/a/10708449
\- https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx

On Mac & Unix, backslashes are allowed in file names. This means that slash will not properly convert those strings:

```
> path.posix.normalize('/src/this_file_has_escaped\\_see.html')
'/src/this_file_has_escaped/_see.html'
```

However, the files we are converting and checking against are all meant for HTML import URL paths. Backward slashes are not allowed in URLs (RFC 2396) and most browsers will just convert them to forward slashes. So this case would already be erroneous, and as a result we do not really need to support it. One possible enhancement would be to throw a helpful error if we detect a backward slash in a posix formatted path, because right now we'll just silently convert it to a forward slash separator.

### DIFF
--- a/test/build/bundle_test.js
+++ b/test/build/bundle_test.js
@@ -23,6 +23,7 @@ const Bundler = bundle.Bundler;
 const StreamAnalyzer = analyzer.StreamAnalyzer;
 
 const root = path.resolve('/root');
+var isPlatformWin = /^win/.test(process.platform);
 
 suite('Bundler', () => {
 
@@ -96,6 +97,21 @@ suite('Bundler', () => {
       ));
     return link != null;
   };
+
+  suite('urlFromPath()', () => {
+
+    // TODO(fschott): Get path-related tests working in multiple environments
+    if(!isPlatformWin) {
+      test('creates a URL path relative to root when called with a file path', () => setupTest({
+        entrypoint: 'entrypointA.html',
+        files: [framework(), entrypointA()],
+      }).then((files) => {
+        let urlPath = bundler.urlFromPath('/root/src/123.html');
+        assert.equal(urlPath, 'src/123.html');
+      }));
+    }
+
+  });
 
   test('entrypoint only', () => setupTest({
     entrypoint: 'entrypointA.html',
--- a/custom_typings/slash.d.ts
+++ b/custom_typings/slash.d.ts
@@ -1,0 +1,3 @@
+declare module 'slash' {
+  export function slash(source: string): string;
+}
--- a/src/build/bundle.ts
+++ b/src/build/bundle.ts
@@ -108,8 +108,16 @@ export class Bundler extends Transform {
     if (!filepath.startsWith(this.root)) {
       throw new Error(`file path is not in root: ${filepath} (${this.root})`);
     }
-    // convert filesystem path to URL
-    return posixPath.normalize(path.relative(this.root, filepath));
+    // On windows systems, convert filesystem path to URL by replacing slashes
+    let isPlatformWin = /^win/.test(process.platform);
+    let isExtendedLengthPath = /^\\\\\?\\/.test(filepath);
+    let hasNonAscii = /[^\x00-\x80]+/.test(filepath);
+    if (isPlatformWin && !isExtendedLengthPath && !hasNonAscii) {
+      return path.relative(this.root, filepath).replace(/\\/g, '/');
+    }
+
+    // Otherwise, just return the relative path between the two
+    return path.relative(this.root, filepath);
   }
 
   _buildBundles(): Promise<Map<string, string>> {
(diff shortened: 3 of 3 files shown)


test/build/bundle_test.js
```diff
@@ -97,6 +97,18 @@ suite('Bundler', () => {
     return link != null;
   };
 
+  suite('urlFromPath()', () => {
+
+    test.skip('creates a URL path relative to root when called with a file path', () => setupTest({
```
FredKSchott: Our testing setup currently can't test path-related behavior on Windows. My next task will be to make this test and the several under it runnable.

justinfagnani: rather than always skipping the test, can you conditionally skip it based on the platform?

garlicnation:
LGTM - please add a TODO to explore getting rid of the added dependency.

test/build/bundle_test.js
FredKSchott: good call, done

FredKSchott:
Got rid of the dependency, its logic actually fit really well into the function itself

just waiting for tests to pass and then I'll merge

PR #239, opened 2016-06-03
Split url-from-path logic out and into its own file

FredKSchott:
This splits our `urlFromPath()` logic out of the bundler and into it's own module. This allows us to add back the proper license for the slash library, as well as clean up some duplicate code by refactoring analyzer to use the module as well.

/cc @garlicnation


garlicnation:
LGTM

PR #240, opened 2016-06-03
Removes unneccesary liftoff dependency

FredKSchott:
The function we were using liftoff for was mostly delegated to findup
a dependency we are already using. This replaces liftoff with a findup-
only implementation, allowing us to remove liftoff (and it's ~50 deps)
from our dep tree.

Also adds some logging.

/cc @garlicnation

### DIFF
--- a/src/build/build.ts
+++ b/src/build/build.ts
@@ -18,6 +18,7 @@ import * as path from 'path';
 import {PassThrough, Readable} from 'stream';
 import File = require('vinyl');
 import * as vfs from 'vinyl-fs';
+import * as findup from 'findup';
 import * as logging from 'plylog';
 
 import {StreamAnalyzer, DepsIndex} from './analyzer';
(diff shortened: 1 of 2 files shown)


src/build/build.ts
```diff
@@ -29,12 +30,24 @@ import {PrefetchTransform} from './prefetch';
...
 import {generateServiceWorker, parsePreCacheConfig, SWConfig} from './sw-precache';
 
-
 // non-ES compatible modules
-const findConfig = require('liftoff/lib/find_config');
 const minimatchAll = require('minimatch-all');
 let logger = logging.getLogger('cli.build.build');
 
+function getGulpfile(): any {
+  // TODO: cwd = root?
```
garlicnation: Any reason not to include this in the PR?

FredKSchott: how do you mean?

garlicnation:
LGTM.

@justinfagnani Should we change this to use config.root?

PR #246, opened 2016-06-06
Refactor init command tests and make the cli commands more testable

FredKSchott:
There was considerable configuration/boilerplate brought into the `init` command tests, as a workaround to how self-contained and complex the command was. This refactoring exposes the yeoman environment as a property of the command, allowing us to inspect yeoman, its generators, and its prompts. As a result we can now remove the mockery dependency that was added for this.

/cc @justinfagnani

### DIFF
--- a/src/init/init.ts
+++ b/src/init/init.ts
@@ -1,0 +1,181 @@
+/**
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+import {execSync} from 'child_process';
+import * as fs from 'fs';
+import * as logging from 'plylog';
+import * as chalk from 'chalk';
+import * as findup from 'findup';
+import {ApplicationGenerator} from '../init/application/application';
+import {ElementGenerator} from '../init/element/element';
+import * as YeomanEnvironment from 'yeoman-environment';
+import * as inquirer from 'inquirer';
+import {createGithubGenerator} from '../init/github';
+
+let logger = logging.getLogger('init');
+
+const templateDescriptions = {
+  'element': 'A blank element template',
+  'application': 'A blank application template',
+  'app-drawer-template': 'A starter application template, with navigation and "PRPL pattern" loading',
+  'shop': 'The "Shop" Progressive Web App demo',
+};
+
+interface GeneratorDescription {
+  name: string;
+  value: string;
+  short: string;
+}
+
+let shopGenerator = createGithubGenerator({
+  owner: 'Polymer',
+  repo: 'shop',
+});
+
+let appDrawerGenerator = createGithubGenerator({
+  owner: 'Polymer',
+  repo: 'app-drawer-template',
+});
+
+function checkIsMinGW(): boolean {
+  let isWindows = /^win/.test(process.platform);
+  if (!isWindows) {
+    return false;
+  }
+
+  // uname might not exist if using cmd or powershell,
+  // which would throw an exception
+  try {
+    let uname = execSync('uname -s').toString();
+    return !!/^mingw/i.test(uname);
+  } catch (err) {
+    logger.debug('`uname -s` failed to execute correctly', {err: err.message});
+    return false;
+  }
+}
+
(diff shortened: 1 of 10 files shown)


src/polymer-cli.ts
```diff
@@ -165,6 +165,13 @@ export class PolymerCli {
     let command = this.commands.get(commandName);
     logger.debug('Running command...');
 
+    // Do any pre-run command initialization here, useful for setting up
+    // your command and any other expensive loading you don't want to do
+    // until neccessary.
+    if (command.init) {
```
FredKSchott: In a future PR I'd like to get rid of these startup-time optimizations that we are building into & maintaining inside of each command, and instead make the CLI smarter about how/when it loads commands (should only do so when needed). @justinfagnani let me know if you disagree with that direction.

src/polymer-cli.ts
```diff
@@ -165,6 +165,13 @@ export class PolymerCli {
     let command = this.commands.get(commandName);
     logger.debug('Running command...');
 
+    // Do any pre-run command initialization here, useful for setting up
```
justinfagnani: Move comment to the declaration of `init`

FredKSchott: done

justinfagnani:
I think there might be a better organization of commands and the top-level CLI that this behavior falls out of naturally. Talked in person.

test/commands/init_test.js
```diff
@@ -10,258 +10,190 @@
...
-    inquirerStub.prompt.returns(Promise.resolve({
-      generatorName: generatorName,
-    }))
-  }
-
-  test('registers default template generators', () => {
-    let registeredGens = {};
-    let registerSpy = sandbox.spy(function (gen, name) {
-      registeredGens[name] = gen;
+  test('calls ', () => {
```
justinfagnani: Extra space in test name. Forgot a

test/commands/init_test.js
```diff
@@ -10,258 +10,190 @@
...
   setup(() => {
     sandbox = sinon.sandbox.create();
-    originalPlatform = Object.getOwnPropertyDescriptor(process, 'platform');
-    mockery.enable({useCleanCache: true});
-    mockery.registerAllowables(['../../lib/commands/init']);
-    mockDependencies();
-    //We need to re-import this each time mockery gets set up
-    // so that we have fresh stubs/spies
-    InitCommand = require('../../lib/commands/init').InitCommand
+    cli = new PolymerCli(['init']);
```
justinfagnani: Might it be easier to just test the init library directly?

test/commands/init_test.js
```diff
@@ -10,258 +10,190 @@
...
-          match.same(elementGeneratorStub.ElementGenerator),
-          'polymer-init-element:app'
-        ));
+  });
 
-        assert.isTrue(registerSpy.calledWith(
-          match.same(applicationGeneratorStub.ApplicationGenerator),
-          'polymer-init-application:app'
-        ));
+  test('prompts with a list of all registered generators', (done) => {
```
justinfagnani: Return a Promise rather than using `done`

src/init/init.ts
```diff
@@ -0,0 +1,181 @@
...
+  });
+}
+
+export function promptGeneratorSelection(options): Promise<any> {
+  options = options || {};
+
+  return new Promise((resolve, reject) => {
+    resolve(options.env || createYeomanEnvironment());
+  }).then(function(env) {
+    let generators = env.getGeneratorsMeta();
```
justinfagnani: Factor this out into a simple function that generates the template list. It'll be easier to test that.

test/commands/init_test.js
```diff
@@ -10,258 +10,190 @@
...
+            },
+          ],
+        },
+      ]));
+      done();
+    });
+  });
 
-        assert.isTrue(registerSpy.calledWith(
+  test('includes user-provided generators in the list when properly installed/registered', () => {
```
justinfagnani: fi we assume we've tested the ability of the command to actually perform the prompting, then testing user-installed generators is once of the things that would be a lot easier by just testing the generator list functionality directly.

test/commands/init_test.js
```diff
@@ -10,258 +10,190 @@
...
+  if (isPlatformWin) {
+
+    test('prompts with a rawlist if being used in MinGW shell', () => {
+      let promptStub = sandbox.stub(inquirer, 'prompt').returns(new Promise(function() {}));
+      sandbox.stub(childProcess, 'execSync').withExactArgs('uname -s').returns('mingw');
+      cli.run();
+
+      assert.isTrue(promptStub.calledWith([
+        {
+          "type":"rawlist",
```
justinfagnani: The only thing we care about here is the type, no? Can we make the test smaller so it's less brittle to changes in the set of generators and their descriptions, etc?

FredKSchott:
@justinfagnani sorry for the confusion, closing this until it's ready for review

PR #251, opened 2016-06-08
Move last of the old commands to load their dependencies at runtime

FredKSchott:
This moves the last of the commands  (besides init, which is complex enough to get its own PR) to use lazy dependency loading. We weren't able to move these previously because they were needed on load for our CLI argument definitions.

This (/w init fixed as well) drops our startup time from the original `~1.6s` down to `~0.2s`. booyah!

/cc @justinfagnani


justinfagnani:
LGTM++!

PR #255, opened 2016-06-09
Refactor init command tests and make the cli commands more testable

FredKSchott:
> *Whoops, looks like you can't reopen the original PR once closed. See https://github.com/Polymer/polymer-cli/pull/246 for past discussion (most of it outdated)*

There was considerable configuration/boilerplate brought into the `init` command tests, as a workaround to how self-contained and complex the command was. This refactoring pulls the main init logic out of the command itself, which has several benefits to our architecture:

- Removes the mockery dependency that was added for testing the original code 
- Reduces startup time considerably by delaying yeoman loading until the init command is run
- Keeps the command small and singular in focus

/cc @justinfagnani

### DIFF
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -9,30 +9,10 @@
  */
 
 import {Command} from './command';
-import {execSync} from 'child_process';
 import {ArgDescriptor} from 'command-line-args';
-import * as fs from 'fs';
 import * as logging from 'plylog';
-import * as chalk from 'chalk';
-import {ApplicationGenerator} from '../init/application/application';
-import {ElementGenerator} from '../init/element/element';
-import * as YeomanEnvironment from 'yeoman-environment';
-import * as findup from 'findup';
 
-let logger = logging.getLogger('cli.init');
-
-const templateDescriptions = {
-  'element': 'A blank element template',
-  'application': 'A blank application template',
-  'app-drawer-template': 'A starter application template, with navigation and "PRPL pattern" loading',
-  'shop': 'The "Shop" Progressive Web App demo',
-};
-
-interface GeneratorDescription {
-  name: string;
-  value: string;
-  short: string;
-}
+let logger = logging.getLogger('cli.command.init');
 
 export class InitCommand implements Command {
   name = 'init';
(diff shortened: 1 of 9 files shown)


src/commands/init.ts
```diff
@@ -9,30 +9,18 @@
  */
 
 import {Command} from './command';
-import {execSync} from 'child_process';
 import {ArgDescriptor} from 'command-line-args';
-import * as fs from 'fs';
 import * as logging from 'plylog';
+<<<<<<< 3a65064217522ab46f3441533e53f6dae58782d3
```
justinfagnani: Holy unmerged files Batman!

FredKSchott: holy moly

FredKSchott: done

test/init/init_test.js
```diff
@@ -0,0 +1,162 @@
...
+        done(new Error('The promise should have been rejected by now'));
+      }).catch((err) => {
+        assert.equal(err.message, `Template ${UNKNOWN_GENERATOR_NAME} not found`);
+        done();
+      });
+    });
+
+  });
+
+  suite('promptGeneratorSelection', () => {
```
justinfagnani: It's nice that `createSelectPrompt` is split out now, but you don't test it directly. Most of the tests here can be directly against that function, and they'll likely be easier to read.

test/init/init_test.js
```diff
@@ -0,0 +1,162 @@
...
+    });
+
+    test('fails if an unknown generator is requested', (done) => {
+      const UNKNOWN_GENERATOR_NAME = 'UNKNOWN-GENERATOR';
+      let yeomanEnv = createFakeEnv();
+      yeomanEnv.getGeneratorsMeta.returns({
+        'TEST-GENERATOR': 'TEST-GENERATOR',
+      });
+
+      polymerInit.runGenerator(UNKNOWN_GENERATOR_NAME, {env: yeomanEnv}).then(() => {
```
justinfagnani: I think you can do:

```
return polymerInit.runGenerator(...).then(() {
  throw new Error(...);
}, (error) {
  assert.equal(error.message, ...);
});
```

to invert the Promise resolve/reject state and check for a specific message.

FredKSchott: Oh nice, I like that pattern much better

test/init/init_test.js
FredKSchott: I really liked your idea to split it out, but I'd rather keep testing the public interface here. While we could export `createSelectPrompt()` and make it public, we'd lose a lot of the test coverage of `promptGeneratorSelection()` by now assuming that it's internally calling & using `createSelectPrompt()` properly.

justinfagnani: I'm not saying remove all coverage of `promptGeneratorSelection()`, but right now there's exactly one branch in that function: the `||` used to create an environment, which is only there for testing :) So you don't need multiple tests to cover it (unless you want to test `createYeomanEnvironment()` and `createSelectPrompt()` throwing, which is too much testing at this point). The tests directly against `createSelectPrompt` should be simpler and easier to read and maintain.

It's not a huge deal here - the tests are written and there's only 5 of them, but I like making the actual code easier to test by factoring out relatively pure implementation functions invoked by simpler function that just do the I/O, as running all the permutations through the I/O functions is usually a pain.

FredKSchott: Totally understand that all the same current behavior *could* be tested via `createSelectPrompt()` directly by making it public, but I've seen internal implementations change enough times that I just always prefer to test the public interface whenever possible even if it's a little harder to write.

Sounds like it's just a difference of opinion :) And I think you're right, with so few tests here this is more of a conceptual discussion than anything. I'm happy to make these change if you'd like me too.

test/init/init_test.js
FredKSchott: done!

PR #260, opened 2016-06-10
assign element using `var` instead of `let`

woutervroege:
`let` makes non ecma script 6 compliant browsers ignore the test (tested using Safari Version 9.1.1)


woutervroege:
I signed it!

FredKSchott:
Great catch, thanks @woutervroege!

PR #270, opened 2016-06-15
Accept relative roots in project config.

cdata:
Most file paths in the project config are resolved to their absolute
paths by the time they are passed to `build`. For some reason, this is
not true for `root`. This change resolves the configured `root` to its
absolute path, so that [a later check](https://github.com/Polymer/polymer-cli/blob/master/src/build/url-from-path.ts#L43-L45) comparing root to other file paths
will succeed when appropriate.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
   },
   "devDependencies": {
     "chai": "^3.5.0",
+    "del": "^2.2.0",
     "depcheck": "^0.6.3",
     "gulp-eslint": "^2.0.0",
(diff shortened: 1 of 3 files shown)


cdata:
It seems likely that there is a more appropriate place to sanitize the config. Please let me know how this proposed change could be improved!

cdata:
After discussion with @justinfagnani it sounds like I misinterpreted the intention of `root`. It is probably broken, but also this PR assumes it means something that it doesn't.

PR #274, opened 2016-06-15
github@1.1.0

FredKSchott:
0.2.4 -> 1.1.0, comes with a small interface change

/cc @justinfagnani

### DIFF
--- a/custom_typings/github.d.ts
+++ b/custom_typings/github.d.ts
@@ -31,13 +31,13 @@ declare module "github" {
   }
   interface GetReleaseMessage {
     headers?: {string: string};
-    owner: string;
+    user: string;
     repo: string;
     id: number;
   }
   interface ListReleasesMessage {
     headers?: {string: string};
-    owner: string;
+    user: string;
     repo: string;
     page?: number;
     per_page?: number;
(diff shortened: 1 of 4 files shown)


FredKSchott:
Fun fact, `v1.0.0` was never published to npm. v1 starts at `v1.1.0`. But yes, it looks like we do work with v1.1.0, dropping down now.

As a small aside, it looks like the library is under very active development and the author isn't _really_ following semver. Something to keep in mind if we run into any sudden, unexpected github issues.

justinfagnani:
Only kind of following semver? On no. Well, maybe we keep it at ^1.1.0 then? Your call here.

FredKSchott:
I saw a few minor changes that were included in patch updates, so the only way to get around that is to pin to a specific version which I think is overkill here. `^v1.1.0` should be fine here

I just realized that a shrinkwrap file would make all of these optimizations moot anyway. We'd end up pinning all versions to their latest (or whatever was on our machine). If we're still planing on adding that, then there's not much benefit to being general and we should aim for the most stable version.

justinfagnani:
I don't think shrinkwrap invalidates this. Having the widest version range possible allows npm3 to do the maximally flat install, then shrinkwrapping that will preserve the versions.

FredKSchott:
Ah, great point. I was thinking about how globally installed dependencies are shared, but you're right there's benefits even to just our own dependency tree.

PR #280, opened 2016-06-16
gulp-typings@2.0.0

FredKSchott:
_This concludes our dependencies update process. We hope you have enjoyed the experience, and look forward to updating again real soon._ 

/cc @justinfagnani

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -18,8 +18,7 @@
     "src/**/*.ts",
     "!node_modules/**",
     "custom_typings/*.d.ts",
-    "typings/main/*.d.ts",
-    "typings/main.d.ts",
+    "typings/index.d.ts",
     "node_modules/typescript/lib/lib.core.es6.d.ts"
   ],
   "files": [
@@ -62,7 +61,7 @@
     "custom_typings/web-component-tester.d.ts",
     "custom_typings/yeoman-environment.d.ts",
     "custom_typings/yeoman-generator.d.ts",
-    "typings/main.d.ts",
+    "typings/index.d.ts",
     "node_modules/typescript/lib/lib.core.es6.d.ts"
   ],
   "atom": {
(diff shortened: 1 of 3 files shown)


justinfagnani:
Doesn't this require an update to `tsconfig.json` to load `typings/index.d.ts` instead of `typings/main.d.ts`?

FredKSchott:
The only change in `gulp-typings` is an update to the internal `typings-core` library version. The major change that was included in that update is here: https://github.com/typings/core/releases/tag/v1.0.0

Nothing in that list stood out as problematic to me, and typings seems to run just as it did before. But I'd love a sanity check, does anything stand out to you?

justinfagnani:
Yes, the thing that stands out is that we still have this in `tsconfig.json`:

```
  "filesGlob": [
    "typings/main/*.d.ts",
    "typings/main.d.ts",
```

while typings 1.0 uses `typings/index.d.ts`.

Are you sure that `gulp init && gulp build` works with this change?

FredKSchott:
1 sec, I'll do a clean npm install and re-run just to make sure.

FredKSchott:
ah ha! `gulp clean` wasn't deleting the old typings directory, so `gulp build` was using the old files found in there. Deleting the typings directory and re-running caused the error you expected: 

```
error TS6053: File 'typings/main.d.ts' not found.
```

Great catch, fixing now

FredKSchott:
@justinfagnani PTAL

tsconfig.json
```diff
@@ -18,8 +18,7 @@
     "src/**/*.ts",
     "!node_modules/**",
     "custom_typings/*.d.ts",
-    "typings/main/*.d.ts",
```
FredKSchott: Including each definition file itself seems to be unnecessary, since the exact file paths are written as references to `typings/index.d.ts`

justinfagnani: agreed, I think there was something about old typings that recommended this.

PR #310, opened 2016-07-21
Move build stream to use polymer-build

FredKSchott:
~~*Note: Currently using a github tag instead of npm in package.json for polymer-build. Will be removed once package is updated on npm*~~

This moves our current build logic over to the polymer-build pipeline. In addition to the much more efficient build pipeline, polymer-build should introduce several bug fixes and bring us back to maintaining a single build system.

/cc @justinfagnani


FredKSchott:
@justinfagnani PTAL, actively using polymer-build in the CLI should be useful for understanding its usage

src/build/build.ts
```diff
@@ -16,86 +16,48 @@ import * as gutil from 'gulp-util';
...
-    foundGulpfileDir = findup.sync(process.cwd(), 'gulpfile.js');
-  } catch (e) {
-    logger.debug(`no gulpfile found (searched up from ${process.cwd()})`);
-    return null;
-  }
-
-  let gulpfulePath = path.join(foundGulpfileDir, 'gulpfile.js');
-  logger.debug('found gulpfile', { path: gulpfulePath });
-  return require(gulpfulePath);
+export interface DepsIndex {
```
justinfagnani: Doesn't this interface exist in polymer-build?

src/build/build.ts
```diff
@@ -16,86 +16,48 @@ import * as gutil from 'gulp-util';
...
-      '!bower_components/**/*',
-      '!node_modules',
-      '!node_modules/**/*',
-    ];
-
-    logger.debug(`shell: ${shell}`);
-    logger.debug(`entrypoint: ${entrypoint}`);
-    logger.debug(`dependencies: ${dependencies}`);
+    let polymerProject = new PolymerProject({
+      root: config.root,
```
justinfagnani: config can be null

src/build/build.ts
```diff
@@ -120,102 +82,95 @@ export function build(options?: BuildOptions, config?: ProjectConfig): Promise<a
...
 
-    let unbundledPhase = new ForkedVinylStream(allFiles)
+    let unbundledPhase = forkStream(buildStream)
       .once('data', () => { logger.info('Generating build/unbundled...'); })
       .pipe(
         gulpif(
           options.insertDependencyLinks,
-          new PrefetchTransform(root, entrypoint,
-                shell, fragments, analyzer)
+          new PrefetchTransform(polymerProject.root, polymerProject.entrypoint,
```
justinfagnani: It might be nice if things like this just depended on PolymerProject, or a project config

justinfagnani:
LGTM

src/build/build.ts
FredKSchott: I intentionally haven't touched this in this PR, but totally agree on this point.

src/build/build.ts
FredKSchott: 👍  good catch

src/build/build.ts
FredKSchott: done

PR #311, opened 2016-07-21
Remove optimization abstraction, use tools in pipeline instead

FredKSchott:
*Note: this builds off of the polymer-build PR #310, address/resolve that and then rebase before reviewing*

This PR continues to remove unnecessary code from polymer-cli now that we have polymer-build pipeline. This removes the optimization & uglify streams, which both encapsulated different optimizers. We can add these tools directly into the pipeline, removing complexity/maintenance.

/cc @justinfagnani


justinfagnani:
nice!

LGTM

Issue #315, opened 2016-07-27
Error code return as 0 on failed polymer test run


pmaudsley:
### Description
I am running test on custom elements using the Polymer CLI on a build server. The build server is setup to report a successful build on an exit code of 0 and failed builds on an exit code of !0. Currently Polymer CLI does not return any codes on a error which is resulting in the build I run to be successful all the time, even on failures. 

I will put a PR adding process.exit(1) on a error from cli.run()

### Versions & Environment
- Polymer CLI: 0.12.0
- node: v6.3.1

#### Steps to Reproduce

1. Create an test element: `polymer init element`
2. Run test: `polymer test`
3. Check exit code: `echo $?`
4. Edit test to result in error: change assert.equal(element.is, 'test-element') to assert.equal(element.is, 'foo');
5. Re-run test: `polymer test`
6. Check exit code: `echo $?`

#### Expected Results
Error code to be 0 on successful test and error code to be !0 on a failed test.

#### Actual Results
Error code is 0 on successful test and on a failed test.

FredKSchott:
Fixed in #316 and released in v0.13.0!

PR #316, opened 2016-07-27
Send an error code on polymer command run error

pmaudsley:
PR request to solve issue https://github.com/Polymer/polymer-cli/issues/315

Adding a process.exit(1) on a error from cli.run() to prevent error code 0 returned for failed polymer test runs

### DIFF
--- a/bin/polymer.js
+++ b/bin/polymer.js
@@ -16,6 +16,9 @@ process.title = 'polymer';
 const resolve = require('resolve');
 const updateNotifier = require('update-notifier');
 const packageJson = require('../package.json');
+const logging = require('plylog');
+
+const logger = logging.getLogger('cli.main');
 
 // Update Notifier: Asynchronously check for package updates and, if needed,
 // notify on the next time the CLI is run.
@@ -26,6 +29,9 @@ resolve('polymer-cli', {basedir: process.cwd()}, function(error, path) {
   let lib = path ? require(path) : require('..');
   let args = process.argv.slice(2);
   let cli = new lib.PolymerCli(args);
-  cli.run()
-  .then(null, () => process.exit(1));
+  cli.run().catch(err => {
+    logger.error(err);
+    if (err.stack) logger.error(err.stack);
+    process.exit(1);
+  });
 });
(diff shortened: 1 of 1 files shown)


bin/polymer.js
```diff
@@ -26,5 +26,6 @@ resolve('polymer-cli', {basedir: process.cwd()}, function(error, path) {
   let lib = path ? require(path) : require('..');
   let args = process.argv.slice(2);
   let cli = new lib.PolymerCli(args);
-  cli.run();
+  cli.run()
+  .then(null, () => process.exit(1));
```
mattyclarkson: You can use `catch` rather than the second parameter of `then`:

```
cli.run().catch(() => process.exit(1));
```

FredKSchott: +1, using `then()` like this is fine, but we use `catch()` throughout the rest of the project. Lets stick to that here.

Also, can you please log a message before exit? Check out the `process.on('unhandledRejection', ...` handler in `polymer-cli.ts` for an example of what you'll need.

FredKSchott:
@pmaudsley thanks for the PR! I can't believe we missed this, thanks for adding.

abdonrd:
Thanks for the PR!

pmaudsley:
I've made the changes as suggested. 

Is it OK that this logger repeats the previous logs i.e.:

```
1 failed tests

error:   1 failed tests
```

FredKSchott:
@pmaudsley where are you seeing that "1 failed tests" error message? Do you have steps to reproduce so I can understand better?

bin/polymer.js
```diff
@@ -26,5 +29,9 @@ resolve('polymer-cli', {basedir: process.cwd()}, function(error, path) {
   let lib = path ? require(path) : require('..');
   let args = process.argv.slice(2);
   let cli = new lib.PolymerCli(args);
-  cli.run();
+  cli.run().catch(err => {
+    logger.error(err);
```
FredKSchott: Can you include a message here? It's helpful in debugging to know where the error was caught & logged. See polymer-cli.ts `uncaughtRejection` handler for example.

FredKSchott:
Superseded by #321

Issue #317, opened 2016-07-27
Element template has incorrect path for polymer.html import


catiejojo:
### Description
Relative path for polymer.html is wrong, so element template doesn't build out of the box.

### Issue
_These aren't relevant to the issue, but just to be thorough..._
- Polymer CLI: 0.12
- Node: 4.4.5
- Operating System: Mac OSX El Capitan 10.11.5

#### Steps to Reproduce
1. Create an element: `polymer init element`
2. Build: `polymer build`

#### Results
```
error:   Promise rejection: Error: no resolver found for ../polymer/polymer.html
error:   Error: no resolver found for ../polymer/polymer.html
    at FileLoader.request (/usr/local/lib/node_modules/polymer-cli/node_modules/hydrolysis/lib/loader/file-loader.js:70:37)
    at Analyzer.load (/usr/local/lib/node_modules/polymer-cli/node_modules/hydrolysis/lib/analyzer.js:121:32)
    at Analyzer._getDependencies (/usr/local/lib/node_modules/polymer-cli/node_modules/hydrolysis/lib/analyzer.js:433:25)
    at Analyzer._dependenciesLoadedFor (/usr/local/lib/node_modules/polymer-cli/node_modules/hydrolysis/lib/analyzer.js:401:25)
    at Analyzer._parseHTML (/usr/local/lib/node_modules/polymer-cli/node_modules/hydrolysis/lib/analyzer.js:227:50)
    at null._onTimeout (/usr/local/lib/node_modules/polymer-cli/node_modules/hydrolysis/lib/analyzer.js:125:39)
    at Timer.listOnTimeout (timers.js:92:15)
```

### Fix
I think the files in this project may have gotten reorganized without the paths getting updated. If you go to line one of the custom-element.html file, the polymer import path is "../polymer/polymer.html". However, that file is actually under the bower_components folder, so the line should be `<link rel="import" href="bower_components/polymer/polymer.html">
`. Making this change and rebuilding it fixes the error.

jfletcher-nuxeo:
This is normal. It's explained here: https://github.com/Polymer/polymer-cli/issues/222#issuecomment-223631794

The idea is that in a deployed environment all the elements exist as siblings so relative paths need to be used.

catiejojo:
[...]

PR #318, opened 2016-07-27
Fix for issue #317

catiejojo:
This template doesn't currently work out of the box because polymer.html isn't stored at "../polymer/polymer.html" as the template says, it's stored at "bower_components/polymer/polymer.html". I filed issue #317 as a result of this.


catiejojo:
This is my first commit, so just signed the CLA. Let me know if I can provide any clarification.

justinfagnani:
Thanks for the PR, but this isn't a correct change. For reusable elements, the correct path for dependencies is `../other-package-name/`

Seems like this was covered in the issue, so I'll close this PR now.

PR #319, opened 2016-07-27
Use polymer-build library for service-worker generation

FredKSchott:
/cc @justinfagnani


custom_typings/polymer-build.d.ts
```diff
@@ -76,6 +76,8 @@ declare module 'polymer-build' {
   }
 
   export interface AddServiceWorkerOptions {
+    project: PolymerProject;
```
justinfagnani: we should be pulling these typings out of polymer-build. polymer-build needs a `"declaration": true` line in its tsconfig

justinfagnani:
LGTM

custom_typings/polymer-build.d.ts
FredKSchott: Oh cool, I didn't know about this feature. 

I have something else relying on this that I want to get into PR, so I'm going to merge now but will look into this.

Issue #305, opened 2016-07-14
Cannot read property 'previous' of null when running polymer build


Davis22xd:
### Description

When I was trying to build my polymer project with polymer build I got the error:
cannot read property 'previous' of null
### Versions & Environment

- Polymer CLI: 0.11.1
- node: v4.2.1
- Operating System: Ubuntu 14.04

#### Steps to Reproduce

1. I create my project that I have in:
**https://github.com/Davis22xd/polymer-table**
2. checkout to origin/simple-table branch
3. I try to build the project with polymer build

#### Expected Results

vulcanized file 

#### Actual Results

info:    Building application...
warn:    Unable to uglify file /home/sumitrag-david/Documents/Proyectos/polymer-table/polymer-table.html_script_0.js
info:    Generating build/unbundled...
info:    Generating build/bundled...
error:   Uncaught exception: TypeError: Cannot read property 'previous' of null
error:   TypeError: Cannot read property 'previous' of null
    at Parser.parseDeclarationOrRuleset (/usr/local/lib/node_modules/polymer-cli/node_modules/shady-css-parser/dist/shady-css/parser.js:192:63)
    at Parser.parseRule (/usr/local/lib/node_modules/polymer-cli/node_modules/shady-css-parser/dist/shady-css/parser.js:80:21)
    at Parser.parseRules (/usr/local/lib/node_modules/polymer-cli/node_modules/shady-css-parser/dist/shady-css/parser.js:62:25)
    at Parser.parseStylesheet (/usr/local/lib/node_modules/polymer-cli/node_modules/shady-css-parser/dist/shady-css/parser.js:55:47)
    at Parser.parse (/usr/local/lib/node_modules/polymer-cli/node_modules/shady-css-parser/dist/shady-css/parser.js:51:19)
    at css (/usr/local/lib/node_modules/polymer-cli/node_modules/css-slam/index.js:51:39)
    at GulpTransform._transform (/usr/local/lib/node_modules/polymer-cli/node_modules/css-slam/index.js:69:36)
    at GulpTransform.Transform._read (_stream_transform.js:167:10)
    at GulpTransform.Transform._write (_stream_transform.js:155:12)
    at doWrite (_stream_writable.js:292:12)

oneezy:
Any luck on this @Davis22xd ?

I created duplicate around exact same time you did.. issue #306

Davis22xd:
No yet, but I'm looking for it.  I hope to find something

[...]

Issue #297, opened 2016-06-29
Error message during build should be more explicit


NicolasRannou:
### Description
Error message during build should be more explicit

#### Expected Results

It should tell us which file is problematic.

#### Actual Results
```
error:   Uncaught exception: TypeError: Cannot read property 'is' of null
error:   TypeError: Cannot read property 'is' of null
    at Parser.parseDeclarationOrRuleset (/usr/local/lib/node_modules/polymer-cli/node_modules/shady-css-parser/dist/shady-css/parser.js:191:33)
    at Parser.parseRule (/usr/local/lib/node_modules/polymer-cli/node_modules/shady-css-parser/dist/shady-css/parser.js:80:21)
    at Parser.parseRules (/usr/local/lib/node_modules/polymer-cli/node_modules/shady-css-parser/dist/shady-css/parser.js:62:25)
    at Parser.parseStylesheet (/usr/local/lib/node_modules/polymer-cli/node_modules/shady-css-parser/dist/shady-css/parser.js:55:47)
    at Parser.parse (/usr/local/lib/node_modules/polymer-cli/node_modules/shady-css-parser/dist/shady-css/parser.js:51:19)
    at css (/usr/local/lib/node_modules/polymer-cli/node_modules/css-slam/index.js:51:39)
    at dom5.queryAll.forEach.styleNode (/usr/local/lib/node_modules/polymer-cli/node_modules/css-slam/index.js:45:36)
    at Array.forEach (native)
    at html (/usr/local/lib/node_modules/polymer-cli/node_modules/css-slam/index.js:43:37)
    at GulpTransform._transform (/usr/local/lib/node_modules/polymer-cli/node_modules/css-slam/index.js:66:36)
```

Is it possible to add this kind of information with the current infrastructure?

mikael2933:
Check files through, I think you forgot a `,` or you added one too many.

NicolasRannou:
Yes - I found the issue it was a typo in the css as the error message suggested but I would hope polymer-cli could tell us directly which file.

In general it is fine to track it down if you commit small portions of code and run CI on every commit though.

FredKSchott:
Fixed in [v0.13.0](https://github.com/Polymer/polymer-cli/releases/tag/v0.13.0)

FredKSchott:
Fixed in [v0.13.0](https://github.com/Polymer/polymer-cli/releases/tag/v0.13.0)

Issue #268, opened 2016-06-14
Add file information to error messages


vaadin-marcus:
I'm building a bundled project with `polymer build` on 0.11.1 and it fails with the following error:

```
error:   Uncaught exception: TypeError: Cannot read property 'is' of null
error:   TypeError: Cannot read property 'is' of null
    at Parser.parseDeclarationOrRuleset (/Users/mhellber/.nvm/versions/node/v6.0.0/lib/node_modules/polymer-cli/node_modules/shady-css-parser/dist/shady-css/parser.js:191:33)
    at Parser.parseRule (/Users/mhellber/.nvm/versions/node/v6.0.0/lib/node_modules/polymer-cli/node_modules/shady-css-parser/dist/shady-css/parser.js:80:21)
    at Parser.parseRules (/Users/mhellber/.nvm/versions/node/v6.0.0/lib/node_modules/polymer-cli/node_modules/shady-css-parser/dist/shady-css/parser.js:62:25)
    at Parser.parseStylesheet (/Users/mhellber/.nvm/versions/node/v6.0.0/lib/node_modules/polymer-cli/node_modules/shady-css-parser/dist/shady-css/parser.js:55:47)
    at Parser.parse (/Users/mhellber/.nvm/versions/node/v6.0.0/lib/node_modules/polymer-cli/node_modules/shady-css-parser/dist/shady-css/parser.js:51:19)
    at css (/Users/mhellber/.nvm/versions/node/v6.0.0/lib/node_modules/polymer-cli/node_modules/css-slam/index.js:51:39)
    at GulpTransform._transform (/Users/mhellber/.nvm/versions/node/v6.0.0/lib/node_modules/polymer-cli/node_modules/css-slam/index.js:69:36)
    at GulpTransform.Transform._read (_stream_transform.js:167:10)
    at GulpTransform.Transform._write (_stream_transform.js:155:12)
    at doWrite (_stream_writable.js:306:12)
    at writeOrBuffer (_stream_writable.js:292:5)
```

It would be very helpful for debugging if the error message could contain information on the file that failed similar to warning messages. I'm sure the actual error is in my code or it's dependencies, but it would be very helpful to get more debug info.

justinfagnani:
By any chance do you have a repro for this so we can make sure we get the file info here?

vaadin-marcus:
Here you go, just run `polymer build` on this. 

[...]

Issue #219, opened 2016-05-27
errors and warnings from `polymer build`


davidmaxwaterman:
I have attempted to use the polymer-cli to start an `application: A blank application template`. The init process works, but `polymer build` produces several warnings and errors :

```
[sc-toolbox]$ which polymer
/home/davidmaxwaterman/.nvm/versions/node/v6.2.0/bin/polymer
[sc-toolbox]$ ls -l $(which polymer)
lrwxrwxrwx 1 davidmaxwaterman davidmaxwaterman 46 May 27 11:51 /home/davidmaxwaterman/.nvm/versions/node/v6.2.0/bin/polymer -> ../lib/node_modules/polymer-cli/bin/polymer.js
[sc-toolbox]$ ls -i /home/davidmaxwaterman/.nvm/versions/node/v6.2.0/bin/../lib/node_modules/polymer-cli/bin/polymer.js
11885941 /home/davidmaxwaterman/.nvm/versions/node/v6.2.0/bin/../lib/node_modules/polymer-cli/bin/polymer.js
[sc-toolbox]$ ls -i ~/z/polymer-cli/bin/polymer.js 
11885941 /home/davidmaxwaterman/z/polymer-cli/bin/polymer.js
[sc-toolbox]$ (cd ~/z/polymer-cli/ && git log | head)
commit 1e4a6841be3b9f1d48be918074c6516dc959adf4
Author: Fred K. Schott <fschott@google.com>
Date:   Thu May 26 15:37:31 2016 -0700

    0.10.0

commit c510b1140318319ac5d3cfef6b49b9da28b371fd
Merge: 7635a5b bf6556d
Author: Justin Fagnani <justin@fagnani.com>
Date:   Thu May 26 14:51:27 2016 -0700
[sc-toolbox]$ ls
[sc-toolbox]$ polymer init
? Which starter template would you like to use? application
info:    Running template application
? Application name Scientific Calculator
? Main element name o1-webapps-scientific-calculator
? Brief description of the application A Scientific Calculator.
   create bower.json
   create index.html
   create manifest.json
   create README.md
   create src/o1-webapps-scientific-calculator/o1-webapps-scientific-calculator.html
   create test/o1-webapps-scientific-calculator/o1-webapps-scientific-calculator_test.html

Project generated!
Installing dependencies...


I'm all done. Running bower install for you to install the required dependencies. If this fails, try running the command yourself.


[...]

PR #320, opened 2016-07-27
Add back custom optimize streams that will warn on error

### LINKED ISSUE #307: warn: Unable to uglify files
Node: 4.2.6
Polymer-CLI: 0.12.0

How can I remove the "Unable to uglify file ..." warning?
What's the exact solution for this?

image

FredKSchott:
Using the community gulp libraries works great for projects with no errors, but we just realized that they all throw fatal errors when a parse error is encountered. This is less of a problem now with the new build stream, but it can still be a jarring user experience when you're on the bleeding-edge like us. 😎

Now, they all print warning to the console but the build continues.

Along with the new polymer-build stream, this should fix #307 #305 #297 #268 #219

/cc @justinfagnani


src/build/build.ts
```diff
@@ -77,9 +75,9 @@ export function build(options: BuildOptions, config: ProjectConfig): Promise<any
     logger.debug(`Reading source files...`);
     let sourcesStream = polymerProject.sources()
       .pipe(polymerProject.splitHtml())
-      .pipe(gulpif(/\.js$/, uglify(optimizeOptions.js)))
-      .pipe(gulpif(/\.css$/, cssSlam(optimizeOptions.css)))
-      .pipe(gulpif(/\.html$/, htmlmin(optimizeOptions.html)))
+      .pipe(gulpif(/\.js$/, new JSOptimizeStream(optimizeOptions.js)))
```
justinfagnani: could we do the same thing with a generic transform wrapper that fails gracefully?

FredKSchott: Good call, I'll look into it now.

FredKSchott: @justinfagnani Check out the new `optimize-streams.ts` and let me know what you think. Personally I prefer the extra code vs. inheritance here, but this definitely does have better code reuse.

justinfagnani: I was thinking something like: `.pipe(gulpif(/\.js$/, passThroughOnFail(uglify()))`

FredKSchott: It's not going to work :( And I really wanted it to.

There are two big problems in the way:

1. Lots of streams end up emitting an error when one is encountered. The only thing that is sent with the event is an error object. This means we have no reliable way to know which file failed, which we need to know so that we can pass through the original. No level of monkey patching can generally solve for this.
2. There is no consistency with how these streams are implemented. Some return an error to the `_transform` callback, some weren't even written with the `_transform` callback at all and instead rely on the `_write` callback, some do neither and just emit an error, some (like gulp-html-minifier') just throw an error, etc. etc. etc. Because of all the different changes Node streams have gone through over the years, a lot of these are all valid implementations. So even if #1 wasn't an issue, it would still be incredibly hard to create a general wrapper.

BUT what I have here is almost exactly what we want. Instead of wrapping a stream it wraps the actual libraries we want to use. This grants the user wayyyy more flexibility into how the library is called and how its options are consumed. Check out the implementation of each of these, and you'll see they're just small wrappers around `GenericOptimizeStream`. This solution also has a smaller dependency footprint (1 less level of dependencies, see [html-minifier](http://npm.anvaka.com/#/view/2d/html-minifier) vs. [gulp-html-minifier](http://npm.anvaka.com/#/view/2d/gulp-html-minifier))

WDYT? If we like, we can create a factory similar to `passThroughOnFail` for `GenericOptimizeStream` (takes its constructor arguments and returns a new stream).

FredKSchott: Also check out https://github.com/hughsk/vinyl-transform & [the relevant blog post it mentions](http://blog.overzealous.com/post/74121048393/why-you-shouldnt-create-a-gulp-plugin-or-how-to). This should be a fine solution for our users (and the best part is it already exists :)

src/build/optimize-streams.ts
```diff
@@ -0,0 +1,110 @@
...
+    this.optimizerOptions = optimizerOptions || {};
+  }
+
+  _transform(file: File, encoding: string, callback: FileCB): void {
+    if (file.contents && file.path.endsWith(`${this.validExtension}`)) {
+      try {
+        let contents = file.contents.toString();
+        contents = this.optimizer(contents, this.optimizerOptions);
+        file.contents = new Buffer(contents);
+      } catch (err) {
```
justinfagnani: naming nit: use `error` instead of `err`

justinfagnani:
LGTM

src/build/optimize-streams.ts
FredKSchott: Good to know, I've always used `err`.  The library  is using `err`, `error`, and even `e` all over the place, I'll change this here and then create a separate PR to get the rest of the repo on board.

PR #330, opened 2016-08-01
replace unneccesary gulp dependency with vinyl-fs

FredKSchott:
With the new build pipeline, gulp is unneccesary as a regular dependency. Instead, we can use vinyl-fs directly and greatly reduce our dependency graph. (internally, `gulp.dest` is just a direct reference to `vinyl-fs.dest`)

See http://npm.anvaka.com/#/view/2d/gulp (212) vs. http://npm.anvaka.com/#/view/2d/vinyl-fs (89)

/cc @justinfagnani

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,6 @@
     "findup": "^0.1.5",
     "fs-extra": "^0.30.0",
     "github": "^1.1.0",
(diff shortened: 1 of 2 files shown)


justinfagnani:
LGTM

PR #334, opened 2016-08-02
return resolved promise on version flag

### LINKED ISSUE #333: --version flag throws an error
Introduced in v0.13.0

Seems like a perfect storm of problems caused this:

1. Typescript didn't catch that undefined was being returned here, even when the implicit type is added. @justinfagnani is that a special behavior of the Promise type that it thinks it's in a `then()` style Promise chain?
2. Testing was covering the CLI directly, and not checking it's return type or the `/bin` file.

Good news is the version is still output, just with the additional error.

FredKSchott:
Fixes #333
Also adds implicit return type to run() function, plus tests.

/cc @justinfagnani

### DIFF
--- a/src/polymer-cli.ts
+++ b/src/polymer-cli.ts
@@ -122,7 +122,7 @@ export class PolymerCli {
     return Array.from(mergedArgs.values());
   }
 
-  run() {
+  run(): Promise<any> {
     let helpCommand = this.commands.get('help');
     let commandNames = Array.from(this.commands.keys());
     let parsedArgs;
@@ -132,7 +132,7 @@ export class PolymerCli {
     // the current version. Useful for globally installed CLIs.
     if (this.args.indexOf('--version') > -1) {
       console.log(require('../package.json').version);
-      return;
+      return Promise.resolve();
     }
 
     try {
--- a/test/commands/cli_test.js
+++ b/test/commands/cli_test.js
@@ -25,59 +25,65 @@ suite('The general CLI', () => {
     let cli = new PolymerCli([]);
     let helpCommand = cli.commands.get('help');
     let helpCommandSpy = sinon.spy(helpCommand, 'run');
-    cli.run();
-    assert.isOk(helpCommandSpy.calledOnce);
-    assert.deepEqual(helpCommandSpy.firstCall.args, [{command: null}, defaultConfig]);
+    return cli.run().then(() => {
+      assert.isOk(helpCommandSpy.calledOnce);
+      assert.deepEqual(helpCommandSpy.firstCall.args, [{command: null}, defaultConfig]);
+    });
   });
 
   test('displays general help when no command is called with the --help flag', () => {
     let cli = new PolymerCli(['--help']);
     let helpCommand = cli.commands.get('help');
     let helpCommandSpy = sinon.spy(helpCommand, 'run');
-    cli.run();
-    assert.isOk(helpCommandSpy.calledOnce);
-    assert.deepEqual(helpCommandSpy.firstCall.args, [{command: null}, defaultConfig]);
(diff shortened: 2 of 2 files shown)


test/commands/cli_test.js
```diff
@@ -25,58 +25,64 @@ suite('The general CLI', () => {
     let cli = new PolymerCli([]);
     let helpCommand = cli.commands.get('help');
     let helpCommandSpy = sinon.spy(helpCommand, 'run');
-    cli.run();
+    let cliRunResult = cli.run();
```
justinfagnani: this would be safer as:

```
return cli.run().then((_) => {
  assert.isOk(...);
});
```

FredKSchott: 👍  good call

FredKSchott:
@justinfagnani PTAL

test/commands/cli_test.js
```diff
@@ -17,67 +17,79 @@ const logging = require('plylog');
 const assert = require('chai').assert;
 const sinon = require('sinon');
 
-suite('The general CLI', () => {
+suite.only('The general CLI', () => {
 
   const defaultConfig = new Config();
 
-  test('displays general help when no command is called', () => {
+  test('displays general help when no command is called', (done) => {
```
justinfagnani: don't use `done`, just return a `Promise`

test/commands/cli_test.js
```diff
@@ -17,67 +17,79 @@ const logging = require('plylog');
 const assert = require('chai').assert;
 const sinon = require('sinon');
 
-suite('The general CLI', () => {
+suite.only('The general CLI', () => {
```
justinfagnani: remove `.only`

FredKSchott: woof, good catch

test/commands/cli_test.js
FredKSchott: huh, for some reason I thought mocha didn't support Promises natively. TIL

FredKSchott:
@justinfagnani PTAL

src/polymer-cli.ts
```diff
@@ -122,7 +122,7 @@ export class PolymerCli {
     return Array.from(mergedArgs.values());
   }
 
-  run() {
+  run(): Promise<any> {
```
rictic: When returning a promise we really want to use an async function to be sure that all returns and all errors are propagated consistently

justinfagnani: Agreed. @FredKSchott and I talked about doing a big conversion of cli and polymer-build to async/await

test/commands/cli_test.js
```diff
@@ -25,59 +25,65 @@ suite('The general CLI', () => {
...
   });
 
   test('displays general help when no command is called with the --help flag', () => {
     let cli = new PolymerCli(['--help']);
     let helpCommand = cli.commands.get('help');
     let helpCommandSpy = sinon.spy(helpCommand, 'run');
-    cli.run();
-    assert.isOk(helpCommandSpy.calledOnce);
-    assert.deepEqual(helpCommandSpy.firstCall.args, [{command: null}, defaultConfig]);
+    cli.run().then(() => {
```
justinfagnani: `return cli.run().then(...`

src/polymer-cli.ts
rictic: Yeah, and it's an easy thing to add onto a function that already returns a Promise. The only differences in that case are that with async:

  - a default return becomes `Promise.resolve();`
  - a throw in the stack frame of the function become a rejected promise

FredKSchott: 👍 I'm going to look into this for the entire repo next week

justinfagnani:
LGTM

Issue #273, opened 2016-06-15
Feature request: configure effective web root


cdata:
I would like to be able to configure the Polymer CLI to treat a sub-directory of my project as the effective web root when building assets. Assuming the following contrived structure:

```
my-app
  |_ public
       |_ index.html
       |_ bower_components/some-dep/some-dep.html
```

And the following contrived contents of `index.html`:

```html
<!doctype html>
<html>
  <head>
    <link rel="import" href="/bower_components/some-dep/some-dep.html">
  </head>
</html>
```

I would like to configure my `polymer build` step so that it correctly treats the web root as `public` (not the root of my project directory).

FredKSchott:
See #277, this is an issue for people migrating from PSK

abdonrd:
Any news about this?

FredKSchott:
You can currently do this by setting a `root` property via CLI flags (`polymer.json` supports this property as well but there is currently a bug, see https://github.com/Polymer/polymer-build/issues/34). 

One constraint to keep in mind is that your `root` really is the "root" of your application, so anything above/outside of your root directory will be inaccessible. So no reaching outside of root for `bower_components`, `manifest.json`, etc.

cdata:
@FredKSchott When I discussed this issue with @justinfagnani , he assured me that `root` was misleading, and not actually a configuration intended to address the feature request above. Perhaps he can comment more on this..

FredKSchott:
I think we may have confused your request with needing to serve an application from a path other than the root URL. The way I understand the feature request, you'd like `public/` to be your web root instead of the project working directory, which is what `root` is supposed to let you do.

@robdodson had a similar request yesterday which we helped him work through, does this sound similar? https://polymer.slack.com/archives/ecosystem-eng/p1473273440000065

cdata:
@FredKSchott I don't think I have access to that Slack archive.

FredKSchott:
whoops, try now

cdata:
After trying to use `root` to address my use case, I can confirm that one of the two things must be true:

 1. `root` is broken
 2. `root` does not address the described use case

FredKSchott:
@cdata if you're reaching that conclusion from the investigation you posted in our Slack channel, I believe your issue exists somewhere else. We've just integrated the new analyzer into polymer-build, so once we hook up error handling you should be able to get a much better idea of what's failing (vs. the current silently failing horribleness).

abdonrd:
[...]

PR #336, opened 2016-08-03
Fix bad application template paths

FredKSchott:
Change absolute URL paths to relative ones so that build is happy & doesn't fail.

/cc @justinfagnani

### DIFF
--- a/src/init/application/templates/index.html
+++ b/src/init/application/templates/index.html
@@ -6,8 +6,8 @@
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <link rel="manifest" href="/manifest.json">
-    <script src="/bower_components/webcomponentsjs/webcomponents-lite.js"></script>
-    <link rel="import" href="/src/<%= elementName %>/<%= elementName %>.html">
+    <script src="bower_components/webcomponentsjs/webcomponents-lite.js"></script>
+    <link rel="import" href="src/<%= elementName %>/<%= elementName %>.html">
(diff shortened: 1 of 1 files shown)


abdonrd:
I remember @justinfagnani said the `index.html` should have absolute paths.

FredKSchott:
Ah, I hadn't thought of that. That makes sense (and works) for the app-shell model (where shell, entrypoint, etc are all defined in a polymer.json). Is that also true for an older monolith-style application?

src/init/application/templates/index.html
```diff
@@ -6,8 +6,8 @@
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <link rel="manifest" href="/manifest.json">
-    <script src="/bower_components/webcomponentsjs/webcomponents-lite.js"></script>
-    <link rel="import" href="/src/<%= elementName %>/<%= elementName %>.html">
+    <script src="bower_components/webcomponentsjs/webcomponents-lite.js"></script>
```
justinfagnani: These really should be absolute though. In a routable app, they need to load from the root.

FredKSchott: you're right, this isn't a solution. Closing to fix the problem in polymer-build instead!

web-devel:
@justinfagnani what if I want to serve an app from some context path i.e:
http://somedomain/app/...

FredKSchott:
@web-devel see https://github.com/Polymer/polymer-cli/issues/273

Issue #351, opened 2016-08-08
CLI doesn't use name/description for 


robdodson:
### Description
Creating a custom generator, the CLI doesn't display its description when I run `polymer init`

Generator repo:
https://github.com/PolymerElements/generator-polymer-init-custom-build

Screenshot:
![screen shot 2016-08-08 at 1 32 36 pm](https://cloud.githubusercontent.com/assets/1066253/17496472/5ffb3814-5d72-11e6-9f83-787143c7634e.png)

### Versions & Environment
node v6.3.0

#### Steps to Reproduce
Clone the generator
Run `npm link`
Run`polymer init`

#### Expected Results
Should see a description of the generator, maybe pulled from package.json?

#### Actual Results
No description

FredKSchott:
Run with verbose, you get:
```
debug:   problem reading/parsing package.json for generator 
{ generator: 'polymer-init-vaadin-charts-app:app',
  err: 'EISDIR: illegal operation on a directory, read' }
```

FredKSchott:
Looks like even `yo` doesn't get a description, it just "title-izes" the generator name.

```js
pkg.prettyName = titleize(humanizeString(namespaceToName(generator.namespace)));
```

Instead of implementing something similar that doesn't add any new info, lets just hide the description for custom generators. @justinfagnani @robdodson savvy?

justinfagnani:
I swear that at one point I had descriptions coming from `package.json`, I just can't recall what happened to it. Doesn't seem like it should be very difficult.

FredKSchott:
Here's where my globally installed generator lives

```
{ 
  resolved: '/Users/fschott/.nvm/versions/node/v5.10.1/lib/node_modules/generator-polymer-init-vaadin-charts-app/generators/app/index.js',
  namespace: 'polymer-init-vaadin-charts-app:app' 
}
```

FredKSchott:
I think the fact that yeoman doesn't do this is a good sign that we shouldn't either. I wasn't able to get it working but @addyosmani might have some good insight here.

justinfagnani:
I kind of remember looking into Yeoman and thinking it was just a bug, but it was so long ago that my memory is fuzzy.

FredKSchott:
In the interest of maximizing effort:results, I recommend merging fix #360 in the short term so that we get rid of that "undefined" now. We can then investigate bringing the description back when we have more time and potentially even help contribute a fix to the yeoman cli.

[...]

PR #360, opened 2016-08-10
stop showing undefined template description

FredKSchott:
A fix for #351 (see issue for background on fix)

/cc @justinfagnani @robdodson


robdodson:
LGTM! Not sure what's up with appveyor though :P

Issue #345, opened 2016-08-06
[bug] staticFileGlobs in sw-precache-config not process in build


AliMD:
### Description
`staticFileGlobs` in `sw-precache-config.js` not process in build and compiled `service-worker.js` not contain these files!

### Versions & Environment
- Polymer CLI: 0.13.0
- node: 6.3.1 and 4.4.7
- Operating System: OS X 10.11.6

#### Steps to Reproduce
1. edit `sw-precache-config.js`
```javascript
module.exports = {
  staticFileGlobs: [
    '/index.html',
    '/src/manifest.json',
    '/bower_components/webcomponentsjs/webcomponents-lite.min.js',
    '/src/images/*'
  ]
};
```
3. polymer build

#### Expected Results
compiled `service-worker.js` in PrecacheConfig array contain staticFileGlobs files
or in browser app work fully offline in second visit

#### Actual Results
app not work offline for staticFileGlobs files and images or manifest load every time from net or fail in offline mode!

Other auto detect files and components work offline

VandeurenGlenn:
Same here, `PrecacheConfig` contains all items defined in [polymer.json](https://github.com/Polymer/polymer-cli#polymerjson) butt doesn't contain the files defined in sw-precache-config.js

Eugenitals:
+1 and it is very critical.
My bundled version does not work in Firefox and Safari anymore (

perciag:
i had copy to folder build/unbundled manually, that's very bad.

veith:
Do not forget to copy the files before you run polymer build. Otherwise your service-worker.js will not contain the files, because  sw-precache checks if the files exists.

FredKSchott:
Oh yea, jeeze look at this (output from when I run `polymer build --verbose` in the shop template): 

```
{ staticFileGlobs: 
   [ 'build/bundled/build/unbundled/index.html',
     'build/bundled/build/unbundled/manifest.json',
     'build/bundled/build/unbundled/bower_components/webcomponentsjs/webcomponents-lite.min.js',
     'build/bundled/build/unbundled/images/*',
     'build/bundled/build/unbundled/src/shop-app.html',
     'build/bundled/build/unbundled/src/shop-list.html',
     'build/bundled/build/unbundled/src/shop-detail.html',
     'build/bundled/build/unbundled/src/shop-cart.html',
     'build/bundled/build/unbundled/src/shop-checkout.html',
     'build/bundled/build/unbundled/src/lazy-resources.html',
[...]

PR #372, opened 2016-08-22
[Issue #345] Copy configured globs from sw-precache-config.js

veith:
maybe there is a better way to do this.


veith:
node-modules/sw-precache/sw-precache.js checks if the file which is defined in params.staticFileGlobs (line 105) exists in build/bundled/ or build/unbundled/.

PR #391, opened 2016-09-02
Remove application-specific info from the README

kito99:
The text is misleading for working with elements. Not only is 'Application' confusing, but  `polymer build` doesn't work for them out of the box with this setup.


kito99:
I signed it!

FredKSchott:
Polymer CLI works with both elements and applications. If you are having trouble building/serving your app please reach out to us on [Slack](https://polymer-slack.herokuapp.com/) or file a [new issue](https://github.com/Polymer/polymer-cli/issues/new) and we can help you out.

PR #399, opened 2016-09-09
support relative root path in polymer.json

FredKSchott:
Previously, `root` only supported full system file paths. This made the option useless in a shared project with a shared polymer.json file. 

Now we support paths relative to the current working directory where the CLI is run.

/cc @robdodson @justinfagnani


robdodson:
LGTM

addyosmani:
LGTM2

PR #410, opened 2016-09-23
Convert project to typescript 2.0

FredKSchott:
Oh boy, lots of bugs found here. Still can't turn on `noImplicitAny` but this gets us closer.

### todo (future PRs):
- move all typings DT declarations to npm `@types`
- turn on `noImplicitAny`


custom_typings/chalk.d.ts
```diff
@@ -0,0 +1,25 @@
+declare namespace Chalk {
```
rictic: This exists at `@types/chalk` I think

tsconfig.json
```diff
@@ -1,67 +1,20 @@
 {
   "compilerOptions": {
     "target": "es6",
+    "lib": ["es2016"],
```
rictic: Recommend es2017

typings.json
```diff
@@ -2,8 +2,6 @@
   "globalDependencies": {
     "clean-css": "registry:dt/clean-css#3.4.9+20160316155526",
     "clone": "registry:dt/clone#0.1.11+20160317120654",
-    "express": "registry:dt/express#4.0.0+20160317120654",
-    "express-serve-static-core": "registry:dt/express-serve-static-core#0.0.0+20160322035842",
     "glob": "registry:dt/glob#5.0.10+20160317120654",
```
rictic: I know a lot of these exist within the npm `@types` ecosystem, mind removing as many as you can?

rictic: Ah, this is handled in a TODO, ok

custom_typings/chalk.d.ts
FredKSchott: It does but unfortunately it doesn't support the `dim` feature (older version) and seems to be written in a way that can't be easily extended. I'll add a comment and maybe after your PR is merged we can update the DT version.

tsconfig.json
FredKSchott: I'm good with this if we can agree to standardize across all repos? I've been using `es2016` up to this point.

typings.json
FredKSchott: coming up in 3... 2... 1... https://github.com/Polymer/polymer-cli/pull/411

tsconfig.json
rictic: The difference is very minor, I say update as needed.

FredKSchott: 👍

gulpfile.js
```diff
@@ -28,12 +28,13 @@ gulp.task('init', () => gulp.src("./typings.json").pipe(typings()));
...
 gulp.task('lint', ['tslint', 'eslint', 'depcheck']);
 
-gulp.task('build', () =>
-  mergeStream(
-    gulp.src('src/**/*.ts').pipe(typescript(tsProject)),
-    gulp.src(['src/**/*', '!src/**/*.ts'])
-  ).pipe(gulp.dest('lib'))
-);
+gulp.task('build', () => {
+  let tsReporter = typescript.reporter.defaultReporter();
```
justinfagnani: how far are we deviating from tools-common now? I suspect we need to do a re-sync...

src/commands/build.ts
```diff
@@ -12,6 +12,11 @@ import {ArgDescriptor} from 'command-line-args';
 import {Command} from './command';
 import * as logging from 'plylog';
 
+// Only import type definitions here, otherwise this line will be included in
+// the JS output, triggering  the entire build library & its dependencies to
+// be loaded and parsed.
+import {BuildOptions} from '../build/build';
```
justinfagnani: could you use a /// reference for this?

src/init/init.ts
```diff
@@ -124,11 +124,7 @@ function createYeomanEnvironment(): Promise<any> {
     env.registerStub(ApplicationGenerator, 'polymer-init-application:app');
     env.registerStub(shopGenerator, 'polymer-init-shop:app');
     env.registerStub(pskGenerator, 'polymer-init-starter-kit:app');
-    env.lookup((error) => {
-      if (error) {
-        reject(error);
-        return;
-      }
+    env.lookup(() => {
```
justinfagnani: can `createYeomanEnvironment` return `Promise<YeomanEnvironment>`?

src/init/init.ts
```diff
@@ -174,7 +170,7 @@ export function runGenerator(generatorName, options): Promise<any> {
 
   return new Promise((resolve, reject) => {
```
justinfagnani: this could just be: `const env = await options.env || createYeomanEnvironment();`

src/init/init.ts
justinfagnani: Is it really ok to remove the error argument? An error will cause the Promise to never settle. The Yeoman docs say the callback takes an error.

gulpfile.js
FredKSchott: definitely, this is the new gulp-typescript@v3 interface

src/commands/build.ts
FredKSchott: in addition to the import statement? We still want to use the BuildOptions object in our definitions in this file

src/init/init.ts
FredKSchott: So I made this change because upgrading to typescript 2.0 exposed that this callback doesn't actually return an error in its type definition.

BUT looking again just now, and I realized that this was coming from our own custom type declaration and not an official one. The official documentation actually does mention that an error can be returned (although if you drill into the code even further, it's always null).

I'll revert this change and update the type declaration, thanks for catching this!

src/commands/build.ts
rictic: I don't think a /// reference is enough here.

PR #411, opened 2016-09-23
move from typings over to npm types

FredKSchott:
🎉


gulpfile.js
```diff
@@ -24,8 +24,6 @@ const typings = require('gulp-typings');
```
rictic: no need to require gulp-typings any more

src/build/build.ts
```diff
@@ -26,6 +24,10 @@ import {PrefetchTransform} from './prefetch';
 import {waitFor} from './streams';
 import {parsePreCacheConfig} from './sw-precache';
 
+// TODO(fks) 09-22-2016: Latest npm type declaration resolves to a non-module
+// entity. Upgrade to import once compatible .d.ts file is released.
+import mergeStream = require('merge-stream');
```
rictic: Is that a problem? I'm pretty sure `import * as mergeStream from 'merge-stream'` works with non-module entities (even if it shouldn't)

FredKSchott: I get this error when using `@types/merge-stream`:
```
src/build/build.ts(29,30): 
  error TS2497: 
  Module '"/Users/fschott/Code/polytool/node_modules/@types/merge-stream/index"' 
  resolves to a non-module entity and cannot be imported using this construct.
```

I don't think it's worth writing a custom declaration now, but I'll re-write the TODO to mention that one could be added if someone does refactor some of this code.

rictic: Interesting. I wonder if I'm gonna start seeing errors in other repos as I update to the real typescript v2.0.x

rictic:
Looks like there's still an issue with `gulp-typings` being imported somewhere? https://travis-ci.org/Polymer/polymer-cli/jobs/162092389

rictic:
Need to get rid of npm `init` task, and the `npm run init` line in our travis.yaml

FredKSchott:
@rictic haha you're finding these faster than I can push them :)

re init & travis: I want to clean up our gulpfile a bit in general after this, now that typings is gone but we still have these `build` & `build-all` tests.

PR #414, opened 2016-09-23
add back env.lookup() error handling

FredKSchott:
I looked into adding a test but because we call lookup() in the same function scope where we create the Env object it's not easy. Since this error [never actually happens](http://yeoman.io/environment/resolver.js.html#line49) I'm okay with leaving this untested for now, unless anyone feels strongly against.

/cc @justinfagnani @rictic

### DIFF
--- a/custom_typings/yeoman-environment.d.ts
+++ b/custom_typings/yeoman-environment.d.ts
@@ -4,7 +4,7 @@ declare module 'yeoman-environment' {
     registerStub(generator: any, namespace: string);
     run(target: string, options: any, done: Function);
     getGeneratorsMeta(): YeomanEnvironment.GeneratorMeta[];
-    lookup(callback: () => void);
+    lookup(callback: (error?: Error) => void);
   }
 
   namespace YeomanEnvironment {
--- a/src/init/init.ts
+++ b/src/init/init.ts
@@ -124,7 +124,11 @@ function createYeomanEnvironment(): Promise<any> {
     env.registerStub(ApplicationGenerator, 'polymer-init-application:app');
     env.registerStub(shopGenerator, 'polymer-init-shop:app');
     env.registerStub(pskGenerator, 'polymer-init-starter-kit:app');
-    env.lookup(() => {
+    env.lookup((error?: Error) => {
+      if (error) {
+        reject(error);
+        return;
+      }
       resolve(env);
     });
   });


custom_typings/yeoman-environment.d.ts
```diff
@@ -4,7 +4,7 @@ declare module 'yeoman-environment' {
     registerStub(generator: any, namespace: string);
     run(target: string, options: any, done: Function);
     getGeneratorsMeta(): YeomanEnvironment.GeneratorMeta[];
-    lookup(callback: () => void);
+    lookup(callback: (error?: Error) => void);
```
rictic: Be careful about aspirationally strict types in typings for external code. Unless we're confident that this is part of the contract for yeoman-environment, this should probably be typed as `any`

src/init/init.ts
```diff
@@ -124,7 +124,11 @@ function createYeomanEnvironment(): Promise<any> {
     env.registerStub(ApplicationGenerator, 'polymer-init-application:app');
     env.registerStub(shopGenerator, 'polymer-init-shop:app');
     env.registerStub(pskGenerator, 'polymer-init-starter-kit:app');
-    env.lookup(() => {
+    env.lookup((error?: Error) => {
```
rictic: ditto

custom_typings/yeoman-environment.d.ts
FredKSchott: Agreed, I'm usually too trusting when it comes to this.

[Their official documentation](http://yeoman.io/environment/Environment.html#lookup) reads "Take err as first parameter." Given the `err` name and the fact that "error-as-a-first-argument" is the default node callback style, I'm confident. Do you still have concerns?

justinfagnani: It's just that when JS code says "error", they often don't really mean `Error`. It's case-by-case. In this case they mean `null` :/

rictic: Yeah, node callback `err` params are sometimes strings, sometimes Errors, sometimes numbers...

rictic: Oh wow, it's always `null` in this case. lulz

justinfagnani:
LGTM though my existential leanings are triggered by the fact that since `lookup()` only ever passes `null` to the callback, this change will likely never matter.

custom_typings/yeoman-environment.d.ts
rictic: In a point release yeoman-environment would be in their rights to pass anything in as an err value though. I still recommend `any`

FredKSchott: That's crazy, I don't think I've ever seen a reasonably-supported module return strings, numbers, etc. in a `error/err` callback argument.

PR #417, opened 2016-09-23
add "no-unused-variable" rule to tslint

FredKSchott:
Now linting unused variables on lint. 

Since we have tslint we don't need to add this to tsconfig to warn on build as well, and I'd rather keep all of our linting rules in one place instead of two.

/cc @rictic @justinfagnani


tslint.json
```diff
@@ -10,6 +10,7 @@
         "no-internal-module": true,
         "no-trailing-whitespace": true,
         "no-var-keyword": true,
+        "no-unused-variable": true,
```
rictic: IMO we should use tsconfig for this, not tslint

```
        "noUnusedLocals": true,
        "noUnusedParameters": true,
```

FredKSchott: Yea I actually wanted to talk about this a bit. Because these are linting errors, I wouldn't want our `gulp lint` task to pass if these problems existed in our code. I also would want my text editor to show these errors if it had `tslint`/linting integrated. This was why I ultimately added them to tslint instead of tsconfig.

rictic: Those things should be even more true with tsconfig errors, shouldn't they?

rictic: Hm, gulp-typescript doesn't fail on type errors. I think it should, let's see if that's an option...

FredKSchott: I don't see it as a big problem if our project builds with unused variables, if we can trust that they are already being caught during linting & testing.

While it would be nice to check in both steps, practically we don't want to be maintaining two different linting rulesets.

justinfagnani: To be clear, `tsc` errors don't stop the build, they're more like warnings.

FredKSchott: As discussed in person, we'll put in both and I'll expand my miiiiiiind
miiiiind

src/init/init.ts
```diff
@@ -172,9 +172,10 @@ export function runGenerator(generatorName, options): Promise<any> {
   options = options || {};
   let templateName = options.templateName || generatorName;
 
-  return new Promise((resolve, reject) => {
-    resolve(options.env || createYeomanEnvironment());
-  }).then(function(env: YeomanEnvironment) {
+  return ((options.env)
```
rictic: extra parens here made this a little harder for me to read, I initially wanted to parse it as an arrow function

src/environments/reyserve.ts
```diff
@@ -55,7 +55,10 @@ export class ReyServe implements Environment {
           cwd: options.root
         });
       devServer.on('close', (code) => {
-        resolve();
+        resolve(code);
+      });
+      devServer.on('error', (error) => {
+        reject(error);
```
rictic: actual bug caught! nice

FredKSchott:
@rictic @justinfagnani PTAL

src/init/init.ts
```diff
@@ -205,9 +206,10 @@ export function promptGeneratorSelection(options): Promise<any> {
   options = options || {};
   let env: YeomanEnvironment;
 
-  return new Promise((resolve, reject) => {
-    resolve(options.env || createYeomanEnvironment());
-  }).then(function(_env: YeomanEnvironment) {
+  return ((options.env)
```
rictic: here as well

src/init/init.ts
FredKSchott: agreed, removing...

PR #419, opened 2016-09-24
Align html templates closer to what html5-boilerplate recommends

yandavid:


yandavid:
For context, here's html5-boilerplate's rationale for putting the utf-8 charset declaration before title: https://github.com/h5bp/html5-boilerplate/blob/5.3.0/dist/doc/html.md#the-order-of-the-title-and-meta-tags

As for the appveyor build fix, the gist is that npm 3 solves all the things. See here for more info: https://github.com/npm/npm/issues/9696#issuecomment-168622331.

justinfagnani:
Thanks for the context and npm2 fix!

PR #426, opened 2016-10-01
Add analyze command

justinfagnani:


src/commands/analyze.ts
```diff
@@ -0,0 +1,55 @@
...
+      urlLoader: new FSUrlLoader(config.root),
+      urlResolver: new PackageUrlResolver(),
+    });
+
+    const elements = new Set<Element>();
+
+    for (const input of options.input) {
+      const document = await analyzer.analyze(input);
+      const docElements = Array.from(document.getByKind('element'))
+          .filter((e) => !e.sourceRange.file.startsWith('bower_components'));
```
rictic: or node_modules?

justinfagnani: done

src/commands/analyze.ts
```diff
@@ -0,0 +1,56 @@
...
+import * as logging from 'plylog';
+import {Command} from './command';
+import {generateElementMetadata} from 'polymer-analyzer/lib/generate-elements';
+
+let logger = logging.getLogger('cli.command.analyze');
+
+export class AnalyzeCommand implements Command {
+  name = 'analyze';
+
+  description = 'Analyzes the input file with the Polymer Analyzer';
```
FredKSchott: why would I run this? What is the output? How is it output?

src/commands/analyze.ts
```diff
@@ -0,0 +1,56 @@
...
+    const elements = new Set<Element>();
+
+    for (const input of options.input) {
+      const document = await analyzer.analyze(input);
+      const docElements = Array.from(document.getByKind('element'))
+          .filter((e) => !e.sourceRange.file.startsWith('bower_components') &&
+            !e.sourceRange.file.startsWith('node_modules'));
+      docElements.forEach((e) => elements.add(e));
+    }
+    const metadata = generateElementMetadata(Array.from(elements), '');
```
FredKSchott: aside: I'm getting a little code smell from this interface. Can we make this second argument optional in analyzer?

src/commands/analyze.ts
```diff
@@ -0,0 +1,56 @@
...
+    console.log(JSON.stringify(metadata, null, 2));
```
FredKSchott: nit: If we want to output raw data, I'd rather see a direct call to `process.stdout.write` to skip any weirdness inside console.log (like calling `util.format()`)

src/commands/analyze.ts
```diff
@@ -0,0 +1,56 @@
...
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+import {Analyzer} from 'polymer-analyzer';
```
FredKSchott: can you time the impact that adding this has on the `help` or `--version` commands? If it's significant, move this into `require()` calls inside of run.

rictic:
Gentle ping on this PR, I've been seeing a steady and consistent demand for it, as well as it being very helpful in debugging and getting better error reports for all of our tools downstream of the analyzer.

justinfagnani:
k, I'll get this in today

src/commands/analyze.ts
justinfagnani: Split the command into a lazy part. It's actually a little tricky to get TS to load just the type declarations from a module.

src/commands/analyze.ts
justinfagnani: Yes, this function needs a bit of tweaking to be more usable.

src/commands/analyze.ts
justinfagnani: I changed this to be slightly better, but since we only have a short description, it's limited right now. I think we should offer long descriptions and per-command help to fix this.

justinfagnani:
PTAL

Sorry for the push -f, but I rebased to make sure it still worked.

package.json
```diff
@@ -56,7 +56,7 @@
     "parse5": "^2.2.3",
     "plylog": "^0.4.0",
     "polylint": "^2.10.0",
-    "polymer-analyzer": "^2.0.0-alpha.10",
+    "polymer-analyzer": "^2.0.0-alpha.17",
```
rictic: Recommend `=2.0.0-alpha.17` as we do breaking changes in alpha releases.

src/analyze/analyze.ts
```diff
@@ -0,0 +1,34 @@
...
+    urlLoader: new FSUrlLoader(root),
+    urlResolver: new PackageUrlResolver(),
+  });
+
+  const elements = new Set<Element>();
+
+  for (const input of inputs) {
+    const document = await analyzer.analyze(input);
+    const docElements = Array.from(document.getByKind('element'))
+        .filter((e: Element) => !e.sourceRange.file.startsWith('bower_components') &&
```
rictic: I'd say that we should exclude even nested `bower_components` and `node_modules` directories. So something like:

```typescript
!/(\b|\/|\\)(bower_components|node_modules)(\/|\\/.test(e.sourceRange.file)
```

RegExp tested here: https://regex101.com/r/anjD41/1/tests

src/commands/analyze.ts
```diff
@@ -8,49 +8,33 @@
...
 
-  async run(options, config): Promise<any> {
-    if (!options || !options.input) {
+  async run(options: CommandOptions, config: ProjectConfig): Promise<any> {
+    const analyze = require('../analyze/analyze').analyze;
+    const root = config.root;
+    const inputs = options['input'];
+
+    if (!options || !options['input']) {
       logger.debug('no inputs given');
```
rictic: In the future I think running with no arguments should run over the entire project (defined by some criteria, potentially in the Analyzer)

src/analyze/analyze.ts
justinfagnani: By nested do you mean within bower_components, or say, something in `test/`?

rictic: The latter

src/commands/analyze.ts
justinfagnani: totally agreed, but right now this would pull in demos, tests, etc.

src/analyze/analyze.ts
rictic: people with unexpected project layouts, like a `server/` and `client/` toplevel directories or similar

src/commands/analyze.ts
rictic: good point. we'd want to lint in that case, but not do this kind of analysis

src/analyze/analyze.ts
justinfagnani: ok, done

marcus7777:
so now we can analyze?

PR #428, opened 2016-10-03
convert build command to use async-await

FredKSchott:
no behavior change, just a big indentation change and cleaner code

/cc @justinfagnani

### DIFF
--- a/src/build/build.ts
+++ b/src/build/build.ts
@@ -40,105 +40,100 @@ export interface BuildOptions {
   };
 }
 
-export function build(options: BuildOptions, config: ProjectConfig): Promise<any> {
+export async function build(options: BuildOptions, config: ProjectConfig): Promise<void> {
 
-  return new Promise<any>((buildResolve, _) => {
-    let polymerProject = new PolymerProject({
-      root: config.root,
-      shell: config.shell,
-      entrypoint: config.entrypoint,
-      fragments: config.fragments,
-      sourceGlobs: config.sourceGlobs,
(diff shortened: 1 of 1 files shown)


src/build/build.ts
```diff
@@ -40,105 +40,100 @@ export interface BuildOptions {
...
-    let bundledPostProcessing = Promise.all([loadSWConfig, waitFor(bundledPhase)]).then((results) => {
-      let swConfig: SWConfig = results[0];
-      return addServiceWorker({
-        buildRoot: 'build/bundled',
-        project: polymerProject,
-        swConfig: swConfig,
-        bundled: true,
-      });
+  // Once the unbundled build stream is complete, create a service worker for the build
+  let unbundledPostProcessing = Promise.all([loadSWConfig, waitFor(unbundledPhase)]).then((results) => {
```
justinfagnani: I don't suppose it would be any nicer to use an IIAFE?

```typescript
(async () => {
  const result = await Promise.all([loadSWConfig, waitFor(unbundledPhase)]);
  const swConfig = result[0];
  await addServiceWorker(...);
}());
```

Just brainstorming to see if there's any nice pattern for doing all this parallel work...

FredKSchott: Normally I would agree, but since the result is being saved to `unbundledPostProcessing` as a promise and then used later inside of a `Promise.all()` call I kind of like that we're explicitly using a Promise here.

tl;dr: ¯\_(ツ)_/¯

PR #429, opened 2016-10-04
Add --.js.compile flag to build

justinfagnani:


FredKSchott:
@justinfagnani can you rebase to master?

src/build/babel-transform.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  constructor() {
+    super({objectMode: true});
+    try {
+      this._babel = requireg('babel-core');
+    } catch (e) {
+      logger.error(`${e.message}\nDid you install babel-core? Run \`npm i -g babel-core\``);
+    }
+    try {
+      this._preset = requireg('babel-preset-es2015');
```
web-padawan: FYI: there is a [bug](https://github.com/web-animations/web-animations-next/issues/402) in web-animations-next-lite.min.js used by `neon-animated-page`. It becomes broken in strict mode   introduced by `babel-preset-es2015`. I ended up with using [babel-preset-es2015-without-strict](https://github.com/fancyboynet/babel-preset-es2015-without-strict) instead.

FredKSchott: @justinfagnani we need to support custom .babelrc files. Its actually easier to support this than it is to work around it, since babel-core will automatically look for presets/plugins in the cwd. Then we can remove the `babel-preset-es2015` dependency and let users bring their own config, plugins, presets, etc. for each project.

Check out my WIP branch for example of how to do this: https://github.com/Polymer/polymer-cli/compare/wip-babel?expand=1

FredKSchott:
Superseded by https://github.com/Polymer/polymer-cli/pull/525

PR #430, opened 2016-10-05
added configFile argv for WCT.

hvdb:
see: https://github.com/Polymer/web-component-tester/pull/388

### DIFF
--- a/src/commands/test.ts
+++ b/src/commands/test.ts
@@ -106,7 +106,7 @@ export class TestCommand implements Command {
     },
     {
       name: 'configFile',
-      description: "Config file that needs to be used by wct. ie: wct.config-sauce.js",
+      description: 'Config file that needs to be used by wct. ie: wct.config-sauce.js',
       type: String,
     },
   ];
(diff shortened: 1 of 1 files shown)


FredKSchott:
Thanks @hvdb! I'll try to take a look at this and Polymer/web-component-tester#388 as soon as possible (The team is busy getting ready for Polymer Summit next week so it might be a little while before I can)

src/commands/test.ts
```diff
@@ -104,6 +104,11 @@ export class TestCommand implements Command {
       description: "Set job visibility to 'public', 'public restricted', 'share', 'team' or 'private'",
       type: String,
     },
+    {
+      name: 'configFile',
```
FredKSchott: all cli flags have followed the `dash-case` format, please change to `config-file` and handle the mapping into WCT below.

hvdb: will do!

hvdb: The fix would be in WCT to specify the correct full name:
https://github.com/Polymer/web-component-tester/blob/master/runner/config.ts#L219

in addition i think the simpleOutput should also be fixed, here and in WCT

PR #439, opened 2016-10-13
Added --bundled and --unbundled flags to build

### LINKED ISSUE #303: build optimisation by selecting bundled or unbundled at build time
### Description

It could be interesting to set up a specific build type:
polymer build --unbundled
polymer build --bundled

 (many people doesn't use both at the same time:)

### Versions & Environment

- Polymer CLI:
- node:
- Operating System:

whenlit:
Fixes #303 by adding --bundled and --unbundled flags. Builds both bundled and unbundled when both flags or no flags are specified.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ Use `index.html` as the entrypoint, or read from `polymer.json`:
 
     $ polymer build
 
-`build` is opinionated and defaults to a good build for app-shell apps. It writes the built output to `build/bundled` and `build/unbundled` folders. Both outputs have been run though HTML, JS and CSS optimizers, and have a Service Worker generated for them. The bundled folder contains the application files process by Vulcanize, Polymer's HTML bundler, for optimal loading via HTTP/1. The unbundled folder is optimized for HTTP/2 + Push.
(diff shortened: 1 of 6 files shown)


FredKSchott:
@whenlit thanks for the PR! Instead of doing two builds by default, we'd like to default to bundled and provide an `unbundled` flag for users who can take advantage of http2/push. It would be great if we could then place the build directly in the `build` directory, instead of nested under `bundled` or `unbundled`.

whenlit:
@FredKSchott Ok, I have a version now that does exactly that, and generates exactly the sets of files that the existing build did, one set or the other. 

The only thing is: I now run the analyzer before or after the splitHtml / rejoinHtml depending on whether the bundled flag is set, and I can't figure out why it won't just work when I put it in one place :)

When I run the analyzer before the splitHtml() call, the bundled build generates a much larger index.html file (unsurprisingly).
When I run the analyzer after the splitHtml() call, the unbundled build does not include the dependencies (whereas the bundled build does).

Long story short, it works but it doesn't seem to be optimal. Any ideas?

FredKSchott:
Ah yea, we've seen this problem before where analysis has trouble analyzing the minified code after optimziation. We have a PR out to fix this in polymer-build (https://github.com/Polymer/polymer-build/pull/77) so that analysis is run automatically for you. Lets wait for that to land and then it will be much easier to get this working.

FredKSchott:
The team is planning some larger improvements to the build interface/workflow, and we're going to tackle this as a part of that larger project. See #507 for details.

thanks again for taking a crack at this @whenlit.

Issue #453, opened 2016-11-02
Build fails when run out of the box


skamansam:
### Description
Build fails when run out of the box. The path to the polymer.html file is incorrect.

### Versions & Environment
- Polymer CLI: 0.17.0
- node: v4.2.4
- Operating System: Fedora 24

#### Steps to Reproduce
1. Create an application project: `polymer init element`
2. Build: `polymer build`

#### Expected Results
No error is thrown

#### Actual Results
```
info:    Preparing build...
info:    Building application...
info:    Generating build/unbundled...
info:    Generating build/bundled...
error:   Uncaught exception: Error: file path is not in root: /path/to/element/../polymer/polymer.html (/path/to/element)
error:   Error: file path is not in root: /path/to/element/../polymer/polymer.html (/path/to/element/../)
...
```

FredKSchott:
`polymer build` is for building applications for deployment. There is currently no way to build an element since an element is meant to live in `bower_components`. See https://github.com/Polymer/polymer-cli/issues/317 & https://github.com/Polymer/polymer-cli/issues/338 for more discussion

nicholaswmin:
@FredKSchott Then how do we transpile an element to ES5? Are there any guides/tools you can point me to?

justinfagnani:
@nicholaswmin you don't transpile individual elements to ES5. Distribute them as ES6 (since custom elements _must_ be classes, compiling them would break them), then let applications compile to ES5 if necessary, and if they set up the ES5 custom elements adapter so that they will work.

nicholaswmin:
@justinfagnani I didn't mention I was talking about Polymer 1.x which uses the factory method instead of class. Nevertheless I get the point. Thanks.

PR #454, opened 2016-11-02
make sure the template builds and works with out of the box [#453]

skamansam:

Just changed some of the paths in the templates so everything works cleanly out of the box.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## Unreleased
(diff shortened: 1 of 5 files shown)


FredKSchott:
This is incorrect, an element references bower components as siblings, since it is hosted alongside them in the `bower_components/` directory.

justinfagnani:
I do think this reinforces the need for `build` and `serve` to detect app vs element projects

PR #457, opened 2016-11-04
noImplicitAny: true

justinfagnani:


src/commands/lint.ts
```diff
@@ -73,11 +72,11 @@ export class LintCommand implements Command {
     }
   ];
 
-  run(options, config): Promise<any> {
+  run(options: CommandOptions, config: ProjectConfig): Promise<any> {
     // Defer dependency loading until this specific command is run
     const polylint = require('polylint/lib/cli');
 
-    let lintFiles: string[] = options.input;
+    let lintFiles: string[] = options['input'];
```
FredKSchott: was this clang or TS, or is this just personal preference? If we're talking personal preference I prefer indexing properties only when necessary.

justinfagnani: The object is used as a dictionary, so this syntax represents its use the best. We could create a type per command, but I wanted to do a small change to enable `noImplicitAny`.

justinfagnani:
Add issue #461

PR #459, opened 2016-11-04
add common clang-format rules

FredKSchott:

***Only merge when there are no outstanding PRs, and remember to run `npm run format` after merging!***

/cc @rictic @justinfagnani

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,8 @@
     "lint": "gulp lint",
     "build": "gulp build",
     "test": "gulp lint test",
-    "prepublish": "npm run build && npm test"
+    "prepublish": "npm run build && npm test",
+    "format": "find src test | grep '\\.js$\\|\\.ts$' | xargs clang-format --style=file -i"
   },
   "repository": {
     "type": "git",
@@ -78,6 +79,7 @@
   "devDependencies": {
     "bower": "^1.7.9",
     "chai": "^3.5.0",
+    "clang-format": "^1.0.45",
     "depcheck": "^0.6.3",
     "gulp": "^3.9.1",
     "gulp-eslint": "^2.0.0",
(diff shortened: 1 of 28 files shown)


package.json
```diff
@@ -74,6 +75,7 @@
   "devDependencies": {
     "bower": "^1.7.9",
     "chai": "^3.5.0",
+    "clang-format": "^1.0.45",
```
rictic: pin to a specific version please

FredKSchott:
I will rebase this off of master once the current batch of PRs are merged, and then commit the reformatting for a final look before merging.

rictic:
Will there ever be a time when there's not a batch of open PRs?

FredKSchott:
If we can get https://github.com/Polymer/polymer-cli/pull/490 in we should be good. Trying to avoid unnecessary work for @justinfagnani if we don't need to.

No harm in seeing it here though, and then rebasing later. Pushing formatted code up to this branch...

src/commands/help.ts
```diff
@@ -21,6 +25,8 @@ const m = chalk.magenta;
 const CLI_TITLE = chalk.bold.underline('Polymer-CLI');
 const CLI_DESCRIPTION = 'The multi-tool for Polymer projects';
 const CLI_USAGE = 'Usage: \`polymer <command> [options ...]\`';
+
+// clang-format off - keep the custom white space here to help readibility
```
FredKSchott: looks like this is working now! Woo!

FredKSchott:
@rictic PTAL

src/build/optimize-streams.ts
```diff
@@ -1,8 +1,9 @@
-import {minify as uglify, MinifyOptions as UglifyOptions} from 'uglify-js';
 import {css as cssSlam} from 'css-slam';
```
rictic: Need to add the license header

src/commands/help.ts
rictic: nice!

PR #462, opened 2016-11-08
Enable strictNullChecks

justinfagnani:


src/build/prefetch.ts
```diff
@@ -42,8 +42,9 @@ export class PrefetchTransform extends Transform {
     type: 'import' | 'prefetch'
   ) {
     let contents = file.contents.toString();
-    let ast = parse5.parse(contents);
-    let head = dom5.query(ast, dom5.predicates.hasTagName('head'));
+    const ast = parse5.parse(contents);
+    // parse5 always produces a <head> element
+    const head = dom5.query(ast, dom5.predicates.hasTagName('head'))!;
```
rictic: +1 to `const`ifying. Looks like typescript finds more and more places where they're able to do better type inference with `const` over `let`. (yes, `const`-like behavior can ~always be inferred but the convenience of not having to do that does seem to be paying out for the tsc team) https://github.com/Microsoft/TypeScript/pull/10676

src/build/prefetch.ts
```diff
@@ -81,27 +82,29 @@ export class PrefetchTransform extends Transform {
       return done();
     }
     this.analyzer.analyzeDependencies.then((depsIndex: DepsIndex) => {
-      let fragmentToDeps = new Map(depsIndex.fragmentToDeps);
+      const fragmentToDeps = new Map(depsIndex.fragmentToDeps);
 
       if (this.config.entrypoint && this.config.shell) {
-        let file = this.fileMap.get(this.config.entrypoint);
+        const file = this.fileMap.get(this.config.entrypoint)!;
+        console.assert(file != null);
```
rictic: I think this only logs, when I think you want to throw.

https://developer.mozilla.org/en-US/docs/Web/API/console/assert

src/build/prefetch.ts
```diff
@@ -81,27 +82,29 @@ export class PrefetchTransform extends Transform {
...
         this.push(file);
         this.fileMap.delete(this.config.entrypoint);
       }
 
-      for (let im of this.config.allFragments) {
-        let file = this.fileMap.get(im);
-        let deps = fragmentToDeps.get(im);
+      for (const importUrl of this.config.allFragments) {
+        const file = this.fileMap.get(importUrl)!;
+        console.assert(file != null);
```
rictic: Ditto

src/polymer-cli.ts
```diff
@@ -184,7 +184,8 @@ export class PolymerCli {
 
     let commandName = parsedArgs.command;
     let commandArgs = parsedArgs.argv;
-    let command = this.commands.get(commandName);
+    let command = this.commands.get(commandName)!;
+    console.assert(command != null);
```
rictic: ditto

src/build/prefetch.ts
justinfagnani: done

src/polymer-cli.ts
justinfagnani: done

src/build/prefetch.ts
justinfagnani: Oh nice, if I check null and throw, I don't need the `!` type operator.

justinfagnani:
PTAL

src/build/prefetch.ts
rictic: +1, agree and affirm.

PR #471, opened 2016-11-20
Fix runtime error on `polymer init`

tony19:

`createYeomanEnvironment()` returns a promise, but a couple callers weren't updated to await it, resulting in a runtime error when they tried to access methods not available on a promise.

### DIFF
--- a/src/init/init.ts
+++ b/src/init/init.ts
@@ -177,7 +177,7 @@ export async function runGenerator(generatorName: string, options: {[name: strin
   options = options || {};
   const templateName = options['templateName'] || generatorName;
 
-  const env: YeomanEnvironment = await options['env'] || createYeomanEnvironment();
+  const env: YeomanEnvironment = await options['env'] || await createYeomanEnvironment();
 
   logger.info(`Running template ${templateName}...`);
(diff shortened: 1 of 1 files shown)


src/init/init.ts
```diff
@@ -177,7 +177,7 @@ export async function runGenerator(generatorName: string, options: {[name: strin
   options = options || {};
   const templateName = options['templateName'] || generatorName;
 
-  const env: YeomanEnvironment = await options['env'] || createYeomanEnvironment();
+  const env: YeomanEnvironment = await options['env'] || await createYeomanEnvironment();
```
justinfagnani: Options['env'] is always a non-Promise value. The `await` was intended to apply to the whole or'ed expression. I'd suggest keeping just one `await` by doing: 

```
const env: YeomanEnvironment = await (options['env'] || createYeomanEnvironment());
```

justinfagnani:
Thanks @tony19! One small comment on the need for two `await`s on that line.

justinfagnani:
Thanks @tony19 

FYI, I wrote a test for this which I'll merge afterward.

PR #472, opened 2016-11-20
Fix generator names

tony19:

The generator name was parsed from the directory name (`./generator-polymer-init-foo/` --> `"foo"`), and this name was given to all sub-generators in the directory, regardless of their individual `package.json` config for `name`, which made them indistinguishable without descriptions. Technically, Yeoman generators are supposed to be named based on their `package.json` config [[1]](http://yeoman.io/authoring/).

### Example

```
generator-polymer-init-my-app
├── generators
│   ├── foo
│   │   ├── index.js
│   │   ├── package.json   # name: 'polymer-init-foo-el', description: 'My foo element'
│   │   └── templates
│   │      └── _element_test.html
│   └── bar
│       ├── index.js
│       ├── package.json   # name: 'polymer-init-bar-el', description: 'My bar element'
│       └── templates
│          └── _element_test.html
└── package.json # description: 'My test app'
```

*Before*
```
$ polymer init
? Which starter template would you like to use? 
  element - A blank element template 
  application - A blank application template 
  shop - The "Shop" Progressive Web App demo 
  starter-kit - A starter application template, with navigation and "PRPL pattern" loading 
❯ my-app - My foo element
  my-app - My bar element
```

*After*
```
$ polymer init
? Which starter template would you like to use? 
  element - A blank element template 
  application - A blank application template 
  shop - The "Shop" Progressive Web App demo 
  starter-kit - A starter application template, with navigation and "PRPL pattern" loading 
❯ foo-el - My foo element
  bar-el - My bar element
```

### DIFF
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -37,7 +37,8 @@ export class InitCommand implements Command {
 
     const templateName = options['name'];
     if (templateName) {
-      const generatorName = `polymer-init-${templateName}:app`;
+      const subgen = (templateName.indexOf(':') !== -1) ? '' : ':app';
+      const generatorName = `polymer-init-${templateName}${subgen}`;
       logger.debug('template name provided', {
         generator: generatorName,
         template: templateName,
--- a/src/init/init.ts
+++ b/src/init/init.ts
@@ -136,9 +136,9 @@ function getGeneratorMeta(
 
 /**
  * Extract the meaningful name from the full Yeoman generator name.
- * Strip the standard generator prefixes ("generator-" and "polymer-init-"),
- * and extract the remainder of the name (the first part of the string
- * before any colons).
+ * Strip the standard generator prefixes ("generator-", "polymer-init-",
+ * and "polymer-starter-kit-"), and extract the remainder of the name
+ * (the first part of the string before any colons).
  *
  * Examples:
  *
@@ -150,16 +150,18 @@ function getGeneratorMeta(
  *   'foo-bar:ccc'                        === 'foo-bar'
  */
 function getDisplayName(generatorName: string) {
-  // Breakdown of regular expression to extract name (group 3 in pattern):
+  // Breakdown of regular expression to extract name (group 4 in pattern):
   //
-  //  Pattern           | Meaning
-  //  -------------------------------------------------------------------
-  //  (generator-)?     | Group 1; Match "generator-"; Optional
-  //  (polymer-init-)?  | Group 2; Match "polymer-init-"; Optional
-  //  ([^:]+)           | Group 3; Match one or more characters != ':'
-  //  (:.*)?            | Group 4; Match ':' followed by anything; Optional
-  return generatorName.replace(/(generator-)?(polymer-init-)?([^:]+)(:.*)?/g,
-                               '$3');
+  // Pattern                 | Meaning
+  // -------------------------------------------------------------------
+  // (generator-)?           | Grp 1; Match "generator-"; Optional
+  // (polymer-init)?         | Grp 2; Match "polymer-init-"; Optional
+  // (polymer-starter-kit-)? | Grp 3; Match "polymer-starter-kit-"; Optional
+  // ([^:]+)                 | Grp 4; Match one or more characters != ":"
+  // (:.*)?                  | Grp 5; Match ":" followed by anything; Optional
+  return generatorName.replace(
+      /(generator-)?(polymer-init-)?(polymer-starter-kit-)?([^:]+)(:.*)?/g,
+      '$4');
 }
 
 /**
--- a/src/init/init.ts
+++ b/src/init/init.ts
@@ -82,16 +82,15 @@ function checkIsMinGW(): boolean {
 function getGeneratorDescription(
     generator: YeomanEnvironment.GeneratorMeta,
     generatorName: string): GeneratorDescription {
-  const name = getDisplayName(generatorName);
   let description: string = '';
+  let genName = generatorName;
 
-  if (templateDescriptions.hasOwnProperty(name)) {
-    description = templateDescriptions[name];
-  } else if (generator.resolved && generator.resolved !== 'unknown') {
+  if (generator.resolved && generator.resolved !== 'unknown') {
     try {
       const metapath = findup('package.json', {cwd: generator.resolved});
       const meta = JSON.parse(fs.readFileSync(metapath, 'utf8'));
-      description = meta.description;
+      description = meta.description || '';
+      genName = meta.name || generatorName;
     } catch (error) {
       if (error.message === 'not found') {
         logger.debug('no package.json found for generator');
@@ -103,6 +102,13 @@ function getGeneratorDescription(
       }
     }
   }
+
+  const name = getDisplayName(genName);
+
+  if (templateDescriptions.hasOwnProperty(name)) {
+    description = templateDescriptions[name];
+  }
+
   // If a description exists, format it properly for the command-line
   if (description.length > 0) {
     description = chalk.dim(` - ${description}`);
@@ -118,14 +124,31 @@ function getGeneratorDescription(
 }
 
 /**
- * Extract the meaningful name from the full yeoman generator name
+ * Extract the meaningful name from the full Yeoman generator name.
+ * Strip the standard generator prefixes ("generator-" and "polymer-init-"),
+ * and extract the remainder of the name (the first part of the string
+ * before any colons).
+ *
+ * Examples:
+ *
+ *   'generator-polymer-init-foo'         === 'foo'
+ *   'polymer-init-foo'                   === 'foo'
+ *   'foo-bar'                            === 'foo-bar'
+ *   'generator-polymer-init-foo:aaa'     === 'foo'
+ *   'polymer-init-foo:bbb'               === 'foo'
+ *   'foo-bar:ccc'                        === 'foo-bar'
  */
 function getDisplayName(generatorName: string) {
-  let nameEnd = generatorName.indexOf(':');
-  if (nameEnd === -1) {
-    nameEnd = generatorName.length;
-  }
-  return generatorName.substring('polymer-init-'.length, nameEnd);
+  // Breakdown of regular expression to extract name (group 3 in pattern):
+  //
+  //  Pattern           | Meaning
+  //  -------------------------------------------------------------------
+  //  (generator-)?     | Group 1; Match "generator-"; Optional
+  //  (polymer-init-)?  | Group 2; Match "polymer-init-"; Optional
+  //  ([^:]+)           | Group 3; Match one or more characters != ':'
+  //  (:.*)?            | Group 4; Match ':' followed by anything; Optional
+  return generatorName.replace(/(generator-)?(polymer-init-)?([^:]+)(:.*)?/g,
+                               '$3');
 }
 
 /**
--- a/src/init/init.ts
+++ b/src/init/init.ts
@@ -82,31 +82,12 @@ function checkIsMinGW(): boolean {
 function getGeneratorDescription(
     generator: YeomanEnvironment.GeneratorMeta,
     generatorName: string): GeneratorDescription {
-  let description: string = '';
-  let genName = generatorName;
+  const meta = getGeneratorMeta(generator.resolved, generatorName, '');
+  const displayName = getDisplayName(meta.name);
+  let description = meta.description;
 
-  if (generator.resolved && generator.resolved !== 'unknown') {
-    try {
-      const metapath = findup('package.json', {cwd: generator.resolved});
-      const meta = JSON.parse(fs.readFileSync(metapath, 'utf8'));
-      description = meta.description || '';
-      genName = meta.name || generatorName;
-    } catch (error) {
-      if (error.message === 'not found') {
-        logger.debug('no package.json found for generator');
-      } else {
-        logger.debug('unable to read/parse package.json for generator', {
-          generator: generatorName,
-          err: error.message,
-        });
-      }
-    }
-  }
-
-  const name = getDisplayName(genName);
-
-  if (templateDescriptions.hasOwnProperty(name)) {
-    description = templateDescriptions[name];
+  if (templateDescriptions.hasOwnProperty(displayName)) {
+    description = templateDescriptions[displayName];
   }
 
   // If a description exists, format it properly for the command-line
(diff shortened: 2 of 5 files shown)


src/init/init.ts
```diff
@@ -112,11 +118,7 @@ function getGeneratorDescription(generator: YeomanEnvironment.GeneratorMeta, gen
  * Extract the meaningful name from the full yeoman generator name
  */
 function getDisplayName(generatorName: string) {
-  let nameEnd = generatorName.indexOf(':');
-  if (nameEnd === -1) {
-    nameEnd = generatorName.length;
-  }
-  return generatorName.substring('polymer-init-'.length, nameEnd);
+  return generatorName.replace(/(generator-)?polymer-init-([^:]+)(:.*)?/g, '$2');
```
justinfagnani: Doesn't this replace the whole string with one group? If so, I don't think replace() is necessary, and you could just return the group.

tony19: > Doesn't this replace the whole string with one group?

Yep. Example:

```
getDisplayName('generator-polymer-init-foo:aaa')  === 'foo'
getDisplayName('polymer-init-bar:bbb')            === 'bar'
getDisplayName('this-is-my-super-cool-name:ccc')  === 'this-is-my-super-cool-name'
getDisplayName('generator-polymer-init-foo')      === 'foo'
getDisplayName('polymer-init-bar')                === 'bar'
getDisplayName('this-is-my-super-cool-name')      === 'this-is-my-super-cool-name'
```

> If so, I don't think replace() is necessary, and you could just return the group.

Ok, I think you're saying you prefer this: `/(generator-)?polymer-init-([^:]+)(:.*)?/g.exec('generator-polymer-init-foo-el')[2]`. Let me know if I'm wrong here.

The reason I went with `String.replace()` is because it accepts strings that don't match the pattern (i.e., if the regex doesn't pass, it replaces nothing and returns the original string), which allows more flexibility in the display name. `RegExp.exec(str)` returns `null` if no match found, so it would need a little bit of extra code to safely get the capture group.

FredKSchott: Thanks for the detailed explanation, mind including an explanation like this of what the regex is trying to do in a comment (function documentation is fine). Complex regex can be hard to maintain when you don't know the original intent.

Otherwise I think what you have is fine.

tony19: @FredKSchott Good point 👍  I've added comments in 1d21cf5. Let me know if wordsmithing is needed.

src/init/init.ts
```diff
@@ -73,16 +73,15 @@ function checkIsMinGW(): boolean {
...
 
-  if (templateDescriptions.hasOwnProperty(name)) {
-    description = templateDescriptions[name];
-  } else if (generator.resolved && generator.resolved !== 'unknown') {
+  if (generator.resolved && generator.resolved !== 'unknown') {
     try {
       const metapath = findup('package.json', {cwd: generator.resolved});
       const meta = JSON.parse(fs.readFileSync(metapath, 'utf8'));
-      description = meta.description;
+      description = meta.description || '';
```
FredKSchott: can we still default to description here instead of `''`? Like `description = meta.description || description`;

src/init/init.ts
```diff
@@ -109,14 +115,30 @@ function getGeneratorDescription(generator: YeomanEnvironment.GeneratorMeta, gen
 }
 
 /**
- * Extract the meaningful name from the full yeoman generator name
+ * Extract the meaningful name from the full Yeoman generator name.
```
FredKSchott: Love the comments, can you just make sure they're all within 80 characters? Especially the `// -----` line. We're going to start automatically reformatting these and I don't want this comment to break when we do.

justinfagnani:
@tony19 can you rebase this one? Thanks!

tony19:
@justinfagnani Will do. I meant to address @FredKSchott's comments. I'll finish up soon (by Sat).

src/init/init.ts
```diff
@@ -73,16 +73,15 @@ function checkIsMinGW(): boolean {
  * read the description from its package.json.
  */
 function getGeneratorDescription(generator: YeomanEnvironment.GeneratorMeta, generatorName: string): GeneratorDescription {
-  const name = getDisplayName(generatorName);
   let description: string = '';
+  let genName = generatorName;
```
tony19: I was trying to avoid calling `getDisplayName()` multiple times (maybe unnecessarily so?). I'm thinking of pulling out this `try-catch` block for improved readability, which would also result in `getDisplayName()` being called only once. What do you think?

tony19: It would look something like this:
```javascript
/**
 * Get a description for the given generator. If this is an external generator,
 * read the description from its package.json.
 */
function getGeneratorDescription(generator: YeomanEnvironment.GeneratorMeta, generatorName: string): GeneratorDescription {
  const meta = getGeneratorMeta(generator.resolved, generatorName, '');
  const displayName = getDisplayName(meta.name);
  let description = meta.description;

  if (templateDescriptions.hasOwnProperty(displayName)) {
    description = templateDescriptions[displayName];
  }

  // If a description exists, format it properly for the command-line
  if (description.length > 0)  {
    description = chalk.dim(` - ${description}`);
  }

  return {
    name: `${displayName}${description}`,
    value: generatorName,
    // inquirer is broken and doesn't print descriptions :(
    // keeping this so things work when it does
    short: displayName,
  };
}

/**
 * Get the metadata of a generator from its package.json
 */
function getGeneratorMeta(rootDir: string,
                          defaultName: string,
                          defaultDescription: string): {name: string, description: string} {
  let name = defaultName;
  let description = defaultDescription;

  if (rootDir && rootDir !== 'unknown') {
    try {
      const metapath = findup('package.json', {cwd: rootDir});
      const meta = JSON.parse(fs.readFileSync(metapath, 'utf8'));
      description = meta.description || description;
      name = meta.name || name;
    } catch (error) {
      if (error.message === 'not found') {
        logger.debug('no package.json found for generator');
      } else {
        logger.debug('unable to read/parse package.json for generator', {
          generator: defaultName,
          err: error.message,
        });
      }
    }
  }
  return {name, description};
}
```

tony19: Ok, addressed in de1925353f824b628a39571e59245123c15d3763

FredKSchott:
Thanks @tony19! Code looks great, but I just ran this on my machine (I have a lot of templates installed) and it looks like there are some regressions here:

`master`: 
screen shot 2016-12-12 at 11 31 03 am

This PR:
screen shot 2016-12-12 at 11 16 52 am

It looks like custom-build has a much longer package.json name based on `polymer-starter-kit-*` which has then been forked and reused across a few templates.

There are a few different ways we could solve this, but I think the easiest right now would be to filter by `polymer-starter-kit-` in addition to `polymer-init-` and `generator-`. That way we aren't blocked by having to go hunt all these projects down and ask them to update and push out a change. wdyt?

tony19:
@FredKSchott Ok, sounds good. I've added `polymer-starter-kit-` to the prefixes. Also added some unit tests for this.

test/unit/init/init_test.js
```diff
@@ -87,15 +89,74 @@ suite('init', () => {
...
+    ];
 
     setup(() => {
+      let generators = {};
+
+      for (let g of GENERATORS) {
+        if (!g.resolved) {
+          let tmpDir = temp.mkdirSync();
+          let packageJsonPath = `${tmpDir}/package.json`;
+          fs.writeFileSync(packageJsonPath, JSON.stringify({
```
tony19: Yep, still needed. It's actually `polymer-cli` that's checking. [`getGeneratorMeta()`](https://github.com/Polymer/polymer-cli/pull/472/files#diff-6eb970fedd99f33facb11eefde800f22R119) reads the `package.json` from the directory indicated by the generator's `resolved` attribute. I'm not sure how to mock that out, but let me know.

(The link for `getGeneratorMeta()` jumps to the relevant line (119), but it's hidden underneath the topbar, so you'll have to scroll a bit to see the line :octocat: :bug:).

test/unit/init/init_test.js
```diff
@@ -87,15 +89,74 @@ suite('init', () => {
...
+        metaName: 'custom-build-3',
+        shortName: 'custom-build-3',
+        description: 'custom build 3',
+      },
+    ];
 
     setup(() => {
+      let generators = {};
+
+      for (let g of GENERATORS) {
```
FredKSchott: avoid one-letter variable names, `g -> generator`

test/unit/init/init_test.js
```diff
@@ -124,10 +185,15 @@ suite('init', () => {
         assert.equal(error.message, 'Template TEST not found');
       }).then(() => {
         let choices = promptStub.firstCall.args[0][0].choices;
-        assert.equal(choices.length, 1);
-        assert.equal(stripAnsi(choices[0].name), 'element - A blank element template');
-        assert.equal(choices[0].value, 'polymer-init-element:app');
-        assert.equal(choices[0].short, 'element');
+        assert.equal(choices.length, GENERATORS.length);
+
+        for (let c of choices) {
```
FredKSchott: same here re: 1 letter variable names

src/commands/init.ts
```diff
@@ -37,7 +37,8 @@ export class InitCommand implements Command {
 
     const templateName = options['name'];
     if (templateName) {
-      const generatorName = `polymer-init-${templateName}:app`;
+      const subgen = (templateName.indexOf(':') !== -1) ? '' : ':app';
```
justinfagnani: Use includes(): `templateName.includes(':') ? '' : ':app'`

test/unit/init/init_test.js
tony19: Ok, done

test/unit/init/init_test.js
tony19: Ok, done

src/init/init.ts
```diff
@@ -82,50 +82,86 @@ function checkIsMinGW(): boolean {
...
+  //
+  // Pattern                 | Meaning
+  // -------------------------------------------------------------------
+  // (generator-)?           | Grp 1; Match "generator-"; Optional
+  // (polymer-init)?         | Grp 2; Match "polymer-init-"; Optional
+  // (polymer-starter-kit-)? | Grp 3; Match "polymer-starter-kit-"; Optional
+  // ([^:]+)                 | Grp 4; Match one or more characters != ":"
+  // (:.*)?                  | Grp 5; Match ":" followed by anything; Optional
+  return generatorName.replace(
+      /(generator-)?(polymer-init-)?(polymer-starter-kit-)?([^:]+)(:.*)?/g,
```
justinfagnani: 1. Why do we need the `polymer-starter-kit-` prefix support? Do we have generators that begin with that? Similarly, in what cases are we asking using a generator name that still includes the `generator-` prefix?
 2. seems like the second `?` should be an '|'

justinfagnani:
Hmm, not sold on stripping `polymer-starter-kit-` then. 1) They're based on polymer-starter-kit - that's important info: `custom-build` and `polymer-starter-kit-custom-build` would be two different things. 2) If we want display name and package name to different in this many ways we should let packages custom their display name. I thought Yeoman did something like that, but it's been a while...

src/init/init.ts
tony19: The Yeoman guide ([How to add a generator](http://yeoman.io/generators/)) prescribes:

> Have the following in the module's `package.json` file:
> * The name should be `generator-your-unique-name`. This means it should begin with `generator-` and be followed by a unique (to npm) name. Scoped name support is coming in the future.
> * `yeoman-generator` in the keyword property.
> * A good description. If you forked the generator, it should describe what is different and why it was forked.

Yeoman generators typically have the `generator-` prefix in their package name, although Yeoman apparently doesn't enforce the requirements of `package.json` (only the generator's root directory name is required to contain the prefix).

In order for `getDisplayName()` to be flexible, it checks for all standard generator prefixes. I included `polymer-starter-kit-` per the suggestion in https://github.com/Polymer/polymer-cli/pull/472#issuecomment-266532084, but I'm open to other ways to resolve that problem.

tony19:
@justinfagnani Ok. Technically, this PR does allow custom package names by setting it inside the subgenerator directory's `package.json`. If that file is missing, it falls back to the root `package.json`, maintaining the original behavior.

Let me elaborate on the issue @FredKSchott points out (Fred, correct me if I'm wrong):

`PolymerElements/generator-polymer-init-custom-build` has two `package.json` files, where the subgenerator's package name differs significantly in the prefix:

  1. [`/package.json`](https://github.com/PolymerElements/generator-polymer-init-custom-build/blob/907b4aa04a2e145adda0e0899bec54fed029dc42/package.json#L2)  (prefix is `generator-polymer-init-`)

        ```json
        {
          "name": "generator-polymer-init-custom-build",
          ...
        }
        ```

  2. [`/generators/app/package.json/package.json`](https://github.com/PolymerElements/generator-polymer-init-custom-build/blob/907b4aa/generators/app/package.json#L2)  (prefix is `polymer-starter-kit-`)

        ```json
        {
          "name": "polymer-starter-kit-custom-build",
          ...
        }
        ```

The current version of Polymer CLI only uses `/package.json`, but this PR adds support for using meta from subgenerators, including `/generators/app/package.json/package.json`. This could lead to surprising results upon upgrading `polymer-cli` if users have a different name in `/generators/app/package.json/package.json`.

Given the two `package.json` names above, here's the output **before** this PR:

```
$ polymer init
? Which starter template would you like to use? 
  ...
❯ custom-build - A starting point for Polymer 1.0 custom build apps 
  custom-build-gae - A starting point for Polymer 1.0 custom build apps 
  custom-build-es6 - ES6 support for Polymer 1.0 custom build apps, written on top of Rob Dodson's polymer-starter-kit-custom-build plugin.
```

and the output **after** this PR (without stripping `polymer-starter-kit-` prefix):

```
$ polymer init
? Which starter template would you like to use? 
  ...
❯ polymer-starter-kit-custom-build - A starting point for Polymer 1.0 custom build apps 
  polymer-starter-kit-custom-build-gae - A starting point for Polymer 1.0 custom build apps 
  polymer-starter-kit-custom-build-es6 - ES6 support for Polymer 1.0 custom build apps, written on top of Rob Dodson's polymer-starter-kit-custom-build plugin. 
```

There are at least 2 published generators forked from `generator-polymer-init-custom-build` that have inherited the problem:

 * [`polymer-starter-kit-custom-build-es6`](https://github.com/jay8t6/generator-polymer-init-custom-build-es6/blob/ffe4f65/generators/app/package.json)
 * [`polymer-starter-kit-custom-build-gae`](https://github.com/chuckh/generator-polymer-init-custom-build-gae/blob/aeaa9bf/generators/app/package.json#L2).

It seems the options are:

 1. Fix it up automatically (internally stripping the prefix)
    *Note that stripping `polymer-starter-kit-` could be a problem for users who intentionally use that prefix.*
 2. Emit a warning
 3. Include an integration step in a release note

We should additionally fix the root problem in `generator-polymer-init-custom-build` to prevent the bug in new forks.

FredKSchott:
Thanks for the summary @tony19, that all sounds right to me. I think Justin's only concern is that we're losing valuable information by stripping the `polymer-starter-kit` prefix. I don't feel strongly one way or the other, my only concern at the time I brought it up was to maintain current behavior.

Since this PR adds support for (currently broken) bundled generators and doesn't change current naming behavior, I feel comfortable merging now. @justinfagnani I'll be happy to create or review a new PR afterwards that adds `polymer-starter-kit-` to those generators, replaces it with `psk-`, or anything else you'd prefer.

Thanks for  all your work on this @tony19, and for your patience as we reviewed.

justinfagnani:
I still don't like this carve out

justinfagnani:
I think if there's any issue here with the starter kit based template names, it rests with those templates. I'm not sure why the current behavior is deemed the best - it was relatively coincidental that it worked that way in the first place - but this special casing of a `polymer-starter-kit-` prefix seems like a one-off heuristic rather than solving a general problem of long template names. I als don't think it's a good idea to just strip off an indication that these templates are PSK based, or to introduce more alias potential among names. I could very well see both `custom-build` and `polymer-starter-kit-custom-build` templates that end up with the same name.

FredKSchott:
@justinfagnani understood, by merging I was separating this bug fix from the interface enhancement that it brought up. Follow up PR to address new template interface: https://github.com/Polymer/polymer-cli/pull/528

PR #477, opened 2016-11-29
Add regression tests for #471

justinfagnani:

### DIFF
--- a/test/unit/init/init_test.js
+++ b/test/unit/init/init_test.js
@@ -71,6 +71,17 @@ suite('init', () => {
       });
     });
 
+    test('works with the default yeoman environment', () => {
+      // Note: Do not use a fake Yeoman environment in this test so that we get
+      // coverage of the case where env isn't specified.
+      const UNKNOWN_GENERATOR_NAME = 'UNKNOWN-GENERATOR';
+      return polymerInit.runGenerator(UNKNOWN_GENERATOR_NAME).then(() => {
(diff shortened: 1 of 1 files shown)


justinfagnani:
One more time.

Verified that these tests fail w/o #471

test/unit/init/init_test.js
```diff
@@ -58,13 +58,10 @@ suite('init', () => {
     });
 
     test('fails if an unknown generator is requested', () => {
+      // Note: Do not use a fake Yeoman environment in this test so that we get
```
FredKSchott: good note. This may be worth mentioning in the test description as well, since this is a big part of what we're actually testing here.

test/unit/init/init_test.js
```diff
@@ -87,6 +84,14 @@ suite('init', () => {
       });
     });
 
+    test('works with default env', () => {
```
FredKSchott: this is testing the "not found" case as much as it's testing that it doesn't throw with a default env. Would be worth mentioning in the test description

justinfagnani: The not found case is covered in another test, checking not found is an implementation detail of this test.

test/unit/init/init_test.js
justinfagnani: split into two tests to each test one thing

justinfagnani:
PTAL

PR #478, opened 2016-11-30
Add install command, which installs “variants”

justinfagnani:


src/commands/install.ts
```diff
@@ -0,0 +1,34 @@
...
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+// import * as logging from 'plylog';
+import {Command, CommandOptions, ProjectConfig} from './command';
+
+// const logger = logging.getLogger('cli.install');
```
justinfagnani: removing

package.json
```diff
@@ -38,6 +38,9 @@
     "@types/vinyl": "^1.1.29",
     "@types/vinyl-fs": "0.0.28",
     "@types/yeoman-generator": "0.0.30",
+    "bower": "^1.8.0",
```
rictic: Since we're mucking about with private fields of a bower object, we should pin to a specific version

src/install/install.ts
```diff
@@ -0,0 +1,137 @@
...
+  const json = await new Promise<any>((resolve, reject) => {
+    const config = defaultBowerConfig({
+      save: false,
+    });
+    const cwd = config.cwd || process.cwd();
+    readBowerJson(cwd, {}, (err: any, json: any) => {
+      err ? reject(err) : resolve(json);
+    });
+  });
+  const variants = json['variants'];
```
rictic: How do you create a `variants` field? How do you deal with conflicts? I guess you specify all the versions of all dependencies that need overwriting in there?

src/install/install.ts
```diff
@@ -0,0 +1,137 @@
...
+}
+
+async function installDefault(): Promise<void> {
+  logger.info(`Installing Bower components`);
+  await _install();
+  logger.info(`Finished installing Bower components`);
+}
+
+async function installVariants(): Promise<void> {
+  const json = await new Promise<any>((resolve, reject) => {
```
FredKSchott: can this get a more descriptive variable name? Hard to track which `json` object this represents below.

src/install/install.ts
```diff
@@ -0,0 +1,137 @@
...
+    // names in `from` to control merging:
+    //   * "key=" would always overwrite, not merge, the property
+    //   * "key|" could force a union (merge), even for Arrays
+    //   * "key&" could perform an intersection
+    const fromValue = from[key];
+    const toValue = to[key];
+    if (isPrimitiveOrArray(fromValue) || isPrimitiveOrArray(toValue)) {
+      merged[key] = fromValue;
+    } else {
+      merged[key] = mergeJson(fromValue as JsonObject, toValue as JsonObject);
```
FredKSchott: I don't love checking for things that an object isn't (primitive or array) and then assuming what it *is* based on that. Could we instead run`mergeJson()` on anything that should be merged, and then return the fromValue for everything else?

src/install/install.ts
```diff
@@ -0,0 +1,140 @@
...
+
+export async function install(): Promise<void> {
+  await Promise.all([
+    installDefault(),
+    installVariants(),
+  ]);
+}
+
+function _install(bowerJson?: JsonObject, directory?: string): Promise<void> {
+  return new Promise(async (resolve, reject) => {
```
justinfagnani: Just removed. It was there because I had `await`s previously.

package.json
justinfagnani: Made it a `~`, presuming we might catch a useful patch in the CI. I'll look into repeatable installs very soon.

src/commands/install.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export class InstallCommand implements Command {
+  name = 'install';
+
+  // TODO(justinfagnani): Expand and link to eventual doc on variants.
+  description = 'installs Bower dependencies, optionally installing "variants"';
+
+  args = [
+    {
+      name: 'variants',
```
justinfagnani: Oh, I forgot about this, I'm going to wire it up.

src/install/install.ts
```diff
@@ -0,0 +1,137 @@
...
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+// <reference path="../../custom_typings/bower.d.ts" />
```
justinfagnani: indeed. cargo culted this part

src/install/install.ts
```diff
@@ -0,0 +1,137 @@
...
+interface JsonArray extends Array<JsonValue> {}
+
+export async function install(): Promise<void> {
+  await Promise.all([
+    installDefault(),
+    installVariants(),
+  ]);
+}
+
+function _install(bowerJson?: JsonObject, directory?: string): Promise<void> {
```
justinfagnani: commented and renamed

src/install/install.ts
```diff
@@ -0,0 +1,137 @@
...
+      directory,
+    });
+    const cwd: string = config.cwd || process.cwd();
+    const renderer = new StandardRenderer('install', {
+      cwd,
+      color: true,
+    });
+    const bowerLogger = new BowerLogger();
+
+    const project = new BowerProject(config, bowerLogger);
```
justinfagnani: reordered lines a bit

src/install/install.ts
```diff
@@ -0,0 +1,137 @@
...
+      renderer.end(data);
+      resolve();
+    });
+    bowerLogger.on('error', (err: Error) => {
+      renderer.error(err);
+      reject(err);
+    });
+
+    // This is the only way I could find to provide a JSON object to the
+    // Project. It's a hack, and might break in the future, but it works.
```
justinfagnani: done

src/install/install.ts
```diff
@@ -0,0 +1,137 @@
...
+    if (bowerJson) {
+      (project as any)._json = bowerJson;
+      (project as any)._jsonFile = path.join(cwd, 'bower.json');
```
justinfagnani: nope, and without this... 💥

src/install/install.ts
justinfagnani: done

src/install/install.ts
```diff
@@ -0,0 +1,137 @@
...
+      save: false,
+    });
+    const cwd = config.cwd || process.cwd();
+    readBowerJson(cwd, {}, (err: any, json: any) => {
+      err ? reject(err) : resolve(json);
+    });
+  });
+  const variants = json['variants'];
+  if (variants) {
+    await Promise.all(Array.from(Object.keys(variants)).map(async (variantName) => {
```
justinfagnani: Because JavaScript sucks and `Map.prototype.keys` returns an iterator, so I thought `Object.keys` would too, and iterators don't have `.map()`.

src/install/install.ts
```diff
@@ -0,0 +1,137 @@
...
+      const variant = variants[variantName];
+      const variantBowerJson = mergeJson(variant, json);
+      const variantDirectory = `bower_components-${variantName}`;
+      logger.info(`Installing variant ${variantName} to ${variantDirectory}`);
```
justinfagnani: done

src/install/install.ts
justinfagnani: This is the only type left that it can be. I tried to type `Json` as a discriminated union, but TS doesn't allow recursive type aliases.

package.json
justinfagnani: done

justinfagnani:
PTAL

src/install/install.ts
FredKSchott: [shame](https://m.popkey.co/ecc8de/6GJWk_f-thumbnail-100-0.jpg)

src/install/install.ts
FredKSchott: yea, but JavaScript... and user-provided data... ☠️ 

I only bring it up because I've been bit by this in the past when at the time it was impossible. If it's not hard to do, I'd still prefer to see it organized that way.

package.json
```diff
@@ -38,6 +38,9 @@
     "@types/vinyl": "^1.1.29",
     "@types/vinyl-fs": "0.0.28",
     "@types/yeoman-generator": "0.0.30",
+    "bower": "~1.8.0",
```
FredKSchott: I'd still like to see us pin to a specific version. I know plenty of big projects that would treat a refactoring of files or private properties as a patch no-breaking-changes bug fix.

src/commands/install.ts
FredKSchott: 👍

src/install/install.ts
```diff
@@ -0,0 +1,138 @@
...
+
+async function installDefault(): Promise<void> {
+  logger.info(`Installing Bower components...`);
+  await _install();
+  logger.info(`Finished installing Bower components`);
+}
+
+async function installVariants(): Promise<void> {
+  const bowerJson = await new Promise<any>((resolve, reject) => {
+    const config = defaultBowerConfig({
```
FredKSchott: nit: 1 line (I could have sworn we had clang set up in the CLI)

package.json
rictic: +1

src/install/install.ts
rictic: Would you mind writing up just a small amount of documentation of it here for now?

package.json
justinfagnani: done

src/install/install.ts
justinfagnani: I'll do the change, just want to point out that it's the exact same logic, it just moves the conditional out of the loop - iow: push it one level down in recursion. We still have to assume that if an object is not a primitive or array that it's a JsonObject, and we still have to cast. And then, because _mergeJson now deals in all JSON types, we have to cast the result in the only place we use it, in installVariants.

The "danger" of getting a type that's outside of our declared types is exactly the same. It's also not quite user-provided data - the user-provided data is a string, the JSON values are coming from Bowers config loading API, which claims it's only returning JSON values.

Here's the change: same number of casts internally, one less else, one more cast externally.

```typescript
export function _mergeJson(from: JsonValue, to: JsonValue): JsonValue {
  if (isPrimitiveOrArray(from) || isPrimitiveOrArray(to)) {
    return from;
  }
  // Now we must assume from and to are `JsonObject`s

  // First, make a shallow copy of `to` target
  const merged = Object.assign({}, to as JsonObject);

  // Next, merge in properties from `from`
  for (const key in (from as JsonObject)) {
    // TODO(justinfagnani): If needed, we can add modifiers to the key
    // names in `from` to control merging:
    //   * "key=" would always overwrite, not merge, the property
    //   * "key|" could force a union (merge), even for Arrays
    //   * "key&" could perform an intersection

    merged[key] = _mergeJson(from, to);
  }
  return merged;
}
```

justinfagnani: Actually, I didn't try to compile that. It really needs to be like this:

```typescript
export function _mergeJson(from: JsonValue, to: JsonValue): JsonValue {
  if (isPrimitiveOrArray(from) || isPrimitiveOrArray(to)) {
    return from;
  }
  const toObject = to as JsonObject;
  const fromObject = from as JsonObject
  // First, make a shallow copy of `to` target
  const merged = Object.assign({}, toObject);

  // Next, merge in properties from `from`
  for (const key in fromObject) {
    // TODO(justinfagnani): If needed, we can add modifiers to the key
    // names in `from` to control merging:
    //   * "key=" would always overwrite, not merge, the property
    //   * "key|" could force a union (merge), even for Arrays
    //   * "key&" could perform an intersection
    merged[key] = _mergeJson(fromObject[key], toObject[key]);
  }
  return merged;
}
```

FredKSchott: why not keep `mergeJson()` as-is, and just change it's invocation to:

```js
if (isJsonObject(fromValue) && isJsonObject(toValue)) {
  merged[key] = mergeJson(fromValue as JsonObject, toValue as JsonObject);
} else {
  merged[key] = fromValue;
}
```

justinfagnani:
PTAL

src/commands/install.ts
justinfagnani: done

src/install/install.ts
justinfagnani: done, but looking to write up real docs for the site & readme's today

src/install/install.ts
```diff
@@ -0,0 +1,148 @@
...
+      logger.info(`Finished installing variant ${variantName}`);
+    }));
+  }
+}
+
+/**
+ * Exported only for testing
+ */
+export function _mergeJson(from: JsonValue, to: JsonValue): JsonValue {
+  if (isPrimitiveOrArray(from) || isPrimitiveOrArray(to)) {
```
FredKSchott: Continuing the conversation here since github is crazy and won't let me comment in original thread.

> What would the implementation of isJsonObject look like?

`return typeof x === 'object' && !Array.isArray(x)` (You can add a check for null in there as well, I wasn't sure if that was intentionally meant to be run through `mergeJson()` in your implementation or not.)

src/install/install.ts
```diff
@@ -0,0 +1,148 @@
...
+    // names in `from` to control merging:
+    //   * "key=" would always overwrite, not merge, the property
+    //   * "key|" could force a union (merge), even for Arrays
+    //   * "key&" could perform an intersection
+    merged[key] = _mergeJson(fromObject[key], toObject[key]);
+  }
+  return merged;
+}
+
+function isPrimitiveOrArray(value: any) {
```
FredKSchott: Continuing the conversation here since github is crazy and won't let me comment in original thread.

> What would the implementation of isJsonObject look like?

```
return typeof x === 'object' && !Array.isArray(x) && x !== null
```

That would also clarify the case where `isPrimitiveOrArray(null)` currently returns true.

FredKSchott:
also looks like tests are failing

justinfagnani:
Those failures are flaky temp-directory path comparisons. No idea why they're happening more frequently now, but they almost always pass on the next run.

src/install/install.ts
FredKSchott: talked in person, feeling better about this not being strictly user input and that `from = to` isn't a general catch all. Happy with either at this point 👍

FredKSchott:
push typo fix before merging

PR #489, opened 2016-12-05
Fix type warnings

rictic:


package.json
```diff
@@ -62,7 +62,7 @@
     "polymer-analyzer": "2.0.0-alpha.18",
     "polymer-build": "^0.5.1",
     "polymer-project-config": "^1.0.0",
-    "polyserve": "0.16.0-prerelease.3",
+    "polyserve": "^0.16.0-prerelease.4",
```
justinfagnani: I think we want to pin exactly for now, would could have some breaking changes before release.

src/build/build.ts
```diff
@@ -115,7 +115,7 @@ export async function build(options: BuildOptions, config: ProjectConfig): Promi
 
   // Once the unbundled build stream is complete, create a service worker for the build
   let unbundledPostProcessing = Promise.all([loadSWConfig, waitFor(unbundledPhase)]).then((results) => {
-    let swConfig: SWConfig = results[0];
+    let swConfig: SWConfig|undefined = results[0] || undefined;
```
justinfagnani: what error were you seeing here?

justinfagnani: Why weren't we seeing this locally or an the CI?

rictic: Hm. Oh, I'm using a prerelease version of the compiler, I should switch back to stable.

package.json
rictic: Done

rictic:
Pinned, also fixed one other type temporarily

src/build/build.ts
FredKSchott: @rictic still needed after pinning?

rictic: The stable compiler doesn't complain, but this code is more correct than before. `parsePreCacheConfig` returns `Promise<SWConfig|null>` and `addServiceWorker` wants `SWConfig|undefined`

rictic:
I'm just gonna block this on https://github.com/PolymerLabs/polyserve/pull/136 because prerelease 5 has old type definitions.

PR #490, opened 2016-12-05
Fix serve command to pass all options to polyserve

justinfagnani:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "polymer-cli",
-  "version": "0.18.0-alpha.5",
```
FredKSchott: don't do this here, bump in separate commit/PR

src/commands/serve.ts
```diff
@@ -61,7 +78,23 @@ export class ServeCommand implements Command {
...
+      applications: ${
+                    url.format(urls.serverUrl)}
+      reusable components: ${url.format(urls.componentUrl)}
+    `);
+    } else {
+      // We started multiple servers, just tell the user about the control server,
+      // it serves out human-readable info on how to access the others.
+      const urls = getServerUrls(options, serverInfos.control.server);
+      console.log(`Started multiple servers with different variants:
+      More info here: ${url.format(urls.serverUrl)}`);
```
FredKSchott: I don't love this wording, "More info here" feels like something went wrong / there's something I need to read to fix.

How about something shorter like: `'Started Polyserve on ${url.format(urls.serverUrl)}'`?

src/commands/serve.ts
```diff
@@ -30,9 +30,12 @@ export class ServeCommand implements Command {
 
   args = polyserveArgs;
 
-  run(options: CommandOptions, config: ProjectConfig): Promise<any> {
+  async run(options: CommandOptions, config: ProjectConfig): Promise<any> {
     // Defer dependency loading until this specific command is run
     const polyserve = require('polyserve');
+    const startServers = polyserve.startServers;
+    const getServerUrls = polyserve.getServerUrls;
+    const url = require('url');
```
FredKSchott: fwiw `url` is probably fine to import at the top of the file, it's small and I'm sure it's already been imported somewhere else at this point. Fine if you want to keep it here though.

package.json
justinfagnani: This is in it's own commit in this PR, it'll show up in the commits list.

package.json
```diff
@@ -62,7 +62,7 @@
     "polymer-analyzer": "2.0.0-alpha.18",
     "polymer-build": "^0.5.1",
     "polymer-project-config": "^1.0.0",
-    "polyserve": "0.16.0-prerelease.3",
+    "polyserve": "0.16.0-prerelease.5",
```
rictic: prerelease 6

rictic:
Please rebase and CI should be happy

PR #492, opened 2016-12-07
import fixups, and adding missing licence headers

FredKSchott:
Looks like these weird imports were just copy-pasted around the different commands.
/cc @rictic @justinfagnani

### DIFF
--- a/src/commands/command.ts
+++ b/src/commands/command.ts
@@ -1,7 +1,19 @@
-import {ArgDescriptor} from 'command-line-args';
+/**
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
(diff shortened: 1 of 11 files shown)


src/commands/command.ts
```diff
@@ -1,7 +1,19 @@
...
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
 
+import {ArgDescriptor} from 'command-line-args';
 import {ProjectConfig} from 'polymer-project-config';
-export {ProjectConfig} from 'polymer-project-config';
```
justinfagnani: I don't think this is a particularly bad export. It's fine to export something that you assert is part of your public contract, and in fact insulates the client from changes somewhat, say in the case that commands extended the config interface somehow.

FredKSchott: Agreed on the general principal of re-exporting imports. but there's no reason for ProjectConfig to come out of `command.ts` here. The CLI has no public contract, and ProjectConfig is really mainly used and set up in the main `polymer-cli.ts`.

I don't think there's a good case here for exporting types unnecessarily within the CLI's internal files.

PR #495, opened 2016-12-07
Add variants and install to README

justinfagnani:


README.md
```diff
@@ -89,6 +89,90 @@ You can specify the project files in `polymer.json` so that commands like `polym
...
+    "polymer-2": {
+      "dependencies": {
+        "polymer": "Polymer/polymer#^2.0.0"
+      }
+    }
+  }
+}
+```
+
+When using `polymer install`, this results in two additional folders,
```
rictic: Probably best to mention that you have to pass `--variants` in to `polymer install` for this behavior right off the bat. e.g. 

```
When using `polymer install --variants` ...
```

justinfagnani: oops!

PR #497, opened 2016-12-08
Add a npm-shinkwrap.json file

justinfagnani:
This should get us more repeatable tests and bug reports, and reduces npm install times by ~15%.

`np` will regenerate the shrinkwrap for every release if one already exists, so this gets things started.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,8 +8,8 @@ install:
   - ps: Install-Product node $env:nodejs_version
   # Get npm 3
   - npm install -g npm
-  # install modules
-  - npm install
+  # install modules. Use --force to workaround permissions issues on Windows
+  - npm install --force
 
 # Post-install test scripts.
 test_script:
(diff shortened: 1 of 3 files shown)


ggirard07:
This shrinkwrap file actually prevent users to install polymer-cli on a windows machine using NPM 3.10.8...
My use case: https://github.com/Microsoft/vsts-tasks/issues/6395
Original issue: https://github.com/npm/npm/issues/2679

justinfagnani:
I believe we should be able to fix that by adding `"fsevents": "^1.0.0"` to our package.json.

ggirard07:
Not much experience with npm here and so don't really understand why this will allow the dependency to be seen as optional for Windows based users with this change.
`fsevent` will still be listed as a dependency in `npm-shrinkwrap.json` and so every user using npm 3.10.8 will still be forced to install it as optional dependencies are not well handled by this version.
I think there is no real solution here except that end users should never use npm 3.10.8!

PR #503, opened 2016-12-09
print error.message with optimize errors

### LINKED ISSUE #502: warn: Unable to optimize .html file /path/to/my-elem.html { err: {} }
_From @bennypowers on November 21, 2016 21:8_

### Description

When running `polymer build` in my PSK-based project, I get this output:
```
info:    Preparing build...
info:    Building application...
info:    Generating build/unbundled...
info:    Generating build/bundled...
warn:    Unable to optimize .html file /Users/benny/web/ux/emr/src/exams/ux-emr-exam-va.html
{ err: {} }
warn:    Unable to optimize .html file /Users/benny/web/ux/emr/src/ux-emr-chart.html
{ err: {} }
info:    Build complete!
```
### Versions & Environment

- polymer-build: 0.4.1
- node: 6.9.1
- Operating System: el capitan

#### Steps to Reproduce

`polymer build`

#### Expected Results

No warning or verbose error
#### Actual Results

warning with silent error

_Copied from original issue: Polymer/polymer-build#82_

FredKSchott:
Fixes #502

### DIFF
--- a/src/build/optimize-streams.ts
+++ b/src/build/optimize-streams.ts
@@ -60,7 +60,7 @@ export class GenericOptimizeStream extends Transform {
       } catch (error) {
         logger.warn(
             `Unable to optimize ${this.validExtension} file ${file.path}`,
-            {err: error});
+            {err: error.message || error});
       }
     }
     callback(null, file);


src/build/optimize-streams.ts
```diff
@@ -60,7 +60,7 @@ export class GenericOptimizeStream extends Transform {
       } catch (error) {
         logger.warn(
             `Unable to optimize ${this.validExtension} file ${file.path}`,
-            {err: error});
+            {err: error.message || error});
```
rictic: If the problem is likely to be internal, `error.stack || error.message || error` is even better.

If the error message is most likely telling the user about a mistake that they made, then this form is better.

FredKSchott: The second one :) we expect the error to be from a parse error in the file, and not in the tool itself. So the message the tool gives us will be more important to the user, who can use it to find and fix their error.

rictic: I see. Ideally in that sort of case we'd have also already emitted a nice Warning from the analyzer too.

PR #509, opened 2016-12-14
Convert build to single build stream

FredKSchott:
- [x] changelog updated

/cc @justinfagnani

### DIFF
--- a/src/build/build.ts
+++ b/src/build/build.ts
@@ -34,7 +34,7 @@ const buildDirectory = 'build/';
 
 export interface BuildOptions {
   swPrecacheConfig?: string;
-  insertDependencyLinks?: boolean;
+  insertPrefetchLinks?: boolean;
   bundle?: boolean;
   // TODO(fks) 07-21-2016: Fully complete these with available options
   html?: {collapseWhitespace?: boolean; removeComments?: boolean};
@@ -55,11 +55,6 @@ export async function build(
     css: Object.assign({stripWhitespace: true}, options.css),
     js: Object.assign({minify: true}, options.js),
   };
-
-  if (options.insertDependencyLinks) {
-    logger.debug(
-        `Additional dependency links will be inserted into application`);
-  }
 
   logger.info(`Deleting build/ directory...`);
   await del([buildDirectory]);
@@ -96,10 +91,10 @@ export async function build(
 
   if (options.bundle) {
     buildStream = buildStream.pipe(polymerProject.bundler);
-  } else {
-    buildStream = buildStream.pipe(gulpif(
-        options.insertDependencyLinks || false,
-        new PrefetchTransform(polymerProject)));
+  }
+
+  if (options.insertPrefetchLinks) {
+    buildStream = buildStream.pipe(new PrefetchTransform(polymerProject));
   }
 
   buildStream = buildStream
--- a/src/build/build.ts
+++ b/src/build/build.ts
@@ -19,7 +19,7 @@ import * as logging from 'plylog';
 import {dest} from 'vinyl-fs';
 
 import mergeStream = require('merge-stream');
-import {PolymerProject, addServiceWorker, forkStream, SWConfig} from 'polymer-build';
+import {PolymerProject, addServiceWorker} from 'polymer-build';
 
 import {InlineCSSOptimizeStream, JSOptimizeStream, CSSOptimizeStream, HTMLOptimizeStream} from './optimize-streams';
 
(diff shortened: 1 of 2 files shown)


abdonrd:
@FredKSchott changelog update?

FredKSchott:
@abdonrd good point! updated

src/build/build.ts
```diff
@@ -59,10 +56,13 @@ export async function build(
...
-  logger.info(`Preparing build...`);
-  await del([unbundledBuildDirectory, bundledBuildDirectory]);
+  if (options.insertDependencyLinks) {
+    logger.debug(
+        `Additional dependency links will be inserted into application`);
+  }
 
-  logger.info(`Building application...`);
+  logger.info(`Deleting build/ directory...`);
+  await del([buildDirectory]);
```
justinfagnani: I think cleaning the build directory should be behind a flag

src/build/build.ts
```diff
@@ -82,66 +82,45 @@ export async function build(
...
-                         .pipe(dest(bundledBuildDirectory));
+  let buildStream =
+      <NodeJS.ReadableStream>mergeStream(sourcesStream, depsStream)
+          .once('data', () => {
+            logger.debug('Analyzing build dependencies...');
+          });
+
+  if (options.bundle) {
+    buildStream = buildStream.pipe(polymerProject.bundler);
+  } else {
```
justinfagnani: I don't think this should be an else, let them both be enabled if requested.

src/build/build.ts
```diff
@@ -82,66 +82,45 @@ export async function build(
...
+    buildStream = buildStream.pipe(gulpif(
+        options.insertDependencyLinks || false,
+        new PrefetchTransform(polymerProject)));
+  }
+
+  buildStream = buildStream
+                    .once(
```
justinfagnani: I think these could be split up into separate statements and maybe improve the formatting. `.pipe`

src/build/build.ts
FredKSchott: Ehh, I've been burned enough by this to think it should be default. Any time any dependency path, source file path, dependency imports, or source imports change, you end up with ghost files in your build. It adds ~500ms on my machine, which is well worth it IMO.

src/build/build.ts
FredKSchott: agreed, made that change in the next PR (https://github.com/Polymer/polymer-cli/pull/510/files#diff-559c8e16c25ba5b201e80baf4b8fc080R90) Just wanted to match current behavior at this point while I was working through.

src/build/build.ts
FredKSchott: What do you have in mind?

I fiddled around with clang-format on this line a bit, but had no luck. The same thing happens if I use `function() {` instead of `() => {`.

FredKSchott:
@justinfagnani PTAL

src/build/build.ts
justinfagnani: And I've been burned by auto-cleaning, which has removed files I needed, especially when combined with other bugs.

src/build/build.ts
```diff
@@ -88,66 +83,45 @@ export async function build(
...
-    }
+  if (options.bundle) {
+    buildStream = buildStream.pipe(polymerProject.bundler);
+  }
+
+  if (options.insertPrefetchLinks) {
+    buildStream = buildStream.pipe(new PrefetchTransform(polymerProject));
+  }
+
+  buildStream = buildStream
```
justinfagnani: what I was thinking here is:

```
buildStream = buildStream.pipe(dest(buildDirectory));
buildStream.once('data', () => {
  ...
});
```

src/build/build.ts
```diff
@@ -45,11 +45,8 @@ export interface BuildOptions {
 export async function build(
     options: BuildOptions, config: ProjectConfig): Promise<void> {
   let polymerProject = new PolymerProject(config);
-
-  if (options.insertDependencyLinks) {
-    logger.debug(
-        `Additional dependency links will be inserted into application`);
-  }
+  let swPrecacheConfig = path.resolve(
```
justinfagnani: can these be `const` here?

src/build/build.ts
```diff
@@ -88,66 +83,45 @@ export async function build(
...
+  if (swConfig) {
+    logger.debug(`Service worker config found`, swConfig);
+  } else {
+    logger.debug(`No service worker configuration found at ${swPrecacheConfig
+                  }, continuing with defaults`);
+  }
+
+  // Once the build stream is complete, create a service worker for the build
+  await waitFor(buildStream);
+  await addServiceWorker({
```
justinfagnani: I think a comment on this line would clarify the waiting. Something like: addServiceWorker reads from the filesystem, so must run after files are written to disk

src/build/build.ts
```diff
@@ -88,66 +83,45 @@ export async function build(
...
+    buildRoot: buildDirectory,
+    project: polymerProject,
+    swPrecacheConfig: swConfig || undefined,
```
justinfagnani: `swConfig || undefined` is a little awkward, can it be fixed in `addServiceWorkers`'s or `swConfig`'s declarations?

src/build/build.ts
```diff
@@ -88,66 +83,45 @@ export async function build(
...
+    bundled: options.bundle
```
justinfagnani: style nit: add trailing comma. Did you run format?

FredKSchott: added. but format doesn't add trailing commas, lets add a tslint rule?

PR #510, opened 2016-12-14
Clarify `--add-prefetch` build flag

FredKSchott:
Clarified this build flag since a lot of people (myself included) weren't sure what it was doing. Also confirmed it's still working as intended.

/cc @justinfagnani

### DIFF
--- a/src/build/build.ts
+++ b/src/build/build.ts
@@ -34,7 +34,7 @@ const buildDirectory = 'build/';
 
 export interface BuildOptions {
   swPrecacheConfig?: string;
-  insertDependencyLinks?: boolean;
+  prefetchDepedencies?: boolean;
   bundle?: boolean;
   // TODO(fks) 07-21-2016: Fully complete these with available options
   html?: {collapseWhitespace?: boolean; removeComments?: boolean};
@@ -55,11 +55,6 @@ export async function build(
     css: Object.assign({stripWhitespace: true}, options.css),
     js: Object.assign({minify: true}, options.js),
   };
-
-  if (options.insertDependencyLinks) {
-    logger.debug(
-        `Additional dependency links will be inserted into application`);
-  }
 
   logger.info(`Deleting build/ directory...`);
   await del([buildDirectory]);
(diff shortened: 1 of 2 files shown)


src/build/build.ts
```diff
@@ -34,7 +34,7 @@ const buildDirectory = 'build/';
 
 export interface BuildOptions {
   swPrecacheConfig?: string;
-  insertDependencyLinks?: boolean;
+  prefetchDepedencies?: boolean;
```
justinfagnani: "insert" seemed meaningful here, in that the effect of this flag is to insert prefetch link tags, the effect isn't to do prefetching.

src/commands/build.ts
```diff
@@ -45,11 +45,12 @@ export class BuildCommand implements Command {
           'used for service worker generation.'
     },
     {
-      name: 'insert-dependency-links',
+      name: 'add-prefetch',
       type: Boolean,
-      description: 'Flatten dependency tree downloads by inserting ' +
-          'additional `<link rel="prefetch">` tags into ' +
-          'entrypoint and `<link rel="import">` tags into fragments and shell.'
+      description: 'Add dependency tree prefetching by inserting ' +
```
justinfagnani: I'd remove "tree", we haven't used the term "dependency tree" before. Maybe:

"Enable dependency prefetching by inserting..."

src/commands/build.ts
```diff
@@ -45,11 +45,12 @@ export class BuildCommand implements Command {
           'used for service worker generation.'
     },
     {
-      name: 'insert-dependency-links',
+      name: 'add-prefetch',
```
justinfagnani: similar to the above comment,  I think "insert" is meaningful here. Maybe `insert-prefetch-links`?

FredKSchott: I prefer the shorter `add-prefetch` when it comes to command line args. And `insert` always felt like it was inserting something new into my application while `add` feels like it's adding a new feature/ability to my application. So I prefer as-is, but happy to make the change if you feel strongly.

src/build/build.ts
FredKSchott: Good point, changed naming to `addPrefetchLinks` to be in line with the cli flag.

src/commands/build.ts
justinfagnani: I'm not really concerned about long flag names for things that'll hopefully be in a config file, if they're more descriptive. I think `insert-prefetch-links` is the most clear name.

FredKSchott: fine by me, I'll make the change

FredKSchott:
@justinfagnani name changed, PTAL

PR #511, opened 2016-12-14
Add `--compile` flag to babelify your build

FredKSchott:
I figured out how to load and use presets within the CLI (not super obvious). This lets users already using babel provide their own `.babelrc` file if they want, but we will default to the basic `es2015` transformation for everyone else.

/cc @justinfagnani

### DIFF
--- a/src/build/build.ts
+++ b/src/build/build.ts
@@ -13,15 +13,17 @@
  */
 
 import * as del from 'del';
+import * as fs from 'fs';
 import * as gulpif from 'gulp-if';
 import * as path from 'path';
 import * as logging from 'plylog';
+import {PassThrough} from 'stream';
 import {dest} from 'vinyl-fs';
 
 import mergeStream = require('merge-stream');
 import {PolymerProject, addServiceWorker} from 'polymer-build';
 
-import {JSOptimizeStream, CSSOptimizeStream, HTMLOptimizeStream} from './optimize-streams';
+import {JSOptimizeStream, CSSOptimizeStream, JSBabelStream, HTMLOptimizeStream} from './optimize-streams';
 
 import {ProjectConfig} from 'polymer-project-config';
 import {PrefetchTransform} from './prefetch';
@@ -36,6 +38,7 @@ export interface BuildOptions {
   swPrecacheConfig?: string;
   prefetchDepedencies?: boolean;
   bundle?: boolean;
+  compile?: boolean;
   // TODO(fks) 07-21-2016: Fully complete these with available options
   html?: {collapseWhitespace?: boolean; removeComments?: boolean};
   css?: {stripWhitespace?: boolean};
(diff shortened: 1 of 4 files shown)


larsgk:
Will this be available for 1.x builds also?

FredKSchott:
@larsgk If you mean builds of applications that use Polymer 1.x then yes, this will work for all versions of Polymer and is working on Polymer 1.x elements/applications today.

larsgk:
Works with the current release? (With the flag?) I didn't see it was merged
and released yet - will check again

larsgk:
... because by default, (my current/latest) polymer build fails on uglify
if any ES2015 code is present anywhere

src/build/build.ts
```diff
@@ -59,10 +62,35 @@ export async function build(
   logger.info(`Deleting build/ directory...`);
   await del([buildDirectory]);
 
+  let babelConfigJson = null;
+  if (options.compile) {
+    try {
+      const babelConfigPath = path.resolve(config.root, '.babelrc');
```
justinfagnani: let's not read the babelrc. Please use the hard-coded list of presets here: https://github.com/PolymerLabs/polyserve/blob/master/src/compile-middleware.ts#L24

src/build/build.ts
```diff
@@ -59,10 +62,35 @@ export async function build(
...
+            {err: err.message});
+      }
+    }
+  }
+
   logger.debug(`Reading source files...`);
   let sourcesStream =
       polymerProject.sources()
           .pipe(polymerProject.splitHtml())
+          .pipe(gulpif(
```
justinfagnani: we need to compile inline scripts as well

src/build/optimize-streams.ts
```diff
@@ -12,12 +12,14 @@
...
  */
 
+import {transform as babelTransform, TransformOptions as BabelTransformOptions} from 'babel-core';
 import {css as cssSlam} from 'css-slam';
 import {minify as htmlMinify, Options as HTMLMinifierOptions} from 'html-minifier';
 import * as logging from 'plylog';
 import {Transform} from 'stream';
 import {minify as uglify, MinifyOptions as UglifyOptions} from 'uglify-js';
 
+const babelPresetES2015 = require('babel-preset-es2015');
```
justinfagnani: we can't use this preset because it compiles modules

src/build/build.ts
FredKSchott: isn't that what `splitHtml()` allows us to do here? Not sure I follow

src/build/optimize-streams.ts
FredKSchott: ah right. What was the reason we need to keep modules again?

FredKSchott:
Superseded by https://github.com/Polymer/polymer-cli/pull/525

PR #519, opened 2016-12-17
Lazy load big dependencies, reducing file size to 118MB production install

### LINKED ISSUE #364: Polymer CLI install /w dependencies is ~180MB, takes ~2min
I could have sworn we already had an issue tracking this, but right now I can't find it. 

(Recent context: https://twitter.com/heisenbugger/status/763084296605294592)

## Current Install Performance

Here are some rough stats on our CLI install times. Before each test the `node_modules/` directory was removed and the npm cache was cleaned via `npm cache clean`. 

This was done from my home wifi, so take install time information with a grain of salt. 

### v0.14.0
```
$ time npm install --production
real	2m29.382s
user	0m42.985s
sys	0m11.895s
$ du -sh node_modules/
182M    node_modules/
```

### v0.14.0 without `test` dependencies (web-component-tester)
```
$ time npm install --production
real	1m46.200s
user	0m27.520s
sys	0m7.205s
$ du -sh node_modules/
89M     node_modules/
```

### v0.14.0 without `build` dependencies
```
$ time npm install --production
real	2m19.393s
user	0m40.012s
sys	0m11.349s
$ du -sh node_modules/
169M	node_modules/
```

### v0.14.0 without `init` dependencies
```
$ time npm install --production
real	2m8.023s
user	0m36.747s
sys	0m9.901s
$ du -sh node_modules/
157M	node_modules/
```

### v0.14.0 without any command-specific dependencies
```
$ time npm install --production
real	0m15.419s
user	0m4.889s
sys	0m1.314s
$ du -sh node_modules/
13M     node_modules/
```

### Turn WCT into an optional dependency

When the user first runs `polymer test` command, they'll see a "please install web-component-tester" message with instructions for how to install. We could also prompt the user to install WCT for them automatically.

This would reduce our installs by ~45 seconds and almost 100MB, and doesn't require any major refactoring on our part. It requires a one-time additional step on our users part when they first use the test command.

### "heavy" commands exist as separate dependencies

When a user first runs some commands (ex: `polymer build`) they'll see a "please install polymer-cli-build" message with instructions for how to install. This would be a new dependency that contains the pulled-out build command logic. We could also prompt the user to install this package for them automatically.

This would reduce our installs by ~2 minutes and ~150MB, but would require some refactoring on our part and an additional one-time step for each command on our users' part.

### Look into lazy-installing command-specific dependencies

There appears to be some experimentation going on in the community with lazy-installing dependencies. (See [lazy-install](https://www.npmjs.com/package/lazy-install) & [lazy-require](https://github.com/bevry/lazy-require)). It seems more fragile and only works because it completely circumvents NPM's normal installation logic. But there could be something interesting here.

This would reduce our installs by ~2 minutes and ~150MB, but would require some investigation and eventually refactoring on our part.

## My Thoughts

Turning WCT into an optional dependency is a good first step towards tackling our big install problem. It gives our users a big win without much additional work for either us or them. It would be nice to have some numbers on how many of our users are using `polymer test`, but I'd guess it's a relatively small number compared to our overall install numbers.

The other two solutions I outlined above would result in even bigger speed ups, but they seem like overkill for right now.

/cc @justinfagnani @robdodson @addyosmani @tjsavage

TimvdLippe:
This PR is a proposal for lazy-loading big dependencies of the CLI as a result of a discussion yesterday on Slack.

`npm install --production` before: 257MB
`npm install --production` after: 118MB

Total file size reduction roughly 55%.

## Requiring global packages

Removing the dependencies on `web-component-tester` and `yeoman-generator` requires loading the packages in the global npm package space. Natively `require(...)` does not do this, however the package [`requireg`](https://github.com/h2non/requireg) contains a resolver that does load the global package. It will first try to fetch locally and if that fails (e.g. not a dev environment) try to fetch globally in multiple possible places. The package is 580KB which is not too bad I think.

## User workflow

Initially installing the CLI will therefore not install the dependencies. However, when the user runs `polymer init` or `polymer test`, a helpful message is shown that the packages are required to be installed.

### How to test this branch

1. `rm -rf node_modules`
2. `npm install`
3. `npm run build`
4. `rm -rf node_modules`
5. `npm install --production`
6. `npm uninstall -g yeoman-generator yeoman-environment`
7. `node bin/polymer.js init`
8. Observe that the message "error:   cli runtime exception: Polymer init requires `yeoman-generator` and `yeoman-environment` to be installed globally. Install with `npm install -g i yeoman-generator yeoman-environment`" is shown in the console
9. `npm install -g yeoman-generator yeoman-environment`
10. `node bin/polymer.js init`
11. Observe that the normal generator options are shown in the console and no error

Fixes #364


justinfagnani:
The problem with `requireg` is that we can't request dependencies at a particular version. For web-component-tester, we need to upgrade it to versions with new features and breaking changes.

As a first crack at reducing install size I'd like to look at reducing sizes and consolidating similar dependencies among the packages we own. Removing selenium-standalone from WCT should have a big impact.

TimvdLippe:
@justinfagnani Understood. From a user perspective, if WCT can be refactored to separate the running logic from the actual project logic (e.g. variants configuration etc...) that would be really great.

I just thought that we can global require `bower` too, since probably a lot of developers have it installed already. If I revert the WCT changes for now and add `bower` in, would that be okay with you?

FredKSchott:
One method I've seen work is to attach a version property to the package. Then the CLI would check that version at runtime.

```js
const wct = require('web-component-tester');

if (!semver(wct.version, DESIRED_RANGE)) {
  throw new Error(`Version ${DESIRED_RANGE} Required!`);
}
```

TimvdLippe:
Closing for now, other proposals are being worked on.

PR #524, opened 2016-12-20
Cleanup how optimize streams work, in preparation for babel stream

FredKSchott:
This PR addresses two problems:
- All the build options had gotten pretty confusing, and assumed a world where each language only had one optimizer. This PR cleans up the CLI commands and simplified the optimize options that the main build function accepts.
- Constructing our optimize stream inside of pipes and gulp if's was too limiting. We now call out to `getOptimizeStreams` to calculate and return the optimize streams that we need.

This PR should be a noop functionality-wise, and prepares for a much easier diff in https://github.com/Polymer/polymer-cli/pull/525.

### DIFF
--- a/src/build/build.ts
+++ b/src/build/build.ts
@@ -26,7 +26,7 @@ import {OptimizeOptions, getOptimizeStreams} from './optimize-streams';
 
 import {ProjectConfig} from 'polymer-project-config';
 import {PrefetchTransform} from './prefetch';
-import {waitFor} from './streams';
+import {waitFor, pipeStreams} from './streams';
 import {parsePreCacheConfig} from './sw-precache';
 
 const logger = logging.getLogger('cli.build.build');
@@ -38,7 +38,8 @@ export interface BuildOptions {
   swPrecacheConfig?: string;
   insertDependencyLinks?: boolean;
   optimize?: OptimizeOptions;
-};
+}
+;
 
 export async function build(
     options: BuildOptions, config: ProjectConfig): Promise<void> {
@@ -55,24 +56,20 @@ export async function build(
   logger.info(`Building application...`);
 
   logger.debug(`Reading source files...`);
-  const sourcesStream = [].concat.apply([], [
+  const sourcesStream = pipeStreams([
     polymerProject.sources(),
     polymerProject.splitHtml(),
     getOptimizeStreams(options.optimize),
     polymerProject.rejoinHtml()
-  ]).reduce((a: NodeJS.ReadableStream, b: NodeJS.ReadWriteStream) => {
-    return a.pipe(b);
-  });
+  ]);
 
   logger.debug(`Reading dependencies...`);
-  const depsStream = [].concat.apply([], [
+  const depsStream = pipeStreams([
     polymerProject.dependencies(),
     polymerProject.splitHtml(),
     getOptimizeStreams(options.optimize),
     polymerProject.rejoinHtml()
-  ]).reduce((a: NodeJS.ReadableStream, b: NodeJS.ReadWriteStream) => {
-    return a.pipe(b);
-  });
+  ]);
 
   let buildStream = mergeStream(sourcesStream, depsStream);
 
--- a/src/build/build.ts
+++ b/src/build/build.ts
@@ -13,7 +13,6 @@
  */
 
 import * as del from 'del';
-import * as gulpif from 'gulp-if';
 import * as path from 'path';
 import * as logging from 'plylog';
 import {dest} from 'vinyl-fs';
--- a/src/build/build.ts
+++ b/src/build/build.ts
@@ -32,15 +32,16 @@ const logger = logging.getLogger('cli.build.build');
 
 const buildDirectory = 'build/';
 
-export interface BuildOptions {
+export interface BuildOptions extends OptimizeOptions {
   swPrecacheConfig?: string;
   insertPrefetchLinks?: boolean;
   bundle?: boolean;
-  optimize?: OptimizeOptions;
 };
 
 export async function build(
     options: BuildOptions, config: ProjectConfig): Promise<void> {
+  const optimizeOptions:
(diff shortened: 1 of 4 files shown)


src/build/optimize-streams.ts
```diff
@@ -30,6 +30,11 @@ export type FileCB = (error?: any, file?: File) => void;
 export type CSSOptimizeOptions = {
   stripWhitespace?: boolean;
 };
+export interface OptimizeOptions {
+  htmlMinify?: HTMLMinifierOptions;
```
justinfagnani: Why not have the nested structure here that we do in BuildOptions? Then in #525 instead of adding a `jsCompile` property, you add `compile` to the `js` property. Or to go further, just get rid of this interface and pass BuildOptions to `getOptimizeStreams`.

src/commands/build.ts
```diff
@@ -32,6 +32,21 @@ export class BuildCommand implements Command {
 
   args = [
     {
+      name: 'js-minify',
```
justinfagnani: Not so sure about this change. The reason for namespacing flags like `html.collapseWhitespace` was to at some point be able to easily map flags to build options. Continuing that pattern would lead to js.minify, css.minify and html.minify.

FredKSchott: I really dislike the `.` in the command line flag, a dash feels much more natural and doesn't break the expected CLI dash-case naming convention. We can still map easily regardless of `.` or `-`.

src/build/optimize-streams.ts
FredKSchott: I wanted to pass this subset of build options to getOptimizeStreams() to simplify the interface. But I can match the nested structure from BuildOptions if you prefer.

src/build/build.ts
```diff
@@ -65,28 +63,24 @@ export async function build(
...
-          // is properly being isolated by splitHtml() & rejoinHtml().
-          .pipe(gulpif(/\.html$/, new InlineCSSOptimizeStream(optimizeOptions.css)))
-          .pipe(gulpif(/\.html$/, new HTMLOptimizeStream(optimizeOptions.html)))
-          .pipe(polymerProject.rejoinHtml());
+  const sourcesStream = [].concat.apply([], [
+    polymerProject.sources(),
+    polymerProject.splitHtml(),
+    getOptimizeStreams(optimizeOptions),
+    polymerProject.rejoinHtml()
+  ]).reduce((a: NodeJS.ReadableStream, b: NodeJS.ReadWriteStream) => {
```
FredKSchott: good call

src/commands/build.ts
justinfagnani: I don't think it's that uncommon to use `.` when flags map to JSON-like structures. Mapping with `-` is harder: does `foo-bar-baz` map to `foo.bar.baz`, `foo.bar-baz`, etc.

src/build/optimize-streams.ts
FredKSchott: Nice, nested looks much cleaner (less mapping and plays much nicer with TS)

src/commands/build.ts
justinfagnani: The user probably will, when we explain that the flags map to the options in polymer.json

FredKSchott: Ah, I hadn't assumed we'd use the same options for `build()` in `polymer.json`. 

Can we table this decision then until we start on the polymer.json options? I think that will be a more productive conversation when those options actually exist vs. planning for something we haven't really thought through yet (at least I haven't yet).

justinfagnani: We can, but the pattern had been set up with that goal in mind. If we defer the discussion can we stick with the current pattern?

FredKSchott: sure thing, rolling back

FredKSchott:
@justinfagnani comments addressed, PTAL

src/commands/build.ts
```diff
@@ -44,11 +59,6 @@ export class BuildCommand implements Command {
           'additional `<link rel="prefetch">` tags into ' +
           'entrypoint and `<link rel="import">` tags into fragments and shell.'
     },
-    {
-      name: 'html.collapseWhitespace',
```
justinfagnani: can we deprecate for one release before removing?

justinfagnani: Actually, collapseWhitespace is only one of many options offered here: https://www.npmjs.com/package/html-minifier

Are we now mapping minify to a specific set of those?

FredKSchott: Yes, see: https://github.com/Polymer/polymer-cli/pull/524/files#diff-6efc8ba30392e1492e440aea247b32f6R155

With the way `command-line-args` works, anyone who is using this will get an error once this is removed. That feels like enough to me, wdyt? Are enough people actually using this?

FredKSchott: I'm not against opening explicit configuration options back up in `polymer.json`, but I have a hard time picturing anyone setting specific optimization JSON options objects via CLI flags every time they want to build. At that point that user really should be saving their build configuration in a polymer.json. I'd rather we encourage them to do that and keep these CLI flags as simple as possible.

Again, I don't think we should spend time at the moment worrying about how these CLI flags will match the polymer.json interface, when that interface doesn't even exist yet. I'd much rather discuss that when we do add the polymer.json interface, either in a design proposal or the PR itself. None of this is going out to users until then, so there is no difference between discussing in this PR vs. that PR, except that we'll have both interfaces concretely in front of us and will be able to reason about them much more easily.

FredKSchott:
@justinfagnani PTAL

src/commands/build.ts
FredKSchott: discussed in person, our flags will use this format (https://docs.google.com/document/d/1mD1cVhxdxY-bUsGBX44anC1875Afx7UHue4uRCNVc4o)

FredKSchott:
@justinfagnani code now matches our new design doc, PTAL

PR #525, opened 2016-12-20
add babel to build stream, replace uglify with babili es2015 minifier

FredKSchott:
Relies on https://github.com/Polymer/polymer-cli/pull/524

This started as a PR to just add babel compile flag. But because babili and babel are so connected, the work to replace uglify with babili got rolled into this PR as well.


src/build/optimize-streams.ts
```diff
@@ -158,6 +180,9 @@ export function getOptimizeStreams(optimizeOptions: OptimizeOptions) {
   if (optimizeOptions.jsMinify) {
     streams.push(gulpif(/\.js$/, new JSOptimizeStream({fromString: true})));
   }
+  if (optimizeOptions.jsCompile) {
+    streams.push(gulpif(/\.js$/, new JSBabelStream({presets: [babelPresetES2015NoModules]})));
```
yordis: @FredKSchott because people will want to have control over the preset, why we don't use `babel-env` preset instead of `es2015` preset? This make flexible the build and people don't need to create custom builds because they want to use any hot feature that is not on `es2015`?

src/build/optimize-streams.ts
```diff
@@ -92,6 +98,22 @@ export class JSOptimizeStream extends GenericOptimizeStream {
 }
 
 /**
+ * JSBabelStream transpiles Javascript down to work in older browsers, rewriting
+ * newer ECMAScript features to only use language features available in major
+ * browsers. If no options are given to the constructor, JSBabelStream will use
+ * a default "ES6 -> ES5" preset (configuration).
+ */
+export class JSBabelStream extends GenericOptimizeStream {
```
justinfagnani: maybe name this `JSCompileTransform`

src/build/optimize-streams.ts
```diff
@@ -34,6 +39,7 @@ export interface OptimizeOptions {
   htmlMinify?: HTMLMinifierOptions;
   cssMinify?: CSSOptimizeOptions;
   jsMinify?: UglifyOptions;
+  jsCompile?: BabelTransformOptions;
```
justinfagnani: After rebasing onto #524, this should be nested under the `js` property.

FredKSchott:
@justinfagnani uglify was added into this PR (since its built off of babel as well). PTAL

src/build/optimize-streams.ts
```diff
@@ -33,17 +39,18 @@ export type CSSOptimizeOptions = {
 export interface OptimizeOptions {
   html?: {minify: boolean};
   css?: {minify: boolean};
-  js?: {minify?: boolean};
-};
+  js?: {minify?: boolean, compile: boolean};
```
justinfagnani: `compile` could be optional too

src/build/optimize-streams.ts
```diff
@@ -75,26 +82,46 @@ export class GenericOptimizeStream extends Transform {
...
-export class JSOptimizeStream extends GenericOptimizeStream {
-  constructor(options: UglifyOptions) {
-    // uglify is special, in that it returns an object with a code property
-    // instead of just a code string. We create a compliant optimizer here
-    // that returns a string instead.
-    const uglifyOptimizer = (contents: string, options: UglifyOptions) => {
-      return uglify(contents, options).code;
+class JSBabelTransform extends GenericOptimizeTransform {
+  constructor(config: BabelTransformOptions) {
+    let transform = (contents: string, options: BabelTransformOptions) => {
```
justinfagnani: const

src/build/optimize-streams.ts
```diff
@@ -75,26 +82,46 @@ export class GenericOptimizeStream extends Transform {
...
+      return babelTransform(contents, options).code!;
```
justinfagnani: can you check if Babel produces warnings and a compile output?

justinfagnani: TODO for source map support?

src/build/optimize-streams.ts
FredKSchott: oh whoops, it should be. Good catch

src/build/optimize-streams.ts
FredKSchott: no, just errors. Discussed in person, but we may need to turn these into true stream errors if code doesn't compile. That comes with its own set of problems (should we block if dependencies don't compile?). 

Will monitor and address if necessary.

PR #528, opened 2016-12-22
stop stripping "polymer-starter-kit" from those templates

FredKSchott:
Follow up from https://github.com/Polymer/polymer-cli/pull/472. 

We are now resolving bundled generators properly, but @justinfagnani pointed out that the `polymer-starter-kit-` prefix that is stripped out is important info. This PR removes that stripping.

### Current Behavior
old behavior

### This PR
screen shot 2016-12-22 at 11 46 01 am

/cc @justinfagnani @tony19

### DIFF
--- a/src/init/init.ts
+++ b/src/init/init.ts
@@ -156,12 +156,11 @@ function getDisplayName(generatorName: string) {
   // -------------------------------------------------------------------
   // (generator-)?           | Grp 1; Match "generator-"; Optional
   // (polymer-init)?         | Grp 2; Match "polymer-init-"; Optional
-  // (polymer-starter-kit-)? | Grp 3; Match "polymer-starter-kit-"; Optional
-  // ([^:]+)                 | Grp 4; Match one or more characters != ":"
-  // (:.*)?                  | Grp 5; Match ":" followed by anything; Optional
+  // ([^:]+)                 | Grp 3; Match one or more characters != ":"
+  // (:.*)?                  | Grp 4; Match ":" followed by anything; Optional
   return generatorName.replace(
-      /(generator-)?(polymer-init-)?(polymer-starter-kit-)?([^:]+)(:.*)?/g,
-      '$4');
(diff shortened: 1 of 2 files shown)


src/init/init.ts
```diff
@@ -156,12 +156,11 @@ function getDisplayName(generatorName: string) {
   // -------------------------------------------------------------------
   // (generator-)?           | Grp 1; Match "generator-"; Optional
   // (polymer-init)?         | Grp 2; Match "polymer-init-"; Optional
-  // (polymer-starter-kit-)? | Grp 3; Match "polymer-starter-kit-"; Optional
```
tony19: Other cleanup changes that should be done:
 1. Remove `"polymer-starter-kit-"` from function comment at line [140](https://github.com/Polymer/polymer-cli/pull/528/files#diff-6eb970fedd99f33facb11eefde800f22L140).
 2. Update `group 4 in pattern` to `group 3 in pattern` in comment at line [153](https://github.com/Polymer/polymer-cli/pull/528/files#diff-6eb970fedd99f33facb11eefde800f22L153).

tony19: I think it would be easier to replace this commit with `git revert e063a60` to undo the [commit that added the `polymer-starter-kit-` prefix](https://github.com/Polymer/polymer-cli/pull/472/commits/e063a604dc9306eff0c6fdca354bfa4fd5e95258).

FredKSchott: I would but the commits were squashed so they don't exist in the branch or anywhere I can find them in our repo. I've mimic'd reverting the commit though, so this should be fine.

FredKSchott:
@justinfagnani PTAL

justinfagnani:
LGTM after we merge https://github.com/Polymer/polymer-cli/pull/533, rebase and re-run tests

justinfagnani:
#533 is in, please rebase to kick off tests

PR #534, opened 2017-01-04
add service worker flag to build command

FredKSchott:
(CHANGELOG will be updated when `new-build-flags` is merged into master).

This PR adds a flag to the build command to generate a service worker. Like all the other new build flags this one is "default off" for now while we work towards the greater `polymer.json` build flags / build permutations story.

/cc @justinfagnani


src/commands/build.ts
```diff
@@ -39,10 +39,19 @@ export class BuildCommand implements Command {
           'requests needed to serve your application.'
     },
     {
-      name: 'sw-precache-config',
```
justinfagnani: why rename this flag? The file is still sw-precache specific?

src/commands/build.ts
```diff
@@ -39,10 +39,19 @@ export class BuildCommand implements Command {
...
-      defaultValue: 'sw-precache-config.js',
-      description: 'Path to an sw-precache configuration to be ' +
-          'used for service worker generation.'
+      name: 'add-service-worker',
```
justinfagnani: until we support more than sw-precache, we may want to name this something sw-precache specific. maybe `--add-sw-precache` or `--use-sw-precache`

FredKSchott:
@justinfagnani the thinking was to not tie the service worker generation to the `sw-precache` implementation directly. Instead the user just knows that they're getting a service worker, and doesn't need to know what sw-precache is or how it works until they want to configure it. This felt like a significant enough usability improvement to warrant the change.

For the config filepath flag I could go either way. But given the `--add-service-worker` flag, keeping the `service-worker` name in the config flag as well felt like a better fit.

justinfagnani:
I don't think we can uncouple the config from sw-precache because they have to know how to configure sw-precache from the config. The new flags docs even reference sw-precache docs.

FredKSchott:
@justinfagnani renamed based on https://docs.google.com/document/d/1mD1cVhxdxY-bUsGBX44anC1875Afx7UHue4uRCNVc4o, PTAL

src/build/build.ts
```diff
@@ -102,29 +101,39 @@ export async function build(
...
   // addServiceWorker() reads from the file system, so we need to wait for
   // the build stream to finish writing to disk before calling it.
-  await addServiceWorker({
-    buildRoot: buildDirectory,
-    project: polymerProject,
-    swPrecacheConfig: swConfig || undefined,
-    bundled: options.bundle,
-  });
+  if (options.addServiceWorker) {
+    logger.debug(`Generating service worker...`);
```
justinfagnani: I could see this as a valuable `logger.info`

Issue #186, opened 2016-05-21
polymer init application missing name?


Alos:
Just installed the polymer-cli on OS X. After that I tried creating an application using:

 `polymer init application`

I get the following:

```
info:    Running template application
error { [AssertionError: A name parameter is required to create a storage]
  name: 'AssertionError',
  actual: '',
  expected: true,
  operator: '==',
  message: 'A name parameter is required to create a storage',
  generatedMessage: false }
AssertionError: A name parameter is required to create a storage
    at new Storage (/usr/local/lib/node_modules/polymer-cli/node_modules/yeoman-generator/lib/util/storage.js:24:3)
    at ApplicationGenerator.Base._getStorage (/usr/local/lib/node_modules/polymer-cli/node_modules/yeoman-generator/lib/base.js:665:10)
    at ApplicationGenerator.Base (/usr/local/lib/node_modules/polymer-cli/node_modules/yeoman-generator/lib/base.js:157:22)
    at ApplicationGenerator (/usr/local/lib/node_modules/polymer-cli/lib/init/application/application.js:15:14)
    at Environment.instantiate (/usr/local/lib/node_modules/polymer-cli/node_modules/yeoman-environment/lib/environment.js:328:10)
    at Environment.create (/usr/local/lib/node_modules/polymer-cli/node_modules/yeoman-environment/lib/environment.js:305:15)
    at Environment.run (/usr/local/lib/node_modules/polymer-cli/node_modules/yeoman-environment/lib/environment.js:364:24)
    at runGenerator (/usr/local/lib/node_modules/polymer-cli/lib/commands/init.js:65:29)
    at /usr/local/lib/node_modules/polymer-cli/lib/commands/init.js:80:21
    at Environment.resolver.lookup (/usr/local/lib/node_modules/polymer-cli/node_modules/yeoman-environment/lib/resolver.js:50:12)
```

I was expecting polymer-cli to ask for the project name or use the directory name as the default. While installing polymer-cli there where some warnings not sure they are related or not:

```
$ sudo npm install -g polymer-cli
Password:
npm WARN deprecated cross-spawn-async@2.2.4: cross-spawn no longer requires a build toolchain, use it instead!
npm WARN deprecated graceful-fs@3.0.8: graceful-fs v3.0.0 and before will fail on node releases >= v7.0. Please update to graceful-fs@^4.0.0 as soon as possible. Use 'npm ls graceful-fs' to find it in the tree.
[...]

PR #536, opened 2017-01-05
Fix https://github.com/Polymer/polymer-cli/issues/186

justinfagnani:

### DIFF
--- a/src/init/application/application.ts
+++ b/src/init/application/application.ts
@@ -23,6 +23,14 @@ export class ApplicationGenerator extends Generator {
     this.sourceRoot(path.join(__dirname, 'templates'));
   }
 
+  // This is necessary to prevent an exception in Yeoman when creating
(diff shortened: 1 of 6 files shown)


src/init/application/application.ts
```diff
@@ -23,6 +23,14 @@ export class ApplicationGenerator extends Base {
     this.sourceRoot(path.join(__dirname, 'templates'));
   }
 
+  // This is necessary to prevent an exception in Yeoman when creating
+  // storage for generators registered as a stub and used in a folder
+  // with a package.json but with no name property.
+  // https://github.com/Polymer/polymer-cli/issues/186
```
FredKSchott: IMO these warnings aren't needed

justinfagnani: Generators wouldn't normally override this method, I think it help to know why.

PR #545, opened 2017-01-19
Add support for polymer.json build configurations

FredKSchott:
This PR adds support for build configurations stored in a user's `polymer.json`. See [Build Permutations Design Doc](https://docs.google.com/document/d/1mD1cVhxdxY-bUsGBX44anC1875Afx7UHue4uRCNVc4o) for design overview.

*Note: Because of https://github.com/Polymer/polymer-build/issues/113, we need to create new instances of polymer-build projects for each build. This adds unnecessary reads and file analysis for each additional build. Once this is no longer required, we'll rewrite this logic to only use a single polymer-project instance for all builds.*

***Note: Requires release of https://github.com/Polymer/polymer-project-config/pull/5. Tests will pass once that is released.***

### DIFF
--- a/src/build/build.ts
+++ b/src/build/build.ts
@@ -12,7 +12,6 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import * as del from 'del';
 import * as path from 'path';
 import * as logging from 'plylog';
 import {dest} from 'vinyl-fs';
@@ -20,32 +19,24 @@ import mergeStream = require('merge-stream');
 import {PolymerProject, addServiceWorker, SWConfig} from 'polymer-build';
 
 import {OptimizeOptions, getOptimizeStreams} from './optimize-streams';
-import {ProjectConfig} from 'polymer-project-config';
+import {ProjectConfig, ProjectBuildOptions} from 'polymer-project-config';
 import {PrefetchTransform} from './prefetch';
 import {waitFor, pipeStreams} from './streams';
 import {loadServiceWorkerConfig} from './load-config';
 
 const logger = logging.getLogger('cli.build.build');
 
-const buildDirectory = 'build/';
-
-export interface BuildOptions extends OptimizeOptions {
-  addServiceWorker?: boolean;
-  swPrecacheConfig?: string;
-  insertPrefetchLinks?: boolean;
-  bundle?: boolean;
-};
-
 export async function build(
-    options: BuildOptions, config: ProjectConfig): Promise<void> {
+    options: ProjectBuildOptions, config: ProjectConfig): Promise<void> {
   const optimizeOptions:
       OptimizeOptions = {css: options.css, js: options.js, html: options.html};
   const polymerProject = new PolymerProject(config);
 
-  logger.info(`Deleting build/ directory...`);
-  await del([buildDirectory]);
+  // If no name is provided, write directly to the build/ directory.
+  // If a build name is provided, write to that subdirectory.
+  const buildDirectory = `build/${options.name || ''}`;
+  logger.debug(`Building ${buildDirectory} with options`, options);
 
-  logger.debug(`Reading source files...`);
   const sourcesStream = pipeStreams([
     polymerProject.sources(),
     polymerProject.splitHtml(),
@@ -53,7 +44,6 @@ export async function build(
     polymerProject.rejoinHtml()
   ]);
 
-  logger.debug(`Reading dependencies...`);
   const depsStream = pipeStreams([
     polymerProject.dependencies(),
     polymerProject.splitHtml(),
@@ -64,10 +54,6 @@ export async function build(
   let buildStream: NodeJS.ReadableStream =
       mergeStream(sourcesStream, depsStream);
 
-  buildStream.once('data', () => {
-    logger.debug('Analyzing build dependencies...');
-  });
-
   if (options.bundle) {
     buildStream = buildStream.pipe(polymerProject.bundler);
   }
@@ -77,7 +63,7 @@ export async function build(
   }
 
   buildStream.once('data', () => {
-    logger.info('Generating build/ directory...');
+    logger.info(`Generating ${buildDirectory}...`);
   });
 
   // Finish the build stream by piping it into the final build directory.
@@ -99,11 +85,12 @@ export async function build(
   // addServiceWorker() reads from the file system, so we need to wait for
   // the build stream to finish writing to disk before calling it.
   if (options.addServiceWorker) {
-    logger.info(`Generating service worker...`);
+    logger.debug(`Generating service worker...`);
     if (swConfig) {
       logger.debug(`Service worker config found`, swConfig);
     } else {
-      logger.debug(`No service worker configuration found at ` +
+      logger.debug(
+          `No service worker configuration found at ` +
           `${swPrecacheConfigPath}, continuing with defaults`);
     }
     await addServiceWorker({
@@ -113,6 +100,9 @@ export async function build(
       bundled: options.bundle,
     });
   }
-
-  logger.info('Build complete!');
+  if (options.name) {
+    logger.info(`Build "${options.name}" complete!`);
+  } else {
+    logger.info(`Build complete!`);
+  }
 }
--- a/src/build/build.ts
+++ b/src/build/build.ts
@@ -27,6 +27,15 @@ import {loadServiceWorkerConfig} from './load-config';
 const logger = logging.getLogger('cli.build.build');
 export const mainBuildDirectoryName = 'build';
 
+/**
+ * Generate a single build based on the given `options` ProjectBuildOptions.
+ * Note that this function is only concerned with that single build, and does
+ * not care about the collection of builds defined on the config.
+ *
+ * TODO(fks) 01-26-2017: Generate multiple builds with a single PolymerProject
+ * instance. Currently blocked because splitHtml() & rejoinHtml() cannot be run
+ * on multiple streams in parallel. See: https://github.com/Polymer/polymer-build/issues/113
+ */
 export async function build(
     options: ProjectBuildOptions, config: ProjectConfig): Promise<void> {
   const optimizeOptions:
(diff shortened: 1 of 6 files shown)


src/commands/build.ts
```diff
@@ -85,36 +83,58 @@ export class BuildCommand implements Command {
     },
   ];
 
-  run(options: CommandOptions, config: ProjectConfig): Promise<any> {
+  async run(options: CommandOptions, config: ProjectConfig): Promise<any> {
     // Defer dependency loading until this specific command is run
-    const build = require('../build/build').build;
+    let build = require('../build/build').build;
+    const hasCLIArgumentsPassed = this.args.some(arg => {
+      return typeof options[arg.name] !== 'undefined';
```
justinfagnani: I don't think you need `typeof`, just: `this.args.some((arg) => options[arg.name] !== undefined)`

src/commands/build.ts
```diff
@@ -85,36 +83,58 @@ export class BuildCommand implements Command {
...
+    logger.info(`Clearing ${mainBuildDirectory} directory...`);
+    await del([mainBuildDirectory]);
 
-    if (options['env'] && options['env'].build) {
-      logger.debug('env.build() found in options');
-      logger.debug('building via env.build()...');
-      return options['env'].build(buildOptions, config);
+    // If any the build command flags were passed as CLI arguments, generate
+    // a single build based on those flags alone.
+    if (hasCLIArgumentsPassed) {
```
justinfagnani: I wonder if this deserves some console output. What happens if I have build defined, but then to `--html-minify`, do I expect to get both config and flag-based builds? Also, this build doesn't have a name, there could be conflicts between folders in the output and other build names, or it could conflict with an unnamed build in the config. Need to validate that.

src/commands/build.ts
```diff
@@ -85,36 +83,58 @@ export class BuildCommand implements Command {
...
+      return build(
+          {
+            addServiceWorker: options['add-service-worker'],
+            swPrecacheConfig: options['sw-precache-config'],
+            insertPrefetchLinks: options['insert-prefetch-links'],
+            bundle: options['bundle'],
```
justinfagnani: `!!` here as well?

src/build/build.ts
```diff
@@ -12,48 +12,38 @@
...
+    options: ProjectBuildOptions, config: ProjectConfig): Promise<void> {
   const optimizeOptions:
       OptimizeOptions = {css: options.css, js: options.js, html: options.html};
   const polymerProject = new PolymerProject(config);
 
-  logger.info(`Deleting build/ directory...`);
-  await del([buildDirectory]);
+  // If no name is provided, write directly to the build/ directory.
+  // If a build name is provided, write to that subdirectory.
+  const buildDirectory = `build/${options.name || ''}`;
```
justinfagnani: Use the `path` library to join the build name into a path

src/commands/build.ts
```diff
@@ -85,36 +83,58 @@ export class BuildCommand implements Command {
...
+          },
+          config);
+    }
+
+    // If any builds were defined or configured via the project config,
+    // generate a build for each configuration.
+    if (config.builds) {
+      return Promise.all(
+          config.builds.map((buildOptions: ProjectBuildOptions, i: number) => {
+            buildOptions.name = buildOptions.name || `build-${i + 1}`;
```
justinfagnani: hmm, this kind of takes care of some of my above comments, but I think it'd still be good to validate first, especially for duplicate build names

FredKSchott: I like validating against duplicate build names. I also like making build names optional with this behavior.

src/commands/build.ts
FredKSchott: See [Build Permutations Design Doc](https://docs.google.com/document/d/1mD1cVhxdxY-bUsGBX44anC1875Afx7UHue4uRCNVc4o) for expected behavior

src/commands/build.ts
FredKSchott: I prefer `typeof !== 'undefined'` vs `!== undefined`, you never know if any of your dependencies are messing with `undefined` so why risk it.

src/commands/build.ts
```diff
@@ -12,13 +12,12 @@
...
 
-import * as logging from 'plylog';
-import {ProjectConfig} from 'polymer-project-config';
-
 // Only import type definitions here, otherwise this line will be included in
 // the JS output, triggering  the entire build library & its dependencies to
 // be loaded and parsed.
-import {BuildOptions} from '../build/build';
+import * as del from 'del';
+import * as logging from 'plylog';
```
FredKSchott: Outdated comment, will remove

src/build/build.ts
```diff
@@ -12,48 +12,38 @@
...
-export interface BuildOptions extends OptimizeOptions {
-  addServiceWorker?: boolean;
-  swPrecacheConfig?: string;
-  insertPrefetchLinks?: boolean;
-  bundle?: boolean;
-};
-
 export async function build(
-    options: BuildOptions, config: ProjectConfig): Promise<void> {
+    options: ProjectBuildOptions, config: ProjectConfig): Promise<void> {
```
FredKSchott: Not sure what you mean. `build()` uses the build `options` passed to it for the single build it generates. It doesn't care about `config.builds`.

src/build/build.ts
FredKSchott: nice, good catch

src/commands/build.ts
FredKSchott: Fixed in polymer-project-config https://github.com/Polymer/polymer-project-config/pull/5

FredKSchott:
@justinfagnani PTAL

src/build/build.ts
justinfagnani: That's my question: why doesn't it care about `config.builds`? Coming into this code, if I know that ProjectConfig has a builds property, I wonder why this doesn't use it. I think you're saying this only performs one of the builds contained in there, which would answer my question, but I think this could use some comments to that effect if so.

src/commands/build.ts
justinfagnani: Because if someone has redefined `undefined` then all hell will break loose, and we will find and ~~prosecute the offenders to the fullest extent of the law~~ remove that dependency. We can't possibly avoid the `undefined` reference in entire whole app plus dependencies, so I'd rather do the simple, sanity-assuming thing. You could also check against `null` or use the `in` operator, both of which in this case will be effectively the same.

src/commands/build.ts
justinfagnani: Ok, to iterate on that, I don't think it's going to be very ergonomic to use an index-based build name. I'd rather just require names, with the possible exception of only supporting an unnamed build if it's the only build.

src/commands/build.ts
```diff
@@ -85,30 +84,35 @@ export class BuildCommand implements Command {
 
   async run(options: CommandOptions, config: ProjectConfig): Promise<any> {
     // Defer dependency loading until this specific command is run
-    let build = require('../build/build').build;
-    const hasCLIArgumentsPassed = this.args.some(arg => {
-      return typeof options[arg.name] !== 'undefined';
-    });
+    let {build} = require('../build/build');
```
justinfagnani: I think this destructuring might fail in node 4

src/commands/build.ts
```diff
@@ -85,30 +84,35 @@ export class BuildCommand implements Command {
...
+    const {mainBuildDirectoryName} = require('../build/build');
```
justinfagnani: rather than `require()` twice, I think just do:

```
const buildLib = require(...);
let build = buildLib.build;
const mainBuildDirectoryName = buildLib.mainBuildDirectoryName;
```

If we could use destructuring, you could get some benefit:

```
const buildLib = require(...);
let {build} = buildLib;
const {mainBuildDirectoryName} = buildLib;
```

src/commands/build.ts
```diff
@@ -85,30 +84,35 @@ export class BuildCommand implements Command {
...
-    logger.info(`Clearing ${mainBuildDirectory} directory...`);
-    await del([mainBuildDirectory]);
+    logger.info(`Clearing ${mainBuildDirectoryName}${pathSeperator} directory...`);
+    await del([mainBuildDirectoryName]);
 
     // If any the build command flags were passed as CLI arguments, generate
     // a single build based on those flags alone.
-    if (hasCLIArgumentsPassed) {
+    const hasCliArgumentsPassed = this.args.some((arg) => {
+      return typeof options[arg.name] !== 'undefined';
```
justinfagnani: you still don't need the braces and return keyword

src/commands/build.ts
FredKSchott: bah you're right, the fact that typescript transpiles import restructuring always trips me up

src/commands/build.ts
FredKSchott: I agree this falls closer to paranoia than reasonably common. But to me its the exact same argument as using semi-colons: "it's almost always fine, but why add risk unnecessarily?"

Happy to bring this up in a weekly standup if this is something you'd like us to come together on as a team.

src/build/build.ts
FredKSchott: Agreed, will add comments. 

Remember that the goal is for `build()` to handle all builds defined in the `config`, and not take a single `BuildOptions` at all. (builds defined in CLI flags may complicate this a bit). That way we'd only need a single `PolymerProject` instance. But this is currently blocked by the split/rejoin bug. 

Will summarize all this in the added comments.

FredKSchott:
@justinfagnani PTAL

src/commands/build.ts
justinfagnani: import destructuring isn't really, it just looks like it 🙄

FredKSchott:
@justinfagnani definitely. Thanks for the reviews 👍 

I'm going to wait until I've gotten out v1.2.0 of polymer-project-config so that we can get a passing Travis CI before merging.

PR #552, opened 2017-02-03
delete unused reyserve environment

FredKSchott:
We've never touched this since it was created. Instead of refactoring it for the github semver matching updates, I just decided to remove it (with the option to always add it back later).

/cc @justinfagnani to confirm.


FredKSchott:
Whoops, forgot that #553 would get merged into this. See https://github.com/Polymer/polymer-cli/pull/552/commits/ce41a80cf110edb925d8ad38ffee5ab9450f2350 for commit that removes

PR #553, opened 2017-02-03
add init template semver matching for github templates

FredKSchott:

This adds semver matching to the github Yeoman generator, so that we can get the latest 1.0 & 2.0 releases from the same repo. It also simplifies the interface by refactoring a lot of the tests around the generator.


test/unit/init/github_test.js
```diff
@@ -10,95 +10,86 @@
 'use strict';
```
FredKSchott: This file is more or less completely redone, but I can't find out how to tell Github to stop diff failing. You'll be better served clicking "View" above to review.

src/init/init.ts
```diff
@@ -169,10 +157,29 @@ function getDisplayName(generatorName: string) {
...
+    // TODO: Add Shop ^2.0.0 generator once Shop for Polymer 2.0 template is released
+    env.registerStub(
+        createGithubGenerator({
+          owner: 'Polymer',
+          repo: 'shop',
+          semverRange: '^1.0.0',
+        }),
+        'polymer-init-shop:app');
+    // Polymer Starter Kit Template
+    // TODO: Add Shop ^3.0.0 generator once PSK for Polymer 2.0 template is released
```
FredKSchott: @robdodson Just want to confirm that this in line with what you were thinking for PSK release versioning:

- The current PSK v2.x branch keep Polymer 1.0 with Hybrid elements
- PSK v3.0 and onward get Polymer 2.0

Our plan is to keep this in for the first few versions after Polymer 2.0 release, and then phase it out once appropriate.

FredKSchott:
I reverted the changes to `init/init.ts` that call this new interface and rolled them into the next PR (for the init+application new templates) that is touching the file more seriously, so that we don't have to review it twice.

src/init/init.ts
```diff
@@ -43,14 +43,17 @@ const templateDescriptions: {[name: string]: string} = {
...
 };
 
+// TODO: Add Shop "^2.0.0" generator once Polymer 2.0 Shop template is released.
 let shopGenerator = createGithubGenerator({
   owner: 'Polymer',
   repo: 'shop',
+  semverRange: '^1.0.0',
 });
-
+// TODO: Add Shop "^3.0.0" generator once Polymer 2.0 PSK template is released.
```
FredKSchott: @robdodson Just want to confirm that this in line with what you were thinking for PSK release versioning:

- The current PSK v2.x branch keep Polymer 1.0 with Hybrid elements
- PSK v3.0 and onward get Polymer 2.0

Our plan is to keep this in for the first few versions after Polymer 2.0 release, and then phase it out once appropriate.

robdodson: That sounds good to me

src/github/github.ts
```diff
@@ -99,51 +105,59 @@ export class Github {
...
-                      response.statusCode, response.statusMessage);
+                  reject(new GithubResponseError(
+                      response.statusCode, response.statusMessage));
+                  return;
                 }
                 logger.info('Unpacking template files');
               })
+          .on('error', reject)
           .pipe(gunzip())
+          .on('error', reject)
```
justinfagnani: dammit 😠

justinfagnani: https://streams.spec.whatwg.org/#rs-pipe-to and rejoice

src/init/init.ts
```diff
@@ -43,14 +43,17 @@ const templateDescriptions: {[name: string]: string} = {
   'shop': 'The "Shop" Progressive Web App demo',
 };
 
+// TODO: Add Shop "^2.0.0" generator once Polymer 2.0 Shop template is released.
 let shopGenerator = createGithubGenerator({
   owner: 'Polymer',
   repo: 'shop',
+  semverRange: '^1.0.0',
```
justinfagnani: Shouldn't this be `^2.0.0`?

src/init/init.ts
justinfagnani: So with this we have to update the code if they ship a new major version. Ok for now, but we might want to make templates self-serving, where we show the last 2 majors or something.

src/github/github.ts
FredKSchott: 🙏

src/init/init.ts
FredKSchott: Yea, happy to talk more about this when the time comes. I don't know if we could ever do a truely automated solution, since major version bumps don't necessarily map 1:1 with Polymer version bumps. So if we showed the last two majors, Polymer 1.0 would immediately get bumped off the list.

I hope that we can make the 1.0 -> 2.0 transition time (especially for someone creating a new project) as short as possible. Once feasible, not encouraging people to create 1.0 projects with `init` would support that goal.

src/init/init.ts
FredKSchott: see comment above. Is shop `#2.0-preview` ready to be presented to users?

FredKSchott:
Looks like depcheck had a strange flakey failure on 1 travis build. Lets keep an eye on that.

PR #554, opened 2017-02-03
add new polymer 2.0 element and application templates

FredKSchott:

The diff is bigger than it looks, there are a lot of files moved around and created to add the two new Yeoman templates. You can find most changes in `src/init/init.ts`.


src/init/application/application.ts
```diff
@@ -15,90 +15,93 @@ import * as chalk from 'chalk';
 import * as path from 'path';
 import Generator = require('yeoman-generator');
 
-export class ApplicationGenerator extends Generator {
-  props: any;
+export function createApplicationGenerator(templateName: string):
```
justinfagnani: doc that the template name is both the display name and where the template files are loaded from

src/init/element/element.ts
```diff
@@ -15,82 +15,86 @@ import * as chalk from 'chalk';
 import * as path from 'path';
 import Generator = require('yeoman-generator');
 
-export class ElementGenerator extends Generator {
-  props: any;
+export function createElementGenerator(templateName: string):
```
justinfagnani: also doc this

FredKSchott:
Comments addressed, letting CI run and then will merge after

PR #557, opened 2017-02-08
Update outdated dependencies

FredKSchott:

And with the next wct release, we'll be down to a single deprecation warning on install! (`use uuid module instead`, much better than the `TERRIBLE DDOS ERROR EVEN THOUGH THIS IS A CLI WHY RISK IT???` warning  our users are currently seeing).


FredKSchott:
This guy: https://github.com/Polymer/wct-local/issues/54, currently blocked by a transitive dependency, unfortunately.

PR #560, opened 2017-02-11
Fix an issue when compiling with the latest TSC.

rictic:
The latest version of typescript's standard library requires that the keys of a WeakMap be objects. So

interface WeakMap<K,V> { ... }

became

interface WeakMap<K extends object, V> { ... }

This conflicts with a declaration of WeakMap in the lodash typings (which the CLI depends on transitively).

Temporary workaround is to not typecheck libs.

More info: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/14324

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,7 +1,9 @@
 {
   "compilerOptions": {
     "target": "es6",
-    "lib": ["es2016"],
+    "lib": [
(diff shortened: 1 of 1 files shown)


justinfagnani:
Skipping lib checks seems fine in general.

PR #561, opened 2017-02-11
Use standard tslint

rictic:


package.json
```diff
@@ -14,7 +14,7 @@
     "build": "gulp build",
     "test": "gulp lint test",
     "prepublish": "npm run build && npm test",
-    "format": "find src test | grep '\\.js$\\|\\.ts$' | xargs clang-format --style=file -i"
+    "format": "find src test | grep '\\.js$\\|\\.ts$' | xargs clang-format --style=file -i && tslint --fix -c ./tslint.json src/*.ts src/**/*.ts src/**/**/*.ts"
```
justinfagnani: why `src/**/**/*.ts`?

rictic: Because without it files were left out on my machine. Looking more closely, I think it's because macOS ships an old bash.

`**` only works in bash >v4: http://unix.stackexchange.com/questions/49913/recursive-glob

justinfagnani: Ugh. This will only work for two levels of nesting now. Freakin' Apple

rictic: True, but it does work for this repo.

tslint.json
```diff
@@ -1,58 +1,5 @@
...
-            true,
-            "check-branch",
-            "check-decl",
-            "check-operator",
-            "check-separator",
-            "check-type"
-        ]
-    }
+    "extends": [
+        "@polymer/tools-common/tslint.json"
```
FredKSchott: what happens if this is changed remotely? Do our builds start breaking?

rictic: We control it, we version it, it's a normal dependency

rictic: And yes, cli's build will break if this changes to become more strict

FredKSchott: oh good, didn't see it added to the package.json above

package.json
```diff
@@ -82,6 +82,7 @@
     "yeoman-generator": "^1.0.1"
   },
   "devDependencies": {
+    "@polymer/tools-common": "^1.0.1",
```
FredKSchott: Follow up to tsconfig.extends chain below: 

I still don't feel 100% comfortable with the `^` meaning that our build config can be changed remotely, especially if the tools-common dependency is used/versioned for more than just tsconfig. How do you feel about pinning to explicit version here?

rictic: IMO we should just be 100% pure semver here. As it's a team internal package there is not even an optics concern to rolling out however many breaking changes we need

src/commands/help.ts
```diff
@@ -30,15 +30,14 @@ const CLI_DESCRIPTION = 'The multi-tool for Polymer projects';
...
-const HELP_HEADER = '\n' +
-  b('   /\\˜˜/   ') + m('/\\˜˜/') + b('\\   ') + '\n' +
-  b('  /__\\/   ') + m('/__\\/') + b('__\\  ') + '  ' + CLI_TITLE + '\n' +
-  b(' /\\  /   ') + m('/\\  /') + b('\\  /\\ ') + '\n' +
-  b('/__\\/   ') + m('/__\\/  ') + b('\\/__\\') + '  ' + CLI_DESCRIPTION + '\n' +
-  b('\\  /\\  ') + m('/\\  /   ') + b('/\\  /') + '\n' +
-  b(' \\/__\\') + m('/__\\/   ') + b('/__\\/ ') + '  ' + CLI_USAGE + '\n' +
-  b('  \\  ') + m('/\\  /   ') + b('/\\  /  ') + '\n' +
-  b('   \\') + m('/__\\/   ') + b('/__\\/   ') + '\n';
+const HELP_HEADER = '\n' + b('   /\\˜˜/   ') + m('/\\˜˜/') + b('\\   ') + '\n' +
```
justinfagnani: Please exclude this block from formatting. It's custom formatted to make it readable.

package.json
FredKSchott: hmm okay, the stakes are low enough that I'm willing to try. But my concern is still that we're giving ourselves a footgun to make small changes to the tslint file and version them as patches/minors, not realizing that they could remotely break other repos. If a human is still needed to manually upgrade major versions, then there's no extra work added by the more stable `=` vs. the more liberal `^`.

src/commands/help.ts
rictic: Fixed by rebasing against master

Issue #562, opened 2017-02-11
Integration tests failing


rictic:
It looks like https://github.com/Polymer/polymer-cli/pull/554 broke the integration tests (e.g. ApplicationGenerator was renamed to createApplicationGenerator). It looks like the integration tests aren't run on travis.

More broadly, while integration tests can be a pain, they're the most valuable tests to have, as the most brittle parts of a system are at the integration points, and there's no substitute for testing a system end to end. I think we'd end up saving time and effort if as many integration tests as possible ran with `npm test`, and I think we should definitely have all tests running on travis.

FredKSchott:
@rictic 100% agreed, our current plan is to work on these in the next few weeks as we get tooling ready for Polymer 2.0. I have a branch to get them running on our sample projects as well, which would be more complete and flexible if we can get it merged.

FWIW these were never working completely, even when they were added, so we were never able to integrate them into our CI process. We've wanted to for a while but just haven't had the resource to prioritize.

PR #563, opened 2017-02-11
Run integration tests on travis

rictic:
Makes progress on #562


rictic:
Hm, looks like there's an issue with bower on travis? The integration tests do some funny stuff with caching it seems. Needs investigation.

FredKSchott:
+1 if you can get tests passing!

We would get more converage if we commented out the specific steps in each template that failed, instead of skipping the entire test project. Even just making sure `init` is working for all of them is a step up from where we are now.

I'm working on a script to run commands against our sample projects as well, which we can combine with the existing tests once they're all passing.

rictic:
Ok, I've got the tests passing. I needed to delete the bower cache for the variants installation (it worked locally but failed on travis… kinda super weird), as well as disabling running `polymer lint` and `polymer test`. Linting was failing locally, while running wct was working locally for about half the examples.

I also set up testing for both 1.x and 2.x templates.

test/integration/install-variants_test.js
```diff
@@ -29,7 +31,7 @@ suite('install-variants', function() {
           assert.fail(err);
           done(err);
         }
-        runCommand(binPath, ['install', '--variants', '--offline'], {cwd: tmpPath}).then(() => {
+        runCommand(binPath, ['install', '--variants'], {cwd: tmpPath}).then(() => {
```
FredKSchott: we don't want to install offline for unit tests?

test/integration/run-command.js
```diff
@@ -30,11 +33,37 @@ function runCommand(path, args, options) {
...
+  }).catch((err) => {
+    return contents.then((out) => {
+      console.log(`Output of failed command 'node ${path} ${args.join(' ')}' in directory ${options.cwd}`);
+      console.log(out);
+      throw err;
+    });
+  });
+}
+
+function pipesToString(stdout, stderr) {
```
FredKSchott: add a comment describing what this does

test/integration/run-command.js
```diff
@@ -30,11 +33,37 @@ function runCommand(path, args, options) {
...
       if (code !== 0) {
-        reject(new Error('exit code: ' + code));
+        reject(new Error(
+          `Error running ${path} with args ${args}. Got exit code: ${code}`));
         return;
       }
       resolve();
     });
+  }).catch((err) => {
+    return contents.then((out) => {
```
FredKSchott: it took a while to figure out what you're trying to do here, a brief comment help a lot

rictic: :+1:, done!

test/integration/run-command.js
rictic: :+1:, done!

test/integration/install-variants_test.js
rictic: We do, but I struggled with making this work on travis. It's very strange that it works locally but not on travis, when all of the data should be local.

IMO hermetic tests are better than nonhermetic tests, but continuous integration is better than hermetic tests if you can't have both.

justinfagnani: What was the symptom on Travis? This is indeed odd as the code path I setup was to try to avoid any environment specifics interfering with the tests.

package.json
```diff
@@ -12,7 +12,8 @@
   "scripts": {
     "lint": "gulp lint",
     "build": "gulp build",
-    "test": "gulp lint test",
+    "test": "gulp lint test && gulp test:integration",
+    "test:smoke": "gulp lint test",
```
justinfagnani: I'd remove lint from smoke tests

test/integration/install-variants_test.js
rictic: Check it: https://travis-ci.org/Polymer/polymer-cli/jobs/200566297#L989

package.json
rictic: Done

test/integration/install-variants_test.js
justinfagnani: My bet is that Travis is setting an environment variable for the Bower cache, which is overriding our setting.

justinfagnani: See https://bower.io/docs/config/ env variables take precedence over local .bowerrc

rictic: It's not env variables apparently, no bower env variables found: https://travis-ci.org/Polymer/polymer-cli/jobs/202150103#L2339

rictic: Well, only except that explicitly building up an env that explicitly set env variables that start with `bower_` to the empty string seems to make offline installing work, even though I don't see any such env variables being printed: https://travis-ci.org/Polymer/polymer-cli/builds/202150101

test/integration/install-variants_test.js
```diff
@@ -29,7 +31,16 @@ suite('install-variants', function() {
           assert.fail(err);
           done(err);
         }
-        runCommand(binPath, ['install', '--variants', '--offline'], {cwd: tmpPath}).then(() => {
+
+        const env = {};
+        for (const envVar of Object.keys(process.env)) {
```
FredKSchott: can you add a comment explaining what you're doing here?

test/integration/run-command.js
```diff
@@ -30,11 +33,49 @@ function runCommand(path, args, options) {
         return;
       }
       if (code !== 0) {
-        reject(new Error('exit code: ' + code));
+        reject(new Error(
+          `Error running ${path} with args ${args}. Got exit code: ${code}`));
         return;
       }
       resolve();
```
FredKSchott: It would be helpful to resolve with the `pipesToString()` contents in case a consumer needs it.

FredKSchott:
Integration tests now run after a full `npm install` in this directory, thanks to npm's prepublish/install behavior. It might be enough to for our team to say "use yarn or wait for npm 4 to stop doing this" but its something to be aware of.

test/integration/run-command.js
```diff
@@ -30,11 +33,49 @@ function runCommand(path, args, options) {
...
     });
+  }).catch((err) => {
+    // If the command was successful there's no need to print anything to the
+    // console, but if it failed then its output is probably helpful.
+    // Print out the output, then reject the final result with the original
+    // error.
+    return contents.then((out) => {
+      console.log(`Output of failed command 'node ${path} ${args.join(' ')}' in directory ${options.cwd}`);
+      console.log(out);
```
FredKSchott: and just to be 100% sure, you've tested this with the `silent` option and it logs all commmand output that was previously silenced?

rictic: Yes, I've confirmed that the output is successfully sent out. See e.g. https://travis-ci.org/Polymer/polymer-cli/jobs/202104841#L829

test/integration/install-variants_test.js
rictic: Sure, pulled out into a little method and documented that.

test/integration/run-command.js
rictic: Done

PR #564, opened 2017-02-11
Give types to lazy loaded libraries

rictic:
- [x] internal change, CHANGELOG.md has not been updated

### DIFF
--- a/src/commands/build.ts
+++ b/src/commands/build.ts
@@ -15,8 +15,12 @@
 // Only import type definitions here, otherwise this line will be included in
 // the JS output, triggering  the entire build library & its dependencies to
 // be loaded and parsed.
+import * as delTypeOnly from 'del';
+import * as path from 'path';
 import * as logging from 'plylog';
 import {ProjectBuildOptions, ProjectConfig} from 'polymer-project-config';
+
+import * as buildLibTypeOnly from '../build/build';
 
 import {Command, CommandOptions} from './command';
 
@@ -84,9 +88,8 @@ export class BuildCommand implements Command {
 
   async run(options: CommandOptions, config: ProjectConfig): Promise<any> {
     // Defer dependency loading until this specific command is run
-    const del = require('del');
-    const pathSeperator = require('path').sep;
-    const buildLib = require('../build/build');
+    const del = require('del') as typeof delTypeOnly;
+    const buildLib = require('../build/build') as typeof buildLibTypeOnly;
     let build = buildLib.build;
     const mainBuildDirectoryName = buildLib.mainBuildDirectoryName;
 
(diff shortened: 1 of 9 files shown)


src/commands/build.ts
```diff
@@ -15,9 +15,13 @@
 // Only import type definitions here, otherwise this line will be included in
 // the JS output, triggering  the entire build library & its dependencies to
 // be loaded and parsed.
+import * as delTypeOnly from 'del';
+import * as path from 'path';
```
FredKSchott: Was there a reason to move this above? it probably doesn't make a huge difference performance-wise, but it contradicts the comment above and goes against our overall strategy of deferring loading all dependencies used during `run()`.

rictic: I think the guideline should be to defer heavyweight imports before running commands, rather than not loading anything at all. That said, I reverted the `path` import here.

rictic:
I reworded the comment slightly and propagated it throughout the `commands/` dir. (clang-format's import reordering makes it hard to section of the type imports from the others like `polylog`, which is why I rewrote the comment)

src/commands/help.ts
```diff
@@ -30,15 +35,14 @@ const CLI_DESCRIPTION = 'The multi-tool for Polymer projects';
...
-const HELP_HEADER = '\n' +
-  b('   /\\˜˜/   ') + m('/\\˜˜/') + b('\\   ') + '\n' +
-  b('  /__\\/   ') + m('/__\\/') + b('__\\  ') + '  ' + CLI_TITLE + '\n' +
-  b(' /\\  /   ') + m('/\\  /') + b('\\  /\\ ') + '\n' +
-  b('/__\\/   ') + m('/__\\/  ') + b('\\/__\\') + '  ' + CLI_DESCRIPTION + '\n' +
-  b('\\  /\\  ') + m('/\\  /   ') + b('/\\  /') + '\n' +
-  b(' \\/__\\') + m('/__\\/   ') + b('/__\\/ ') + '  ' + CLI_USAGE + '\n' +
-  b('  \\  ') + m('/\\  /   ') + b('/\\  /  ') + '\n' +
-  b('   \\') + m('/__\\/   ') + b('/__\\/   ') + '\n';
+const HELP_HEADER = '\n' + b('   /\\˜˜/   ') + m('/\\˜˜/') + b('\\   ') + '\n' +
```
FredKSchott: `// clang-format off` should stop this

src/commands/install.ts
```diff
@@ -37,7 +45,8 @@ export class InstallCommand implements Command {
   ];
 
   async run(options: CommandOptions, _config: ProjectConfig): Promise<void> {
-    const install = require('../install/install').install;
+    const install =
```
FredKSchott: Add defer loading comment here

src/init/init.ts
```diff
@@ -280,7 +280,7 @@ export async function runGenerator(
  * Prompt the user to select a generator. When the user
  * selects a generator, run it.
  */
-export async function promptGeneratorSelection(options: {[name: string]: any}):
+export async function promptGeneratorSelection(options?: {[name: string]: any}):
```
FredKSchott: nice catch!

src/commands/install.ts
rictic: done

src/commands/help.ts
rictic: Huh, that's weird. Maybe a bug with the vscode plugin?

What do you think about this rewrite as a template string? I've verified that it produces visually identical output.

PR #566, opened 2017-02-14
Add new linter alongside the old one.

rictic:

Tests pass locally, should pass on travis once the following PRs land:

- [x] https://github.com/Polymer/polymer-linter/pull/21
- [x] https://github.com/Polymer/polymer-bundler/pull/428 
- [x] https://github.com/Polymer/polymer-bundler/pull/429
- [x] https://github.com/Polymer/polymer-build/pull/128
- [x] https://github.com/Polymer/polymer-cli/pull/568
- [x] https://github.com/Polymer/polymer-linter/pull/22
- [x] #563


CHANGELOG.md
```diff
@@ -7,7 +7,7 @@
   - **Previously default behaviors (minifying JavaScript, generating service workers, etc) are now turned off by default.**
 - `init`: Add new 2.0 polymer element & application templates.
 - Update dependencies.
-
+- **New `lint` Implementation**: now configurable with per-project rulesets, either on CLI or in your polymer.json. Will soon replace the `lint` command, for now run it as `polymer experimental-lint`. Specify "polymer-2", "polymer-2-hybrid", or "polymer-1" to customize the lint warnings that you receive. Run `polymer help experimental-lint` for more detail.
```
FredKSchott: The title makes it sound like we've replaced the lint command, which confused me when I skimmed this and then tried to use it. Would be helpful to remove the \` symbols around lint so that it's not formatted like a command name, or at least adding beta/experimental somewhere in the bolded title.

src/commands/lint.ts
```diff
@@ -17,117 +17,182 @@
...
+   *   - whether to use color (also: can we autodetect if color is supported?)
+   *   - add option for input files to polymer.json
+   *   - modules to load that can register new rules
+   *   - --watch
+   *   - --fix
+   */
+
+  async run(options: Options, config: ProjectConfig): Promise<any> {
+    // Defer dependency loading until this specific command is run.
+    const lintLib: typeof lintLibTypeOnly = require('polymer-linter');
```
FredKSchott: This is getting a little too long to manage, consider moving logic into a new `/lint` subdirectory to simplify the command itself into a single require/import call. (see `build/` and `init/` for examples where we've done this).

src/commands/lint.ts
```diff
@@ -17,117 +17,182 @@
...
+
+    const warnings = await this._lint(linter, options.input);
+
+    const filtered = warnings.filter((w) => !filter.shouldIgnore(w));
+
+    const printer = new WarningPrinter(
+        process.stdout, {analyzer: analyzer, verbosity: 'full', color: true});
+    await printer.printWarnings(filtered);
+
+    process.exitCode = this._getExitCode(filtered);
```
FredKSchott: There is already error handling in place to log the error and return the right exit code on `run()` rejection. I'd much prefer that we use that flow here as well instead of modifying the `process` object directly.

When multiple build errors happen, an error with a message like "3 error(s) occurred during build" is thrown and logged. Something similar with `filtered.length` could work well here.

src/commands/lint.ts
```diff
@@ -17,117 +17,182 @@
...
         })
-        .then(() => undefined);
+        .join('\n');
+  }
+
+  private _loadPlugins(_config: ProjectConfig) {
+    // TODO(rictic): implement.
+  }
+
+  private async _lint(
```
FredKSchott: This may be personal preference, but I prefer to keep the action of a main function like `run()` inlined instead of abstracted. The contents of this helper function are small enough that the abstraction doesn't meaningfully simplify the main function it was taken from (and caused me to jump back and forth while reading `run()` to understand what it was actually doing).

src/commands/lint.ts
rictic: I like the idea of printing a message that summarizes the results when warnings are found.

Logging a full exception with a stack trace or a mention of a cli runtime error when the linter has successfully reported warnings as the user asked isn't right though. It's distracting and irrelevant noise. A good UI must minimize that wherever possible.

How about this: a new type to reject with, indicating that the command failed but that it has already reported this failure to the user. It includes the error code that the CLI should exit with.

src/commands/lint.ts
rictic: Fair point, agreed. I'd written it this way to make `warnings` a const in the main function, but that is outweighed by the loss of legibility that you point out.

rictic:
(pretend the tests pass. I'm still fighting with travis)

rictic:
(tests now pass)

src/commands/lint.ts
FredKSchott: This is a great solution, and something we probably should have done a long time ago. We could definitely leverage that in build and others as well, where the stack-trace of an expected error message is unnecessary.

CHANGELOG.md
```diff
@@ -7,7 +7,7 @@
   - **Previously default behaviors (minifying JavaScript, generating service workers, etc) are now turned off by default.**
 - `init`: Add new 2.0 polymer element & application templates.
 - Update dependencies.
-
+- **New `experimental-lint` command**: configurable with per-project rulesets, either with cli args or in your polymer.json. Will soon replace the `lint` command, for now run it as `polymer experimental-lint`. Specify "polymer-2", "polymer-2-hybrid", or "polymer-1" to customize the lint warnings that you receive. Run `polymer help experimental-lint` for more detail.
```
FredKSchott: 👍 much more clear

bin/polymer.js
```diff
@@ -30,6 +30,9 @@ resolve('polymer-cli', {basedir: process.cwd()}, function(error, path) {
   let args = process.argv.slice(2);
   let cli = new lib.PolymerCli(args);
   cli.run().catch((err) => {
+    if (err && err.constructor && err.constructor.name === 'CommandFailure') {
+      process.exit(err.exitCode);
```
FredKSchott: i know this exits as a side-effect, but adding a return here as well would be nice.

bin/polymer.js
```diff
@@ -30,6 +30,9 @@ resolve('polymer-cli', {basedir: process.cwd()}, function(error, path) {
   let args = process.argv.slice(2);
   let cli = new lib.PolymerCli(args);
   cli.run().catch((err) => {
+    if (err && err.constructor && err.constructor.name === 'CommandFailure') {
```
justinfagnani: If this is a normal exit, I think it shouldn't be an exception. Let's change `run()` to return a `Promise<Result>`.

package.json
```diff
@@ -88,10 +89,12 @@
     "fs-extra": "^2.0.0",
     "gulp": "^3.9.1",
     "gulp-eslint": "^3.0.1",
-    "gulp-mocha": "^3.0.1",
+    "gulp-spawn-mocha": "^3.1.0",
     "gulp-tslint": "^7.0.1",
     "gulp-typescript": "^3.0.0",
+    "istanbul": "^0.4",
```
FredKSchott: ugh, i hate that it warns on this even if you don't want to use it...

src/commands/command.ts
```diff
@@ -24,4 +25,15 @@ export interface Command {
   description: string;
   args: ArgDescriptor[];
   run(options: CommandOptions, config: ProjectConfig): Promise<any>;
-}
\ No newline at end of file
+  extraUsageGroups?(config: ProjectConfig): UsageGroup[];
```
justinfagnani: comment?

src/commands/lint.ts
```diff
@@ -17,117 +17,105 @@
...
-      defaultValue: false
-    },
-    {
-      name: 'config-file',
-      type: String,
-      defaultValue: 'bower.json',
-      description: (
-          'If inputs are specified, look for `config-field` in this JSON file.')
+      description: 'Files to lint. If given, these files will be the only ' +
+          'ones linted, otherwise all files in the project will be linted.'
```
FredKSchott: does this lint ALL files in your project, or just the ones reachable in the app described by polymer.json? Consider rewording if it's the latter.

src/commands/command.ts
```diff
@@ -24,4 +25,15 @@ export interface Command {
...
+}
+
+/**
+ * Running a command may reject with a CommandFailure to indicate that there
+ * was an error but it has already been reported to the user, and the CLI
+ * should exit with the given exit code.
+ */
+export class CommandFailure {
```
FredKSchott: Great first step! In the future I'd like to see this extend the error type have a message / stack trace / more info that we could output when verbose is true.

src/commands/lint.ts
rictic: It's currently all files in your project that the analyzer knows how to parse. Subsetting to the latter makes sense, will do in a followup.

src/commands/command.ts
rictic: For the case where the linter has successfully printed warnings? Or the build stopped because of warnings (that have already been printed)?

rictic:
Got the verbal LGTM from Justin before he left, merging!

PR #567, opened 2017-02-14
Update polyserve dep to 0.16.0

justinfagnani:


package.json
```diff
@@ -65,7 +65,7 @@
     "polymer-analyzer": "2.0.0-alpha.20",
     "polymer-build": "^0.7.1",
     "polymer-project-config": "^1.2.0",
-    "polyserve": "0.16.0-prerelease.9",
+    "polyserve": "0.16.0",
```
rictic: ^0.16.0 instead, so that we can get semver?

rictic:
I folded this in with the other deps changes in #566

PR #570, opened 2017-02-17
unskip integration tests

FredKSchott:
Now that tools-sample-projects is public we can connect them to the CLI tests.

/cc @rictic


rictic:
Tests look good!

PR #572, opened 2017-02-18
Early exit if the user's node version is too low.

### LINKED ISSUE #571: Throw an obvious error when running on the wrong node version
Could we add a small check to throw a clear error when you're running on the wrong node version?

rictic:
This required converting polymer.js to ES5.

Fixes #571

 - [x] CHANGELOG.md not updated, minor change.


FredKSchott:
YES! Thanks for tackling this 💯 💯 💯 

Instead of forcing this file to be Node v0.X-compliant for the foreseeable future, what do you think about about creating a new file, something like `bin/polymer-safe.js`, that performs this check and then requires `bin/polymer.js` as-is?

rictic:
Ok, yeah!

This is also a good chance to move the rest of this file into `src` where the type system can see it.

src/run.ts
```diff
@@ -0,0 +1,47 @@
...
+
+const logger = logging.getLogger('cli.main');
+
+
+// Update Notifier: Asynchronously check for package updates and, if needed,
+// notify on the next time the CLI is run.
+// See https://github.com/yeoman/update-notifier#how for how this works.
+updateNotifier({pkg: packageJson}).notify();
+
+resolve('polymer-cli', {basedir: process.cwd()}, async(_error, path) => {
```
FredKSchott: We should handle _error here, or at least print the message before exiting

FredKSchott: We should handle _error here, or at least print the message before exiting

FredKSchott: Actually, why don't we just `require('lib/polymer-cli')` here? I think there used to be a reason when we were considering using polymer-cli as a library & a CLI, but this may no longer be necessary. I can't think of a good reason, at least.

If there is a reason, lets comment it for future readers. Otherwise consider removing (which should simplify a lot of this).

/cc @justinfagnani who may remember better

justinfagnani: We _should_ consider just removing this feature, but only very carefully.

It was a bit cargo-culted from WCT and polybuild, where the idea is that they're both a CLI and a library, and a global install of the CLI should defer to the local version of the library if present. Gulp essentially does this, except that it requires a local install - something we did and reverted for WCT 6.

I don't think the CLI is installed much as a library... though it could be. The key factor for keeping this feature is - should we let a project pin the version of the CLI it uses by installing it as a dev dependency. Given the rapid changes we've gone through lately, that seems like a prudent feature to me. We could be a little more clear about the working here though: print the version of CLI being used at each invocation, do update checks on both global and local...

Either way, we shouldn't change this in this PR.

justinfagnani: The error here is the most common case, when the package didn't resolve. It's handled on the next line's check for `path`.

rictic:
Oh, tests failing because npm shrinkwrap needs to be updated. Arg.

PR #579, opened 2017-02-24
Update Analyzer to 2.0.0-alpha.27

justinfagnani:


src/analyze/analyze.ts
```diff
@@ -26,15 +26,15 @@ export async function analyze(
     urlResolver: new PackageUrlResolver(),
   });
 
-  const elements = new Set<Element>();
-  const isInDependency = /(\b|\/|\\)(bower_components|node_modules)(\/|\\)/;
+  const isInTests = /(\b|\/|\\)(test)(\/|\\)/;
```
usergenic: I think we need a TODO to address magic constant here.  Also, it feels like \b should cut it for slashes too, so maybe `\btest\b`?

PR #588, opened 2017-03-06
add polymer 2.0 psk to the init command template list

FredKSchott:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Upcoming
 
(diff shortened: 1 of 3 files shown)


FredKSchott:
Ugh, yeoman has some strange behavior when the list gets big: https://drive.google.com/file/d/0BzBgh-hIVm7PNnVsc09GenlMNG8/view

@justinfagnani this defeats almost all of the value of listing both polymer 1.x & 2.x templates. I'll dig around yeoman/inquierer.js to try to find an option to control this but if none exists I think we'll want a follow up PR to move the version choice to be a separate choice after the template.

PR #589, opened 2017-03-07
Use properties getter with 2.0

ruphin:
Small fix to make the init templates work with the new Polymer 2.0.0-rc.1 style


ruphin:
I signed it!

justinfagnani:
Thanks @ruphin!

Looks like the integration tests are failing... This could actually be because of Polymer tagging a pre-release. I'll try to look into that today.

justinfagnani:
@ruphin I fixed the tests on master, so if you rebase this PR should go ✅

ruphin:
Should be ready to go

PR #590, opened 2017-03-07
Update Polymer 2.x templates to use RC1

justinfagnani:
Also skips failing integration tests that depended on external repos.

PSK is fixed in https://github.com/PolymerElements/polymer-starter-kit/pull/979

### DIFF
--- a/test/integration/integration_test.js
+++ b/test/integration/integration_test.js
@@ -100,7 +100,10 @@ suite('integration tests', function() {
         .then(() => runCommand(binPath, ['build'], {cwd: dir}));
     });
 
-    test('test the "starter-kit" template', () => {
+    // TODO(justinfagnani): consider removing these integration tests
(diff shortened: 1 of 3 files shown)


justinfagnani:
I'm really not so sure about continuing to have integration tests that depend on changing external resources. It's great to know that PSK doesn't bower install because Polymer tagged a release, but that should really be part of PSK's tests and not cause our tests to go red, slowing us down.

test/integration/integration_test.js
```diff
@@ -100,7 +100,7 @@ suite('integration tests', function() {
         .then(() => runCommand(binPath, ['build'], {cwd: dir}));
     });
 
-    test('test the "starter-kit" template', () => {
+    test.skip('test the "starter-kit" template', () => {
```
rictic: Add comments for why these are skipped and when to unskip them please

PR #592, opened 2017-03-07
skip optimizing for any web-component polyfills

FredKSchott:
This is a quick fix so that we aren't blocking 2.0 release. After release, we should better address this (ex: as an option for the user to specify).


justinfagnani:
@fks we need to pull in the webcomponentsjs/*.js files into build also

PR #604, opened 2017-03-10
fix polymer-2-application generated unit test

### LINKED ISSUE #603: WCT tests fail on freshly generated polymer-2-application
### Description

After generating a fresh project using `polymer init` and selecting `polymer-2-application`. The WCT test fails.

I'm pretty sure that this is because you can't get a static property from an instance directly. ie:

```javascript
assert.equal(element.is, 'polymer-2-test-app');
```

should be

```javascript
assert.equal(element.constructor.is, 'polymer-2-test-app');
```

### Versions & Environment

- Polymer CLI: 0.18.0-pre.13
- node: v7.7.1
- Operating System: macOS

#### Steps to Reproduce

```
polymer init
# choose polymer-2-application
polymer test
```

#### Expected Results

Tests pass

#### Actual Results

Tests fail with

```
expected undefined to equal 'polymer-2-test-app'
```

pghalliday:
Fixes #603 

### DIFF
--- a/src/init/application/templates/polymer-2.x/test/_element/_element_test.html
+++ b/src/init/application/templates/polymer-2.x/test/_element/_element_test.html
@@ -22,7 +22,9 @@
       suite('<%= elementName %>', function() {
         test('instantiating the element works', function() {
           var element = fixture('basic');
-          assert.equal(element.is, '<%= elementName %>');
+          var elementClass = window.customElements.get('<%= elementName %>');
+          assert.isOk(elementClass);
+          assert.isTrue(element instanceof elementClass);
         });
       });
     </script>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.18.0-pre.13 [03-10-2017]
+
+- polymer-2 generated unit tests should pass
+
 ## v0.18.0-pre.13 [03-08-2017]
 
 - When running `polymer build` and compiling JS to ES5, we will also rewrite script includes of `webcomponents-loader.js` to `webcomponents-es5-loader.js`.
--- a/src/init/element/templates/polymer-2.x/test/_element_test.html
+++ b/src/init/element/templates/polymer-2.x/test/_element_test.html
@@ -22,7 +22,9 @@
       suite('<%= name %>', function() {
         test('instantiating the element works', function() {
           var element = fixture('basic');
-          assert.equal(element.is, '<%= name %>');
+          var elementClass = window.customElements.get('<%= name %>');
+          assert.isOk(elementClass);
+          assert.isTrue(element instanceof elementClass);
         });
       });
     </script>


src/init/application/templates/polymer-2.x/test/_element/_element_test.html
```diff
@@ -22,7 +22,7 @@
       suite('<%= elementName %>', function() {
         test('instantiating the element works', function() {
           var element = fixture('basic');
-          assert.equal(element.is, '<%= elementName %>');
+          assert.equal(element.constructor.is, '<%= elementName %>');
```
rictic: I think there may be some issues with `constructor` when compiled to ES5?

How about something like:

```javascript
assert.true(!!customElements.get('<%= elementName %>'));
assert.true(element instanceof Polymer.Element);
```

rictic:
Thanks for the PR!

Would you mind also updating the element template too?

pghalliday:
@rictic, what's wrong with the element template?

rictic:
It looks like it has the same test assertion as the application template

pghalliday:
ah, I think iwas confused about what the element template was :) I'll take a look

src/init/application/templates/polymer-2.x/test/_element/_element_test.html
rictic: Hm, there's a few Polymer base classes. How about:

```javascript
assert.true(!!customElements.get('<%= elementName %>'));
assert.true(element instanceof customElements.get('<%= elementName %>'));
```

rictic:
It's a sibling directory to `application`

src/init/application/templates/polymer-2.x/test/_element/_element_test.html
pghalliday: This doesn't actually assert that the element is an `<%= elementName %>` does it?

Just that there is a such an element type and this element is a sub class  of `Polymer.Element`

Also i'm not sure that tests get compiled to ES5 (i've been trying to figure that out too) and the element under test is using ES6 classes

rictic: My second suggestion addresses the concern that the element is an upgraded custom element of the expected tagname. We could add:

```javascript
assert.true(Polymer.telemetry.registrations.indexOf(element.localName) !== -1);
```

To also assert that the Polymer element class was finalized without issue (all the properties, observers, and templates have been associated to the class)

pghalliday: I'll be honest - I have no clue what that line of code means :o

pghalliday: Would this be better, for the sake of clarity and also in keeping with the intent of the original test:

```
assert.equal(element.localName, '<%= elementName %>');
```

That passes

pghalliday: I would be concerned about putting a number of obtuse tests that might be hard to understand, given that this should really be an example for people (in reality they probably won't be testing the element name and will actually just be testing that it behaves as they expect)

rictic: `element.localName` will pass pretty much no matter what, even if there's no registered custom element with that name. E.g. paste this into your devtools on this very page: `document.createElement('foo-bar').localName`

The Polymer telemetry assert is a little crazy, that's fair. What about just the other two?

```javascript
// Test that the element was registered properly.
assert.true(!!customElements.get('<%= elementName %>'));
assert.true(element instanceof customElements.get('<%= elementName %>'));
```

That'll end up looking like:

```javascript
// Test that the element was registered properly.
assert.true(!!customElements.get('my-element'));
assert.true(element instanceof customElements.get('my-element'));
```

pghalliday: Thinking about it, a better example might be something like this

```
assert.equal(element.$.heading.textContent, 'Hello World!');
```

and adding the `prop1` to the fixture, and giving the `<h2>` in the element an id of `heading`

tony19: Instead of `assert.true(!!foo)`, how about [`assert.isOk(foo)`](http://chaijs.com/api/assert/#isokobject-message)?

pghalliday:
How about this new test, I think it's a better example of a good unit test as it tests the behavior

rictic:
I think an assert with `customElements` is better as a first test, as it gives very useful diagnostic info when it fails. i.e. that registering the element failed.

There's lots of reasons why some text content might not be right.

I wouldn't be opposed to adding the text content test as an additional test.

pghalliday:
Interestingly, when I `console.log(element.textContent)` I don't get anything that I can really use so I don't know what I should expect there. The reason I added the `id` attribute to the `h2` was to provide some abstraction from the internal implementation and make it part of the public interface - but I see your point.

Anyway I have reverted that and just fixed the instantiation test.

FredKSchott:
Thanks for the contribution @pghalliday! Closing this issue because this was just fixed in a greater rewrite of the application templates to make their tests more thorough: https://github.com/Polymer/polymer-cli/pull/649

PR #607, opened 2017-03-14
use network for bower install during integration tests

FredKSchott:
Fixes broken travis tests (and should prevent similar test failures in the future).


rictic:
It looks like the offline install always fails on travis. This is an improvement over what we have, but I think it'd be more straightforward to just notice that we're on travis and jump straight to the online install rather than trying and falling back.

More better still of course would be to get offline working on travis, ofc.

FredKSchott:
Hmm, I'd assumed it was working at one point on Travis, and that a passing test might somehow update the cache. You might be right though, let me look into how Travis's bower cache works.

rictic:
Perversely, I think it was working when we had tests running in prepublish, as something was causing it to succeed the second time.

FredKSchott:
...

we have to go back

FredKSchott:
You're right, Travis has some caching support across builds but nothing for bower. https://docs.travis-ci.com/user/caching/#Caching-directories-(Bundler%2C-dependencies)

Lets just turn this into a network install in all environments. Even if we try to be smart and keep offline for local test-running we'll still see broken tests whenever a users cache is out of date.

rictic:
So what the code is currently doing _should_ be resilient against issues in the user's cache, as we've checked in a full bower cache into the integration tests directory and configured bower to use it. The tricky part is that bower appears to be listening to some other configuration signal instead of the ones we're giving it, causing it to try to load the resources from another cache that doesn't contain webcomponentsjs (and probably not many of the other packages either).

It'd be way better if we could get it to respect our offline configuration. We could save maybe 40 seconds off of our integration tests and start running them as a normal part of `npm test` if we converted all of them to be offline.

The test time issue by itself isn't as big as it once was, since it looks like removing the prepublish and using yarn have already gotten our travis times down from 20 minutes to 3.5. But it means that we won't be able to go back and run our tests at old versions, and they'll fail unpredictably due to circumstances outside out control, which is not a great place to be in.

FredKSchott:
Ah, I hadn't even seen our local `bower_cache` directory. I just figured we were solely relying on that presubmit install to cache things for us.

So I've dug into bower's offline/storage support some more and we were missing a crucial part of offline support. Even though the packages were being cached locally, bower was still hitting the registry (or Travis's globally cached registry) to find the correct package for each semver range. So even though we had a version of webcomponents cached, it wasn't the one that bower wanted based on the latest info it had gotten from github. 

This adds a local cache of the registry now too. I tested this with a clean global cache and my network connection completely cut and installs still completed.

I'm not super impressed with bower's offline support (or documentation) so far, so if this breaks again I will probably suggest we go back to online installs (timing wise, this only saves us 1 second).

test/integration/fixtures/install-variants/bower-registry/bower.herokuapp.com/lookup/webcomponentsjs_7bba9
```diff
@@ -0,0 +1 @@
+{"expires":1489941177396,"value":{"type":"alias","url":"https://github.com/Polymer/webcomponentsjs.git"}}
```
rictic: 1489941177396 is this Sunday. Should we push this off to the impractically distant future?

2489941177396 is in 2048.

Also, how did you generate this?

FredKSchott:
That's a great point, we *shouldn't* need to ever generate this again but I'll add README instructions just in case.

test/integration/fixtures/install-variants/bower-registry/bower.herokuapp.com/lookup/webcomponentsjs_7bba9
FredKSchott: Good find, I double checked and expired lookups don't affect offline installs, only online refreshes of the cache. Still did it anyway, and added a note about it to the README instructions in case it changes.

FredKSchott:
@rictic PTAL

rictic:
Awesome! Mind doing the same for the other integration tests?

FredKSchott:
:/ I had tried looking into that earlier as well, but because the directories don't exist until after the `init` generator is run, it's not that easy. Also, because we don't own the repos for shop & psk, a cache stored in the CLI would quickly become out of date.

I think we need to have a deeper conversation about what we're trying to accomplish with our init-baed integration tests. The original purpose was to test that our commands all run on each of our pre-installed `init` templates without errors. If they suddenly break downstream we actually do want to know asap because it means our users are also seeing those errors. The problem is that this obviously conflicts with our other goal of "downstream changes shouldn't unexpectedly break our CI."

For now, I think it makes sense to pull the external-dependent tests out of CI and into a separate gulp task that can be run as needed. @rictic wdyt?

We can also look into refactoring our tests to use https://docs.travis-ci.com/user/speeding-up-the-build/#Parallelizing-your-builds-across-virtual-machines. This would allow us to see green on unit & internal integration tests even if network-dependent integration tests turn red, although I believe github will still see travis failures.

justinfagnani:
Sweet! Awesome work Fred. Reproducible tests ftw. I hope we can drive our full test suite time down further too.

rictic:
The argument in favor of hermetic tests is:
  * test results are reproducible. if a test passes locally, it should pass on travis. if it passes today it should pass in six months
  * it separates out tasks. if someone adds a feature to `build` and our `init` tests fail because of a change upstream then either we'll merge on red or the feature will be blocked by unrelated work

I think google's internal system is ~best, where the change that was responsible for a test going red would be blocked from landing, or rolled back. We can approximate that by using the CLI to test shop and psk and elements in a similar manner to what we do here.

rictic:
Er, pressed comment too early, still editing the above

FredKSchott:
In case it came off as otherwise, I'm all for reproducible, hermetic tests (especially in CI). I'm just also in favor of tests that accurately reflect the world of the current user. Because our users are pulling in our "blessed" but externally developed templates and running our commands on them, I disagree that snapshotting the past and assuming that its an accurate representation of the present is our next best solution. We would have no signal when our templates break. I believe this would be doing our users a disservice.

If we have to choose between the two, I vote for live updating tests. Yes that means that external changes can break our tests, but only when external changes have already broken our users work. I think the benefit outweighs the cost there.

Regardless of what we decide here, I'm going to look into running the latest version of the CLI in the shop & psk test suite tonight. Installing and running a few commands via Travis should actually only require a few lines of bash.

rictic:
why not both?

What about a hermetic, offline test that runs as part of `npm test` and then a separate command that runs online tests? Ideally with a separate dot in travis that doesn't count our build as broken but does let us know that there is a problem. This, combined with another npm command to update the offline tests and our tests would be in a really good place. WDYT?

justinfagnani:
@rictic that's exactly what I want to do, but I think we should consider this as part of a broader effort to finally get cross-repo testing happening for tools, elements and core. We can establish a pattern first here, which I imagine possibly being a separate project whose tests run on changes to multiple repos. I'm not sure how tricky that might be to set up in Travis, but it would give us another badge on projects that says whether the current combination passes together.

For now, I'd like to get the offline fix in which has all the pros of fixing tests, making them faster, and stable, and reproducible. Win. 

Unless anyone has last objections, I kind of want to merge this ASAP, because we have PRs that are failing, and then apply this change to the other integration tests later.

FredKSchott:
We've had a separate integration test command before that was rarely run and would be broken often. If we couldn't get it into Travis, I'm still not convinced we'd behave any differently this time around. If we could get it into Travis in a way that didn't "break" the overall build, that would be best in my mind as well.

Regardless, this PR doesn't skip or unskip any tests so it shouldn't affect this discussion. @rictic can you approve and we can merge to get travis green again? 

Happy to still keep the discussion going after this PR has merged.

PR #612, opened 2017-03-17
Switch polymer-2.x templates to use polymer-element.html.

coreyfarrell:
These templates create class based elements that extends
Polymer.Element.  Importing the full polymer.html is not needed.

I'm not sure this needs a CHANGELOG.md entry?  If so do I create a new section for v0.18.0-pre.14?


justinfagnani:
Thanks for the PR @coreyfarrell!

A couple of small requests:

1. Since `polymer-element.html` doesn't include dom-if and dom-repeat, could you add commented out imports for those so that users can see what they need to do if they add them to the template?
2. While you're here, can you add doc comments to the classes with `@polymerElement` to them?

Thanks!

coreyfarrell:
1. Done.
2. polymer-analyzer@2.0.0-alpha.33 detects class based polymer elements without the `@polymerElement` annotation.  Since that comment is no longer needed by the analyzer what other purpose does it serve?  Obviously polymer-cli needs an update to depend on alpha.33 instead of alpha.31.  Unless that comment has another technical purpose that I think it adds bulk with no real value.

https://github.com/Polymer/polymer-analyzer/commit/cd1c8ea29e80b9522343b773ef2dea225aeb314b

justinfagnani:
Thanks.

As for the annotation, it's the only way to definitively mark a class as a Polymer element. Inferring from a `customElements.define()` call works in certain situations. For base classes that are never registered, or for cases where someone registers in another document or using some wrapper on define. I'd like to include it in the template as the canonical method. I can add this myself later though.

coreyfarrell:
I added the annotation.

FredKSchott:
@justinfagnani whoops, approved @sorvell's PR before seeing your comments here. This PR can continue to tackle the improvements on top of the move to `polymer-element.html`. 

LGTM, I'll leave to you to approve & merge.

PR #617, opened 2017-03-19
revise README to have consistent information

ev1stensberg:
I tried this approach, as how I understood how it works. Copied a section from the v.18 changelog, so hopefully that is correct :) #609  relative

### DIFF
--- a/src/init/element/templates/polymer-2.x/README.md
+++ b/src/init/element/templates/polymer-2.x/README.md
@@ -18,9 +18,8 @@ $ polymer serve
 $ polymer build
 ```
 
-This will create a `build/` folder with `bundled/` and `unbundled/` sub-folders
-containing a bundled (Vulcanized) and unbundled builds, both run through HTML,
-CSS, and JS optimizers.
+This will create a `build/` folder with `default/` and `bower_components/` sub-folders
+containing a unbundled build, run through HTML, CSS, and JS optimizers. New build options have been added to give you more control over the generated build. These options can be defined in your project's `polymer.json`, or via CLI flags. Run `polymer build --help` to see a list of new supported CLI flags.
 
 You can serve the built versions by giving `polymer serve` a folder to serve
 from:


src/init/element/templates/polymer-2.x/README.md
```diff
@@ -18,9 +18,8 @@ $ polymer serve
 $ polymer build
 ```
 
-This will create a `build/` folder with `bundled/` and `unbundled/` sub-folders
-containing a bundled (Vulcanized) and unbundled builds, both run through HTML,
-CSS, and JS optimizers.
+This will create a `build/` folder with `default/` and `bower_components/` sub-folders
```
justinfagnani: it shouldn't produce a `bower_components/` folder outside of the individual build folders.

src/init/element/templates/polymer-2.x/README.md
```diff
@@ -18,9 +18,8 @@ $ polymer serve
...
+containing a unbundled build, run through HTML, CSS, and JS optimizers. New build options have been added to give you more control over the generated build. These options can be defined in your project's `polymer.json`, or via CLI flags. Run `polymer build --help` to see a list of new supported CLI flags.
```
justinfagnani: whether it's bundled or unbundled is determined by the build options

ev1stensberg:
This seems to me a bug that's not related to the README. Could you give me some info on how it should look in v2 and the default behaviour? I'll be happy to rephrase

ev1stensberg:
If there's no update on this, I'm going to go ahead and close this by the end of the week :)

justinfagnani:
Hey @ev1stensberg, I'm just getting back from vacation today, give me a couple of days.

ev1stensberg:
@justinfagnani That's fine, speak soon!

FredKSchott:
Thanks for the contribution @ev1stensberg! All of the templates have README issues, so we're going to fix them all at once in #656

PR #630, opened 2017-03-27
Update WCT for Polymer 1.x templates

stramel:

Updating the Polymer 1.x application and element templates to latest released version of WCT. Seems odd that cli and 2.0 were updated a while back but the element templates haven't.

Wasn't sure if we wanted to update them to the pre-release WCT 6 or not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - `init`: Init template elements now properly inherit from the given element/app name.
 - `init`: For `polymer-2-element` template, remove iron-component-page from the served `index.html` file until it can support Polymer 2.0 class-based elements.
 - `init`: Updates polymer 2.0 application & element tests to improve and fix broken tests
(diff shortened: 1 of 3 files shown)


stramel:
Bump?

FredKSchott:
hey @stramel sorry for the delay. We should actually try to match the version of WCT we use in the CLI, which is `^6.0.0-prerelease.5`. Can you change the version to match that semver range for all templates and then we can merge this?

stramel:
@FredKSchott Sure thing! Will have that pushed in a minute! 

Just going to ask this here: 

> Should this repository be tagged deprecated and redirect people to the polymer-cli? Or should the polymer-cli handle the element similarly to how the custom-build and shop are handled?
[PolymerLabs/generator-polymer-init](https://github.com/PolymerLabs/generator-polymer-init/issues/12)

FredKSchott:
thanks @stramel! And yes, that repo should be deprecated. /cc @justinfagnani who owns the repo & npm package

Issue #634, opened 2017-03-28
Build with compile to ES5 doesn't include custom-elements-es5-adapter.js


danners:
### Description
I tried making a Polymer 2 component that includes components written in the new es 6 syntax work with IE11. For that i used polymer build on a index.html that includes webcomponents-lite and also includes a Polymer 2 webcomponent written in ES6. It transpiled succesfully, however the result is not working since it doesn't include custom-elements-es5-adapter.js. When i add this manually to the result the index.html can be loaded and the webcomponents can be loaded too in IE11.

Looking at the code from polymer build it seems that custom-elements-es5-adapter.js is only included when the webcomponent loader is used. Can this be changed to be included also when webcomponents-lite.js  is included ?

### Versions & Environment

- Polymer CLI: 0.18.0-pre.15

#### Expected Results
custom-elements-es5-adapter.js is included when i compile to es5

#### Actual Results
its not included only when i use the loader.js in the index.html

rictic:
@FredKSchott who is working on this feature noe

FredKSchott:
@danners thanks for bringing this up. I'll make sure the new fix will handle both `-loader` & `-lite`

tiagorg:
I am having a similar issue and I found a solution that works for me. 

Disclaimer: I don't use an app-shell because I have a server-side rendered site with just a few isolated Polymer components on the client-side.

After intense debugging the issue came down to the fact that I was including this block (as suggested on the Polymer guides):

```
<div id="autogenerated-ce-es5-shim">
  <script type="text/javascript">
    if (!window.customElements) {
      var ceShimContainer = document.querySelector('#autogenerated-ce-es5-shim');
      ceShimContainer.parentElement.removeChild(ceShimContainer);
    }
  </script>
  <script type="text/javascript" src="/vendors/webcomponentsjs/custom-elements-es5-adapter.js"></script>
</div>
```

Because when I ran `polymer build` it would pick up `custom-elements-es5-adapter.js` from there. 

This is what I did instead:

```
<script type="text/javascript">
  if (!window.customElements) {
    document.write('<scr' + 'ipt type="text/javascript" src="/vendors/webcomponentsjs/custom-elements-es5-adapter.js"></scr' + 'ipt>');
  }
</script>  
```

YES it's not as elegant and quite rustic but, hey, it works! Here I'm tricking the compilers inside `polymer build` and they won't find this file and hence they won't include it in the build.

I hope it helps!

PR #635, opened 2017-03-28
properly inject es5 adapter for browsers without es5 support

### LINKED ISSUE #627: IE11 syntax error in custom-elements-es5-adapter
#625 introduces a bug whereby IE11 will throw a syntax error when loading the custom-elements-es5-adapter. This will not cause any other errors, and pages will work fine, there will just be an ominous error reported in the console.

We have a fix for this in mind, but it's somewhat complex and we want to unblock #605 first, as it does break pages.

The fix is for our compiling pipelines to insert some code that will run `custom-elements-es5-adapter.js` reliably, in a blocking and CSP-compliant manner, if and only if the browser natively supports custom elements. This rules out `document.write` which Chrome may skip for devices on 2G networks, `eval` which is not CSP-compliant, and appending the script dynamically using dom methods which does not block. The solution @azakus @FredKSchott and I came up with is to conditionally remove the parent element from the DOM. This means that we also have to move† the script import of the es5 loader into the body if it isn't there already, as no `<head>`-legal elements can have script children. The desired output looks something like this:

```html
  <div id="ce-es5-shim">
    <script>
      if (!window.customElements) {
        var ceShimContainer = document.querySelector('#ce-es5-shim');
        // This prevents custom-elements-es5-adapter.js from parsing or running.
        ceShimContainer.parentElement.removeChild(ceShimContainer);
      }
    </script>
    <script src="custom-elements-es5-adapter.js"></script>
  </div>
  <script src="webcomponents-loader.js"></script>
```

† moving a script out of head also means moving all subsequent styles, scripts, and links from the head too, in order to maintain the correct order

FredKSchott:
- [X] Fixes #627 #634

/cc @rictic @valdrinkoshi @azakus


src/build/rewrite-webcomponents-loader.ts
```diff
@@ -12,16 +13,43 @@ const attrValueMatches = (attrName: string, regex: RegExp) => {
...
+        predicates.NOT(predicates.hasAttr('type')),
+        predicates.hasAttrValue('type', 'text/css')));
+
 
 /**
  * When compiling ES6 classes down to ES5 we need to include a special form of
- * the webcomponents loader to be compatible with native custom elements.
+ * the webcomponents loader to be compatible with native custom elements. This
+ * includes moving the loader script element and all relevent siblings
+ * following it into the body so that the <ES5 polyfill can be properly
```
valdrinkoshi: Instead of ES5 polyfill can we be more specific and say it is the Custom Elements ES5 adapter?

src/build/rewrite-webcomponents-loader.ts
```diff
@@ -54,16 +82,57 @@ export class UseES5WebcomponentsLoader extends stream.Transform {
...
+</div>
+`);
+
+    // Get references to all relevent siblings following the "loader" script
+    const loaderScriptSiblings = dom5.queryAll(
+        script.parentNode!,
+        dom5.predicates.OR(scriptMatcher, styleMatcher, linkMatcher));
+    const siblingsAfterLoader =
+        loaderScriptSiblings.splice(loaderScriptSiblings.indexOf(script));
+    dom5.insertBefore(headElement, script, loaderMovedComment);
```
valdrinkoshi: Here we assume the loader is a child of `<head>`, but that might not be the case. If the loader is already a child of `<body>` it should be enough to just insert the es5 adapter before it, w/o moving the siblings after it

FredKSchott: good catch, will update to handle this case.

src/build/rewrite-webcomponents-loader.ts
```diff
@@ -49,24 +78,73 @@ export class UseES5WebcomponentsLoader extends stream.Transform {
...
+      return;
+    }
+
+    // otherwise we need to move the webcomponents-loader/webcomponents-lite
+    // loader down to the body so that the es5 adaperter script shim will work
+    const scriptSiblings = dom5.queryAll(
+        script.parentNode!,
+        dom5.predicates.OR(scriptMatcher, styleMatcher, linkMatcher));
+    const scriptSiblingsFollowing =
+        scriptSiblings.splice(scriptSiblings.indexOf(script));
```
rictic: I think you want `slice` rather than `splice` here, as the latter actually mutates the array in place(!)

rictic: And recommend adding one here rather than starting with 1 when you iterate below

src/build/rewrite-webcomponents-loader.ts
```diff
@@ -49,24 +78,73 @@ export class UseES5WebcomponentsLoader extends stream.Transform {
...
+    dom5.insertBefore(headElement, script, loaderMovedComment);
+    if (!bodyElement.childNodes) {
+      bodyElement.childNodes = [];
+    }
+    for (let i = 1, len = scriptSiblingsFollowing.length; i <= len; i++) {
```
rictic: I have a phobia of complex for loops. Indexing is so finicky, and it's so easy to get it slightly wrong. It'd also be worth a comment about why we're doing this backwards. So I'd recommend instead:

```typescript
// Elements are pushed into the top of body, so need to insert them in reverse order.
scriptSiblingsFollowing.reverse();
for (const releventScript of scriptSiblingsFollowing) {
  dom5.insertBefore(
          bodyElement, bodyElement.childNodes[0] || null, releventScript);
}
```

src/build/rewrite-webcomponents-loader.ts
FredKSchott: we actually do want to splice, I'll add a comment

src/build/rewrite-webcomponents-loader.ts
FredKSchott: +1 that's way way cleaner. I have trouble committing obviously unoptimized code, I'm making a mental note to be better at that when the readability/complexity suffers so much.

src/build/rewrite-webcomponents-loader.ts
FredKSchott: correction: we *used* to need to do splice when we were operating directly on the childElements array. No longer needed, good catch

FredKSchott:
@rictic PTAL

src/build/rewrite-webcomponents-loader.ts
```diff
@@ -49,25 +73,78 @@ export class UseES5WebcomponentsLoader extends stream.Transform {
...
-    dom5.insertBefore(script.parentNode!, script, es5AdapterScript);
-
+    // Collect important dom references & create fragments for injection
     const correctedFile = file.clone();
+    const headElement = dom5.query(parsed, headMatcher)!;
+    const bodyElement = dom5.query(parsed, bodyMatcher)!;
+    const loaderScriptUrl = dom5.getAttribute(script, 'src')!;
+    const adapterScriptUrl =
+        url.resolve(loaderScriptUrl, 'custom-elements-es5-adapter.js');
+    const loaderMovedComment = dom5.constructors.comment(
```
rictic: Recommend moving this down to the place where it's used.

src/build/rewrite-webcomponents-loader.ts
```diff
@@ -49,25 +73,78 @@ export class UseES5WebcomponentsLoader extends stream.Transform {
...
+        ' AUTO-GENERATED BY JS-COMPILE: "webcomponents-loader.js" and ' +
```
rictic: JS-COMPILE is isn't super clear. How about:

` AUTO-GENERATED BY polymer-build: `

src/build/rewrite-webcomponents-loader.ts
```diff
@@ -49,25 +73,78 @@ export class UseES5WebcomponentsLoader extends stream.Transform {
...
+  <script type="text/javascript">
+    // This prevents custom-elements-es5-adapter.js from parsing & running
+    // on browsers without native support for ES2015+ & custom elements.
+    if (!window.customElements) {
+      var ceShimContainer = document.querySelector('#autogenerated-ce-es5-shim');
+      ceShimContainer.parentElement.removeChild(ceShimContainer);
+    }
+  </script>
+  <script type="text/javascript" src="${adapterScriptUrl}"></script>
+</div>
```
rictic: Can indent this out to the right for legibility

FredKSchott: +1, some indentation may be a better default than none at all. I had considered trying to detect indentation as well but it didn't seem worth the extra work/complexity

FredKSchott:
@rictic PTAL
@justinfagnani any comments before merging?

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 <!--## Unreleased-->
+- `build`: Update automatic `webcomponentsjs` polyfilling to move it and all affected elements following it into the body so that the `custom-elements-es5-adapter.js` can work properly in IE11. (See [#627](https://github.com/Polymer/polymer-cli/issues/627))
```
justinfagnani: We really need some docs on transformations like this. Can you describe this in the "build" section of the README for now, and maybe we'll find a new place later if that isn't good.

src/build/rewrite-webcomponents-loader.ts
```diff
@@ -12,16 +13,39 @@ const attrValueMatches = (attrName: string, regex: RegExp) => {
...
+const webcomponentsLoaderRegex = /\bwebcomponents\-(loader|lite)\.js\b/;
+const webcomponentsLoaderMatcher = p.AND(
+    p.hasTagName('script'), attrValueMatches('src', webcomponentsLoaderRegex));
+const headMatcher = p.hasTagName('head');
+const bodyMatcher = p.hasTagName('body');
+const scriptMatcher = p.AND(
     p.hasTagName('script'),
-    attrValueMatches('src', /\bwebcomponents-loader\.js$/));
+    p.OR(
+        webcomponentsLoaderMatcher,
```
justinfagnani: I don't think you need this in the predicate. You'll already match it, unless it had a weird type, in which case ¯\_(ツ)_/¯

src/build/rewrite-webcomponents-loader.ts
```diff
@@ -12,16 +13,39 @@ const attrValueMatches = (attrName: string, regex: RegExp) => {
...
 };
 
-const p = dom5.predicates;
-const scriptIncludeWebcomponentsLoader = p.AND(
+const webcomponentsLoaderRegex = /\bwebcomponents\-(loader|lite)\.js\b/;
+const webcomponentsLoaderMatcher = p.AND(
+    p.hasTagName('script'), attrValueMatches('src', webcomponentsLoaderRegex));
+const headMatcher = p.hasTagName('head');
+const bodyMatcher = p.hasTagName('body');
+const scriptMatcher = p.AND(
```
justinfagnani: we really should add this one to dom5, we use it so much

src/build/rewrite-webcomponents-loader.ts
```diff
@@ -12,16 +13,39 @@ const attrValueMatches = (attrName: string, regex: RegExp) => {
...
+ *
+ * NOTE(fks) 03-28-2017: Comments are not currently moved to the body, which
+ * means relevent comments may be left behind in the head. This is fine for now,
+ * but may break if we ever decide to support important metadata in comments
+ * (ex: support for defining lazy-imports via HTML comments).
  *
- * TODO(rictic): test this.
+ * TODO(fks) 03-28-2017: Move into polymer-build and add tests.
  */
 export class UseES5WebcomponentsLoader extends stream.Transform {
```
justinfagnani: change to custom elements adapter?

src/build/rewrite-webcomponents-loader.ts
```diff
@@ -49,25 +73,77 @@ export class UseES5WebcomponentsLoader extends stream.Transform {
...
+    // Otherwise we need to move the webcomponents-loader/webcomponents-lite
+    // loader down to the body so that the es5 adaperter script shim will work.
+    // Any following scripts, styles, and imports need to be moved down to the
+    // body as well to preserve ordering.
+    const orderSensitiveSiblings = dom5.queryAll(
+        script.parentNode!, p.OR(scriptMatcher, styleMatcher, linkMatcher));
+    const scriptIndex = orderSensitiveSiblings.indexOf(script);
+    const afterScriptSiblings = orderSensitiveSiblings.slice(scriptIndex);
+    const loaderMovedComment = dom5.constructors.comment(
+        ' AUTO-GENERATED BY polymer-build: "webcomponents-loader.js" and ' +
```
justinfagnani: polymer-build -> polymer-cli

src/build/rewrite-webcomponents-loader.ts
```diff
@@ -49,25 +73,77 @@ export class UseES5WebcomponentsLoader extends stream.Transform {
...
-
+    // Collect important dom references & create fragments for injection
     const correctedFile = file.clone();
+    const headElement = dom5.query(parsed, headMatcher)!;
+    const bodyElement = dom5.query(parsed, bodyMatcher)!;
+    const loaderScriptUrl = dom5.getAttribute(script, 'src')!;
+    const adapterScriptUrl =
+        url.resolve(loaderScriptUrl, 'custom-elements-es5-adapter.js');
+    const es5AdapterScript = parse5.parseFragment(`
+    <div id="autogenerated-ce-es5-shim">
```
justinfagnani: I'd move the comment below to above this node and describe the purpose of the whole snippet

CHANGELOG.md
FredKSchott: The README is verrrrry out of date, will add for now and then start a new issue to update it

FredKSchott:
@justinfagnani ptal

PR #639, opened 2017-03-30
Allow using .babelrc configuration files

vpalmisano:
Allow using a `.babelrc` configuration file found in the current build directory as option for the `JSBabelTransform` class.
The file should be a valid JSON object, e.g.:

    { 
      "presets": [
        [
          "babili", 
          {
            "replace": {
              "replacements": [
                  {
                    "identifierName": "__DEBUG__",
                    "replacement": {
                      "type": "booleanLiteral",
                      "value": false
                    }
                  }
              ]
            }
        }]
      ] 
    }

### DIFF
--- a/src/build/optimize-streams.ts
+++ b/src/build/optimize-streams.ts
@@ -18,7 +18,7 @@ import * as gulpif from 'gulp-if';
 import {minify as htmlMinify, Options as HTMLMinifierOptions} from 'html-minifier';
 import * as logging from 'plylog';
 import {Transform} from 'stream';
-import {accessSync, readFileSync} from 'fs';
+import * as fs from 'fs';
 
 const babelPresetES2015 = require('babel-preset-es2015');
 const babiliPreset = require('babel-preset-babili');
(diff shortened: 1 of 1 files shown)


justinfagnani:
Hi @vpalmisano,

Our stance so far has been to not allow configuration of the compilation step, because we include it only as a way to target older browsers, and don't want users to start writing code that relies on the compilation even for the modern browsers that support ES6.

What use cases do you have in mind here?

justinfagnani:
I see, you just added this to the minimize transform... interesting. We'll have to think on this a bit. I don't want expectations to get too high around customizing babel.

danners:
Hi,
@justinfagnani 
i also want this. My use case is that i used a ES6 function (https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) that i supposed to be transpiled down with babel for use in IE11, but apparently the used preset don't include that translation. I didn't verify if that can be fixed by a ,babelrc but i suspect this would open that possibility.

justinfagnani:
`.findIndex` is something that can be runtime patched into the Array prototype by loading the appropriate polyfill.

We really don't want to make the compilation configurable. Newer than ES6 features should be compiled to ES6 before reaching our compile step.

Issue #629, opened 2017-03-24
polymer init should not generates .travis.yml from github templates


peterlauri:
### Description
**polymer init polymer-2-starter-kit** generates a .travis.yml file that contains build for node_js 4, 5 and 6. Node 6 is defined as supported version, and running the tests on 4 and 5 doesn't make sense for polymer projects.

### Versions & Environment
- Polymer CLI: 0.18.0-pre.15
- node: v6.10.1
- Operating System: OSX

#### Steps to Reproduce

1. Install polymer-cli@next: 0.18.0-pre.15
2. Run **polymer init polymer-2-starter-kit**
3. cat .travis.yml

#### Expected Results
```
node_js:
  - '6'
```

#### Actual Results
```
node_js:
  - '6'
  - '5'
  - '4'
```

FredKSchott:
+1. Honestly I say we get rid of this file entirely when running polymer init. Right now it's coming through because it's in the PSK github repo but it was never meant to be a recommendation of any sort.

FredKSchott:
Renamed this issue, and filed the verison issue in PSK: https://github.com/PolymerElements/polymer-starter-kit/pull/993

FredKSchott:
@peterlauri thanks again for reporting. Lets keep discussion in the older #446

Issue #446, opened 2016-10-23
Remove .github folder on the Polymer Starter Kit


abdonrd:
We don't want the `.github` folder when the users use the `starter-kit` generator. :)

FredKSchott:
Haha yes, that would be strange.

Should we filter out all `.*` files by default? For starter-kit we have:
```
.github
.gitattributes
.gitignore	
.travis.yml
```

I think this would be a good rule for all projects, by default

abdonrd:
@FredKSchott but we want the `.gitignore`. And maybe the `.gitattributes` (?)

FredKSchott:
I can live with not providing a .gitignore for templates, since I may want to ignore different files as the `init` user. Otherwise I would be fine with whitelisting .gitignore and filtering out all others.

We'd like to open `init` up to use github repos as well as yeoman generators in the future, so I'd like to think of a solution for all projects and not just psk.

PR #647, opened 2017-04-05
Add .polymerignore mechanism

cco3:
Several templates for the init command are stored on github:
- polymer-1-starter-kit
- polymer-2-starter-kit
- shop

These projects have several files that are not meant to be used by
consumers of the template.  Examples:
- .travis.yml
- .github
- .gitattributes

The solution offered in this commit is provides a mechanism to read a
`.polymerignore` file that filters for the purpose of `polymer init`
templates.  This way, any new files that get added to the template do
not need to be registered in polymer-cli.  The `.polymerignore` format
is very much like the `.gitignore` format.
- Each line is a glob pattern (both * and ** are respected).
- Empty lines are ignored.
- Lines that begin with # are ignored.
- Lines that begin with / are considered to be at the root of the project.

Issues:
- #446
- #629

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "@types/parse5": "^2.2.33",
     "@types/request": "0.0.39",
     "@types/resolve": "0.0.4",
+    "@types/rimraf": "^0.0.28",
     "@types/semver": "^5.3.30",
     "@types/temp": "^0.8.28",
     "@types/uglify-js": "^2.6.28",
--- a/src/github/github.ts
+++ b/src/github/github.ts
@@ -18,6 +18,7 @@ import * as logging from 'plylog';
 import * as semver from 'semver';
 
 import request = require('request');
+import rimraf = require('rimraf');
 import GitHubApi = require('github');
 
 const gunzip = require('gunzip-maybe');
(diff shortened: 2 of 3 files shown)


FredKSchott:
@cco3 thanks for the contribution. Before asking all external templates to create and maintain seperate `.polymerignore` files, it would be worth considering what we could do ourselves in polymer-cli to help.

One thing we could do is maintain a blacklist of certain filenames to exclude. Off the bat this could include ('.travis.yml`, `.github`, &`.gitattributes`). This would solve our problem for now (since we currently own all github-based templates) and could then revisit this if we open up github repos for 3rd party templates.

@cco3 wdyt? Any interest in implementing?

cco3:
I'll implement whatever you'd like, but before we go that route, let me offer a few thoughts:
- I understand that it adds a certain complication, but maintaining these files does not seem especially burdensome.
- It's not clear to me that these files should never be included in a template (what about a template that exists to get users started with github?).
- It's not clear to me that no new files, including project-specific files, will arise that will also need to be ignored.

In the case of the latter, it would be awkward to update polymer-cli with the introduction of each file, because then the template projects could not be certain that users of the templates would not get the wrong files depending on what version of polymer-cli they are using.  If the definitions of which files to ignore exists in the projects themselves, there'd be no concern over polymer-cli versions.

Another alternative that keeps the ignore definitions with the template projects but doesn't require a .polymerignore file, the files that need to exist in the template could be defined in a template/ directory, and everything outside of that would be ignored (assuming the template directory exists).  That might be messier than this option here, but I figured I'd mention it.

FredKSchott:
@cco3 Those are all good points. One thing that's important to know about our templates is that all of them are powered by Yeoman generators which can define whatever behavior they'd like. So most of our templates (including all 3rd party templates) don't have this issue.

The only two templates that currently have this issue are Polymer Starter Kit & Shop, because they use a Yeoman Generator that clones a github repo at a specific version.

Because the issue is so limited in scope, I'd like to hold off on defining, adding, and documenting support for a public `.polymerignore` file format that only two templates have any use for right now. I'd much rather handle this all behavior internally in the CLI, and then discuss a public format again in the future if we decide to open up Github-repo-powered templates to 3rd-party developers who will have a much wider range of files to support.

PR #651, opened 2017-04-10
create UpperCamelCase class name from element name

### LINKED ISSUE #650: Polymer 2 element template does not update element title
_From @wmginsberg on April 5, 2017 20:55_

Hey! While testing out the 2.0 element template to get an easier start on creating an element, I noticed that the name of my element wasn't updating in my class decoration. Is this working as intended, or is it a bug? Example attached showing "MyElement" instead of "ScrabbleTile" 
screen shot 2017-04-05 at 1 53 43 pm
 

_Copied from original issue: Polymer/tools#66_

FredKSchott:
- Fixes #650

### DIFF
--- a/src/init/application/application.ts
+++ b/src/init/application/application.ts
@@ -75,6 +75,8 @@ export function createApplicationGenerator(templateName: string):
       ];
 
       this.props = await this.prompt(prompts);
+      this.props.elementClassName = this.props.elementName.replace(
+          /(^|\W)+(.)/g, (_match, _p0, p1) => p1.toUpperCase());
     }
 
     writing() {
--- a/src/init/application/application.ts
+++ b/src/init/application/application.ts
@@ -76,7 +76,7 @@ export function createApplicationGenerator(templateName: string):
 
       this.props = await this.prompt(prompts);
       this.props.elementClassName = this.props.elementName.replace(
-          /(^|\W)+(.)/g,
+          /(^|-)(\w)/g,
           (_match: string, _p0: string, p1: string) => p1.toUpperCase());
     }
 
--- a/src/init/application/application.ts
+++ b/src/init/application/application.ts
@@ -76,7 +76,8 @@ export function createApplicationGenerator(templateName: string):
 
       this.props = await this.prompt(prompts);
       this.props.elementClassName = this.props.elementName.replace(
-          /(^|\W)+(.)/g, (_match, _p0, p1) => p1.toUpperCase());
+          /(^|\W)+(.)/g,
+          (_match: string, _p0: string, p1: string) => p1.toUpperCase());
     }
 
     writing() {
--- a/src/init/element/element.ts
+++ b/src/init/element/element.ts
@@ -70,6 +70,8 @@ export function createElementGenerator(templateName: string):
       ];
 
       this.props = await this.prompt(prompts);
+      this.props.elementClassName = this.props.elementName.replace(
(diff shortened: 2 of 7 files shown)


aomarks:
:lgtm:

---

Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/init/application/application.ts, line 78 at r1](https://reviewable.io:443/reviews/polymer/polymer-cli/651#-KhPCJOFMy5EtMjflUZQ:-KhPCJOFMy5EtMjflUZR:bw3oghm) ([raw file](https://github.com/polymer/polymer-cli/blob/9cc561a4061a4db3d9100a101c6dcf551f480dd0/src/init/application/application.ts#L78)):*
> ```XML
> 
>       this.props = await this.prompt(prompts);
>       this.props.elementClassName = this.props.elementName.replace(
> ```

Might be nice to have this somewhere with a test... but only a minor copy-paste, so up to you.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/polymer/polymer-cli/651#-:-KhPCT4BJd4Q5PtTCpkK:bnfp4nl)*

src/init/application/application.ts
```diff
@@ -75,6 +75,8 @@ export function createApplicationGenerator(templateName: string):
       ];
 
       this.props = await this.prompt(prompts);
+      this.props.elementClassName = this.props.elementName.replace(
+          /(^|\W)+(.)/g, (_match, _p0, p1) => p1.toUpperCase());
```
justinfagnani: I think we should use the same matcher as core polymer: `replace(/-[a-z]/g , (m) => m[1].toUpperCase())`

src/init/element/element.ts
```diff
@@ -70,6 +70,8 @@ export function createElementGenerator(templateName: string):
       ];
 
       this.props = await this.prompt(prompts);
+      this.props.elementClassName = this.props.elementName.replace(
+          /(^|\W)+(.)/g, (_match, _p0, p1) => p1.toUpperCase());
```
justinfagnani: ditto

src/init/element/templates/polymer-2.x/_element.html
```diff
@@ -18,7 +18,7 @@
   </template>
 
   <script>
-    class MyElement extends Polymer.Element {
+    class <%= name %> extends Polymer.Element {
```
justinfagnani: Doesn't this need to be `<% elementClassName %>`

src/init/application/application.ts
FredKSchott: Class name is UpperCamelCase, that regex will convert to lowerCamelCase.

src/init/element/element.ts
FredKSchott: See response above

src/init/element/templates/polymer-2.x/_element.html
FredKSchott: doh, good catch.

FredKSchott:
@justinfagnani PTAL

src/init/application/application.ts
justinfagnani: I still might only uppercase only the first character and characters after a dash: `/(^|-)[a-z]/g`

FredKSchott: +1, always up for some regex golf. Fixed that up a bit and landed on: `/(^|-)([a-z])/g`

FredKSchott:
integration tests ftw! They just found a bug in the template that unit tests hadn't

PR #653, opened 2017-04-11
Deploy CLI support proposal.

aomarks:
Trying a design doc proposal in PR form.

 - [x] CHANGELOG.md does not need updating

### DIFF
--- a/design/deploy.md
+++ b/design/deploy.md
@@ -1,0 +1,98 @@
+# Deploy
+
+A proposal to add support for hosting service deployment to the Polymer CLI.
+
+## Goals
+- Provide new users with a fast way to get a project hosted online.
+- Provide advanced users with an initial bootstrap for more advanced deployments.
+- Demonstrate the [PRPL](https://www.polymer-project.org/2.0/toolbox/server) pattern where HTTP/2 Server Push is available.
+
+## Non-goals
+- Do not try to support every possible deployment environment and configuration, or even a lot of them. Just enough to get users started.
+
+## Design
+
+- Add an `--environment` flag.
+  - Every command knows that it's being run in the context of this environment.
+  - Under the `default` environment, all commands operate as they do today.
+  - Under other environments:
+     - `build` now outputs the boilerplate needed to deploy the project to a given service, in addition to the traditional app build output.
+     - `serve` runs an appropriate HTTP server that emulates the configured hosting environment, instead of polyserve.
+
+- Add an `environments` section to `polymer.json`.
+  - **name**: The name of this environment.
+  - **type**: The type of hosting environment: `app-engine`, `gh-pages`, etc.
+  - Additional type-specific configuration (examples below).
+  - We could also treat environments as a special case of the existing build concept, and include them in the `build` section.
+
+- Add a `deploy` command.
+  - Pushes to the hosting service for the current environment.
+  - Under the `default` environment, offers to help set up a deploy environment.
+     - Could have an `init`-style menu to choose from the available types.
+  - With the `--open` flag, also opens a browser (same as `serve`).
+  - Print the commands we're running so that users can do it themselves.
+
+## Environments
+
+### App Engine
+
+- Has a [free tier](https://cloud.google.com/free/).
+
+- Supports Node.js flex environment.
+
+- Custom `polymer.json` environment configuration:
+  - **project_id**: The App Engine project ID to deploy to.
+
+- `build`
+
+  - Runs both a bundled and unbundled standard build.
+
+  - `app.yaml`
+     - Mostly just specifies the `nodejs` runtime.
+
+  - `package.json`
+     - Note: this can't be the same as the existing *project* `package.json`.  This config defines a specific AppEngine deployment. It has different dependencies to the project, sets the deployed node version, etc.
+
+  - `server.js`
+     - An express HTTP server.
+     - Not polyserve, because it is a bit heavy with development-oriented features. Prefer something as simple as possible since we'd like users to understand it completely.
+     - Implementation is stamped out from a Yeoman template (as opposed to e.g. a binary dependency and generated config file) so that it's easy to customize later.
+     - Serves the main file from all URLs.
+     - Serves bundled or unbundled depending on client capabilities.
+     - Supports server push manifest.
+         - Could we factor out polyserve's push manifest support as an express middleware in a separate repo?
+     - Note: App Engine also supports a built-in static file server configurable directly in `app.yaml`. This is what the [Polymer App Toolbox tutorial](https://www.polymer-project.org/2.0/start/toolbox/deploy#deploy-with-appengine) demonstrates. It would support serving the main file from all URLs, but it not Server Push or User-Agent switching behavior. To show off PRPL we need to include a custom server.
+
+- `serve`
+  - Runs `build`, `npm install`, and `npm start`.
+     - App Engine will give us the Node.js runtime version configured in `package.json` `engines`, but it might not match what the user has installed locally, so behavior could theoretically differ. `npm` should warn if they mismatch, which is probably good enough. The `--engine-strict` flag would make this a hard error.
+     - Note: `dev_appserver.py` does not support the nodejs flex runtime. This wouldn't buy us anything anyway, since flex environments are docker containers with standard runtimes.
+
+- `deploy`
+  - Runs `build`, `npm install`, and `gcloud app deploy --project="$PROJECT_ID"`.
+  - The user will have had to have installed the [Cloud SDK](https://cloud.google.com/sdk/downloads) and run `gcloud init` to save local credentials. From that point on, the `gcloud` command authenticates automatically. We can provide these instructions on screen if `gcloud` is not found or fails.
+
+### GitHub Pages
+
+- Free up to [usage limits](https://help.github.com/articles/what-is-github-pages/#usage-limits).
+
+- Custom `polymer.json` environment configuration:
+  - **repository**: The git repository to push to.
+
+- `build`
+  - Runs a fully bundled standard build, since we don't have Server Push.
+
+- `serve`
+  - Runs some basic off-the-self static file server. It should not support serving the main file from all URLs, Server Push, or other advanced features that GitHub Pages doesn't support.
+
+- `deploy`
+  - Runs `build`, clones the configured repo to a temporary directory, replaces the `gh-pages` branch, and pushes.
+  - Could use [tschaub/gh-pages](https://github.com/tschaub/gh-pages) library.
+
+### Other enviroments
+
+In the future we might also want to support:
+
+- Firebase
+- Apache
+- Nginx


justinfagnani:
Thanks for testing out design-docs as PRs :)

design/deploy.md
```diff
@@ -0,0 +1,98 @@
...
+## Design
+
+- Add an `--environment` flag.
+  - Every command knows that it's being run in the context of this environment.
+  - Under the `default` environment, all commands operate as they do today.
+  - Under other environments:
+     - `build` now outputs the boilerplate needed to deploy the project to a given service, in addition to the traditional app build output.
+     - `serve` runs an appropriate HTTP server that emulates the configured hosting environment, instead of polyserve.
+
+- Add an `environments` section to `polymer.json`.
```
justinfagnani: You use plural here and singular above - which raises the question of whether or not there is only one, or multiple environments. This section seems to describe multiple user-named environments, but I had assumed that we would have environments that come with a name already. Can you clarify a little?

design/deploy.md
```diff
@@ -0,0 +1,98 @@
...
+
+- Add a `deploy` command.
+  - Pushes to the hosting service for the current environment.
+  - Under the `default` environment, offers to help set up a deploy environment.
+     - Could have an `init`-style menu to choose from the available types.
+  - With the `--open` flag, also opens a browser (same as `serve`).
+  - Print the commands we're running so that users can do it themselves.
+
+## Environments
+
```
justinfagnani: Like we allow users to install additional init templates by npm installing packages with the name `polymer-init-*`, we should allow users to install environments. Definitely not the first step though.

justinfagnani: We should have some rough sketch at least of the Environment API - right now we just have `serve()` and `build()` (link to the existing interface too).

For each environment hook we need to specify when it runs, with what arguments, and how it can effect the process it hooks (add warnings, indicate success/failure, modify options, parse flags, etc). We can leave many of those unspecified for now if they wouldn't be used by the two built-in environments we're talking about here. I just want to make sure we're thinking about how this feature evolves.

design/deploy.md
```diff
@@ -0,0 +1,98 @@
...
+### App Engine
+
+- Has a [free tier](https://cloud.google.com/free/).
```
justinfagnani: Let's add a brief description here, especially because we might be talking about a specific flavor of App Engine deployment where we provide the server - which might be a different environment from a more general one where the user bring their own cloud project (in which case we may want to define the point at which `polymer deploy` isn't a great fit and they should just use `gcloud`.)

design/deploy.md
```diff
@@ -0,0 +1,98 @@
...
+
+- Custom `polymer.json` environment configuration:
+  - **project_id**: The App Engine project ID to deploy to.
+
+- `build`
+
+  - Runs both a bundled and unbundled standard build.
+
+  - `app.yaml`
+     - Mostly just specifies the `nodejs` runtime.
```
justinfagnani: app.yaml also specifies the app/project name I believe. How does this info get in there? Do we ned to read it from a flag/config option?

design/deploy.md
```diff
@@ -0,0 +1,98 @@
...
+         - Could we factor out polyserve's push manifest support as an express middleware in a separate repo?
+     - Note: App Engine also supports a built-in static file server configurable directly in `app.yaml`. This is what the [Polymer App Toolbox tutorial](https://www.polymer-project.org/2.0/start/toolbox/deploy#deploy-with-appengine) demonstrates. It would support serving the main file from all URLs, but it not Server Push or User-Agent switching behavior. To show off PRPL we need to include a custom server.
+
+- `serve`
+  - Runs `build`, `npm install`, and `npm start`.
+     - App Engine will give us the Node.js runtime version configured in `package.json` `engines`, but it might not match what the user has installed locally, so behavior could theoretically differ. `npm` should warn if they mismatch, which is probably good enough. The `--engine-strict` flag would make this a hard error.
+     - Note: `dev_appserver.py` does not support the nodejs flex runtime. This wouldn't buy us anything anyway, since flex environments are docker containers with standard runtimes.
+
+- `deploy`
+  - Runs `build`, `npm install`, and `gcloud app deploy --project="$PROJECT_ID"`.
```
justinfagnani: It's interesting that the deploy command runs the build command. Like we talked about yesterday, this is a new dependency between commands. We should talk about this outside of the App Engine environment and show how an environment would invoke another command. We should also talk about whether it's ok for deploy to modify `build` output, or whether new artifacts need to happen in `build`, or if `deploy` gets it's own file system space to transform the build output.

design/deploy.md
```diff
@@ -0,0 +1,98 @@
...
+- Add a `deploy` command.
+  - Pushes to the hosting service for the current environment.
+  - Under the `default` environment, offers to help set up a deploy environment.
+     - Could have an `init`-style menu to choose from the available types.
+  - With the `--open` flag, also opens a browser (same as `serve`).
+  - Print the commands we're running so that users can do it themselves.
+
+## Environments
+
+### App Engine
```
justinfagnani: We might want to name this something like "PRPL Reference Server on App Engine"

design/deploy.md
```diff
@@ -0,0 +1,98 @@
...
+  - Under the `default` environment, all commands operate as they do today.
+  - Under other environments:
+     - `build` now outputs the boilerplate needed to deploy the project to a given service, in addition to the traditional app build output.
+     - `serve` runs an appropriate HTTP server that emulates the configured hosting environment, instead of polyserve.
+
+- Add an `environments` section to `polymer.json`.
+  - **name**: The name of this environment.
+  - **type**: The type of hosting environment: `app-engine`, `gh-pages`, etc.
+  - Additional type-specific configuration (examples below).
+  - We could also treat environments as a special case of the existing build concept, and include them in the `build` section.
```
justinfagnani: I think environments are likely a cross-cutting concern here, instead of part of a build. The PRPL/GAE environment probably wants to use a few builds - compiled and uncompiled - so conceptually is sits above the individual build configurations.

design/deploy.md
```diff
@@ -0,0 +1,98 @@
...
+
+  - `package.json`
+     - Note: this can't be the same as the existing *project* `package.json`.  This config defines a specific AppEngine deployment. It has different dependencies to the project, sets the deployed node version, etc.
+
+  - `server.js`
+     - An express HTTP server.
+     - Not polyserve, because it is a bit heavy with development-oriented features. Prefer something as simple as possible since we'd like users to understand it completely.
+     - Implementation is stamped out from a Yeoman template (as opposed to e.g. a binary dependency and generated config file) so that it's easy to customize later.
+     - Serves the main file from all URLs.
+     - Serves bundled or unbundled depending on client capabilities.
```
justinfagnani: We need to go into more detail on this one. How does the environment know which builds to associate with which UA strings? We can provide a default set of UA matching for features like ES6, native CE/SD/etc. Maybe users can match a feature set to a build name?

design/deploy.md
aomarks: It's no longer recommended to put the application ID in app.yaml, instead you use the `--project` flag to gcloud. We do still need to know it for deploy, though. So it will be read from polymer.json at that point.

https://cloud.google.com/appengine/docs/standard/python/config/appref#syntax

design/deploy.md
aomarks: There are two concepts. We should iterate on names.

The first concept is what I was calling an "environment". It's a specific user deployment configuration (e.g. my particular app engine deploy target). I imagined a polymer.json can have many of these defined in the "environments" section, similar to builds. Maybe your project can be deployed to both GitHub Pages and App Engine. Or maybe you have two App Engine targets, e.g. testing and prod. The --environment flag would pick from one of these configurations. We could also call this an "environment configuration"?

The second concept is what I was calling an "environment type". It's the higher level concept. Each environment would specify one of these types. This doc describes two such types so far: App Engine and GitHub Pages. We could also call this "environment" (and rename the first concept), "environment plugin", or "environment template"?

design/deploy.md
web-padawan: CLI definitely needs to provide a sample UA detection script. For `nginx` support this probably could be integrated using [nginScript](https://www.nginx.com/resources/wiki/nginScript/). If we had a reliable script, I would try to implement it.

design/deploy.md
usergenic: Yeah I agree, that a deployment would probably have a source or distribution folder that is meant to be copied and/or served as root.  The builds may produce those or in a future world with smart servers deployment would just declare that the repo root is that root.  I'd propose a `root` property here, but I don't know if that's as generally applicable as I am naively thinking.

usergenic: Clarification: I vote to keep builds and deployments orthogonal.

design/deploy.md
```diff
@@ -0,0 +1,98 @@
...
+- Do not try to support every possible deployment environment and configuration, or even a lot of them. Just enough to get users started.
+
+## Design
+
+- Add an `--environment` flag.
+  - Every command knows that it's being run in the context of this environment.
+  - Under the `default` environment, all commands operate as they do today.
+  - Under other environments:
+     - `build` now outputs the boilerplate needed to deploy the project to a given service, in addition to the traditional app build output.
+     - `serve` runs an appropriate HTTP server that emulates the configured hosting environment, instead of polyserve.
```
usergenic: I hope `polymer serve` isn't too intrinsically married to trafficking parameters over to `polyserve` to get its juice from an environment specification.  I really like the idea that `serve` could utilize different backends in this way like running a local GAE emulator or a local docker instance.

design/deploy.md
```diff
@@ -0,0 +1,98 @@
...
+## Non-goals
+- Do not try to support every possible deployment environment and configuration, or even a lot of them. Just enough to get users started.
+
+## Design
+
+- Add an `--environment` flag.
+  - Every command knows that it's being run in the context of this environment.
+  - Under the `default` environment, all commands operate as they do today.
+  - Under other environments:
+     - `build` now outputs the boilerplate needed to deploy the project to a given service, in addition to the traditional app build output.
```
usergenic: I'm not sure how `build` should be informed directly by an environment flag.  I understand how valuable it is for a build to produce artifacts like requisite nginx configs or whatever; Environment might have to include a build name to reference the specific output dir for this to work.

Incidentally, unless its already supported, `ProjectBuildOptions#name` property should support names with slashes so you can target deep paths...

design/deploy.md
```diff
@@ -0,0 +1,98 @@
...
+- Provide new users with a fast way to get a project hosted online.
+- Provide advanced users with an initial bootstrap for more advanced deployments.
+- Demonstrate the [PRPL](https://www.polymer-project.org/2.0/toolbox/server) pattern where HTTP/2 Server Push is available.
+
+## Non-goals
+- Do not try to support every possible deployment environment and configuration, or even a lot of them. Just enough to get users started.
+
+## Design
+
+- Add an `--environment` flag.
```
usergenic: Please note this issue https://github.com/Polymer/polymer-build/issues/146 where user talks about `environment` in the typical dev/stage/prod forms.  Worth keeping-in-mind so we don't overload the term in a way that might cause any confusions as to users' conventional use.

TimvdLippe:
Really cool design, looking forward to this shipping! This will probably also fix #808

FredKSchott:
Closing this PR as abandoned.

@aomarks please take a look and rebase/reopen if this is something you'd still like to get merged in.

PR #658, opened 2017-04-13
Fix failing polymer 2 application and element tests

mattpalermo:
The previous tests made the assumption that no classes would
be automagically added to the `<h2>` tags. This assumption
doesn't hold in firefox 52 and I suspect browsers where the
Shadow DOM polyfill is used.

The tests were changed to only test the innerHTML as opposed to
the outerHTML.

The CHANGELOG was not modified. An existing entry already
describes this change.


mattpalermo:
I signed it!

mattpalermo:
Thanks @FredKSchott. Would this be the culprit? https://github.com/Polymer/polymer-cli/blob/master/test/integration/integration_test.js#L60

PR #661, opened 2017-04-13
remove node v4 support

FredKSchott:
Because this is already being done downstream, we might as well announce it in Polymer CLI and make it explicit so that no one is surprised.

If we see demand from users for the latest tagged version of the CLI that still supports Node v4, we should branch from the commit right before polymer-build@v1.0.0 was updated and create a new npm `node4` release tag so that users can `npm install -g polymer-cli@node4`. But it's still tbd if the demand for this is there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 - `init`: Update polymer 2.0 application & element tests to improve and fix broken tests
 - `init`: Update polymer 1.x application & element template WCT dependency to `^6.0.0-prerelease.5`.
 - `init`: Update polymer application & element READMEs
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - `init`: Update polymer 2.0 application & element tests to improve and fix broken tests
 - `init`: Update polymer 1.x application & element template WCT dependency to `^6.0.0-prerelease.5`.
 - `init`: Update polymer application & element READMEs
+- [Breaking] Remove Node v4 support: Node v4 is no longer in Active LTS, so as per the [Polymer Tools Node.js Support Policy](https://www.polymer-project.org/2.0/docs/tools/node-support) polyserve will not support Node v4. Please update to Node v6 or later to continue using the latest verisons of Polymer tooling.
```
aomarks: s/polyserve/polymer-cli (or just polymer tools).

FredKSchott: whoops, thought that was a clean copy-paste

rictic:
see e.g. https://github.com/Polymer/polymer-linter/pull/76/files

FredKSchott:
@rictic 👍  oooh good point about updating node types & engines.

PR #662, opened 2017-04-13
update polyserve@v0.17.0

FredKSchott:
- Includes changes to support Chromium & Edge auto-compilation
- [X] Changelog updated


justinfagnani:
probably don't need 4 reviewers for this one :)

FredKSchott:
@justinfagnani but you are all equally qualified to review :)

In all seriousness, would you prefer to be /cc'd? Or not pinged at all?

justinfagnani:
I think we should treat the reviewers sections as a request for that person to review, ie. an AND

PR #681, opened 2017-04-19
Pass entrypoint from polymer.json to polyserve.

aomarks:
Fixes https://github.com/Polymer/polyserve/issues/180.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- `serve` now respects the `entrypoint` configured in `polymer.json`.
```
justinfagnani: I would add what practical effect this has for users. Does this make `polymer serve -o` do the expected thing? Does it make the custom-elements-es5-adapter go into the right file?

PR #686, opened 2017-04-21
remove ability for CLI to resolve to run a different version of the CLI

FredKSchott:
- Closing the loop on yesterdays CLI wild goose chase into npm by removing this functionality
- This behavior was undocumented and has no useful purpose in our modern CLI.
- Please comment if you are using this behavior for anything / believe it should continue to be supported.

/cc @keanulee

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - `serve` now respects the `entrypoint` configured in `polymer.json`.
+- Remove global command-line behavior to run a locally install version of the CLI if it existed in the current working directory. This unexpected behavior was never documented but some users could be running an incorrect version of the CLI as a result.
 
 ## v0.18.0 [04-13-2017]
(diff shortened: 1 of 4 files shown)


justinfagnani:
again, this doesn't need 4 reviewers

FredKSchott:
@justinfagnani when things calm down for 2.0 (or are they already calm?) I want to drive a discussion around review etiquette. I think we all have different expectations here and it's hard to track. 

For example: there's a valid need for "i just need a quick review from anyone on the team" PRs like this one where a single assignee can end up taking days to review if they're too busy. I don't mind being added as a reviewer in that case but I know you do.

Regardless, happy to not bother you with these in the meantime :)

jstclair:
Just a quick note - we are using this as a way to avoid installing the polymer cli globally on our build server(s). This change will block us from upgrading our use of polymer-cli on our build infrastructure.

Issue #689, opened 2017-04-25
some complex observers and computed properties are not evaluated anymore when minified


christophe-g:
With 0.18.1, complex observers and computed properties are not evaluated anymore when some parameters are `null` !

## Versions & Environment

Polymer CLI: 0.18.1
node: 6.10.1
Operating System: Linux 

## Test case : 

```html
<link rel="import" href="../../bower_components/polymer/polymer.html">
<dom-module id="test-cli-app">
  <template>
    <style>
    :host {
      display: block;
    }
    </style>
    <p>null: [[propNullComputed]]</p>
    <p>empty: [[propEmptyComputed]]</p>
    <p>complex: [[propComplexComputed]]</p>
    <p>observer: [[propObserved]]</p>
  </template>
  <script>
  (function() {
    'use strict';

    Polymer({

      is: 'test-cli-app',

      properties: {
      
        propObserved: {
          type: String
        },

        propNull: {
          type: String, 
          value: null
        },

        propEmpty: {
          type: String, 
          value: ''
        },
      
        propComplexComputed: {
          type: String,
          computed: '_computeProp(propNull, propEmpty)'
        },

        propNullComputed: {
          type: String, 
          computed: '_computeProp(propNull)'
        },

        propEmptyComputed: {
          type: String, 
          computed: '_computeProp(propEmpty)'
        },
      },

      observers: [
        '_observeProp(propNull, propEmpty)'
      ],

      _computeProp: function() {
        return ' I have been computed';
      },

      _observeProp: function() {
        this.propObserved = ' I have been observed';
      }

    });
  })();
  </script>
</dom-module>
```
### Expected result
```
null: I have been computed
empty: I have been computed
complex: I have been computed
observer: I have been observed
```

### Actual result
 The minified version will fail to evaluate `propComplexComputed` and `propObserved`, and only displays : 

```
null: I have been computed
empty: I have been computed
complex:
observer:
```
This issue is of course  causing some unexpected behaviors, like https://github.com/firebase/polymerfire/issues/197.

christophe-g:
The culprit seems to be the way the minifying vector (babel-preset-babili ?) transforms expression like `value === undefined` to `v == void 0`. If value is `null` it evaluates to `false` in the non-minified version **and `true` in the minified version**...  

When evaluating complex observers or computed properties `_marshalArgs`  method is called.
The minified version of this part :
```js
  if (bailoutEarly && v === undefined) {
          return;
        }
```
[...]

PR #698, opened 2017-04-26
Safe minify

christophe-g:
solves #689 and replaces #696


justinfagnani:
Thanks @christophe-g! It'd be great to have a regression test for this to make sure we don't break it in the future.

PR #707, opened 2017-04-30
Propagate `description` to `index.html` in application templates

jsilvermist:
Propagate `description` entered from command line during `polymer init` for either application template to the `<meta name="description">` tag in `index.html`.

If the user doesn't enter a description, fall back to previous template of:
```html
<meta name="description" content="<app-name> description">
```

I wasn't sure of the best way to format this, so will happily change if wanted.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- `init`: Propagate `description` from `init` to application templates in `index.html` meta tag.
+
 ## v0.18.1 [04-25-2017]
 
 - `init` small template fixes.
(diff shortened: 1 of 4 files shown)


jsilvermist:
@FredKSchott So I'm not exactly sure how to add the tests without a lot of redundancy.

I added tests covering `polymer-1.x`, but should I duplicate them for `polymer-2.x`? I tried using a `forEach` for each generator, but then I still have to call `done` after both are asynchronously complete.

What's the best practise in this type of scenario?

FredKSchott:
@jsilvermist duplicating similar code is fine here. Ideally we wouldn't have to, but having simple tests -- even if they repeat some code -- is better than creating and maintaining something clever and complex to fix the problem.

jsilvermist:
Duplicated tests for polymer-2.x, lmk if you want me to change anything!

FredKSchott:
Great, LGTM! Thanks for the contribution @jsilvermist!

Issue #699, opened 2017-04-26
[lint] warnings exit in error


stramel:
### Description

When `polymer lint` finds issues that result in warnings the command exits with code 1. 

### Versions & Environment

- Polymer CLI: 0.18.1
- node: 6.9.1
- Operating System: Windows 10

#### Steps to Reproduce

1. Clone PSK
2. npm i -g polymer-cli
3. npm i;polymer install
4. npm run lint

#### Expected Results

Warnings are displayed and command exits with code 0 if no errors are present.

#### Actual Results

Warnings are displayed and command exits with code 1

stramel:
I would be happy to add at least an initial go at this if someone can point me in a desired direction.

FredKSchott:
Thanks for filing this, someone on the team mentioned that this might be the case the other day.

Let me run through this today and take a look at what exactly is breaking. Hopefully it's a simple fix in PSK and not a bug in the linter/analyzer.

FredKSchott:
Just realized I misunderstood your issue, and thought that you were seeing a `1` exit code when there were no warnings or errors. As of right now this is expected behavior. 

Discussing a possible enhancement to change this behavior here: https://github.com/Polymer/polymer-project-config/pull/20

PR #709, opened 2017-05-02
Prevent Warnings from erroring

stramel:
Quick fix to prevent linter from erroring out when warnings and/or info messages are present. 

Ideally would be configurable and I would be happy to take a stab at it. I'm assuming it would be better to make the `minimumSeverity` configurable or add another property for controlling what severity causes failure?

Addresses #699 for the short-term.

### DIFF
--- a/src/lint/lint.ts
+++ b/src/lint/lint.ts
@@ -84,6 +84,6 @@ export async function lint(options: Options, config: ProjectConfig) {
       message += ` ${infos.length} ${chalk.green('info')} messages`;
     }
     console.log(`\n\nFound ${message}.`);
-    return new CommandResult(1);
+    return new CommandResult(errors.length ? 1 : 0);
   }
 }
--- a/src/lint/lint.ts
+++ b/src/lint/lint.ts
@@ -84,6 +84,7 @@ export async function lint(options: Options, config: ProjectConfig) {
       message += ` ${infos.length} ${chalk.green('info')} messages`;
     }
     console.log(`\n\nFound ${message}.`);
-    return new CommandResult(errors.length > 0 ? 1 : 0);
+    const shouldLintFail = errors.length > 0;
+    return new CommandResult(shouldLintFail ? 1 : 0);
   }
 }
--- a/src/lint/lint.ts
+++ b/src/lint/lint.ts
@@ -75,15 +75,15 @@ export async function lint(options: Options, config: ProjectConfig) {
     const warnings = filtered.filter((w) => w.severity === Severity.WARNING);
     const infos = filtered.filter((w) => w.severity === Severity.INFO);
     if (errors.length > 0) {
-      message += ` ${errors.length} ${chalk.red('errors')}`;
+      message += ` ${errors.length} ${chalk.red('error(s)')}`;
     }
     if (warnings.length > 0) {
-      message += ` ${warnings.length} ${chalk.yellow('warnings')}`;
+      message += ` ${warnings.length} ${chalk.yellow('warning(s)')}`;
     }
     if (infos.length > 0) {
       message += ` ${infos.length} ${chalk.green('info')} messages`;
     }
-    console.log(`\n\nFound ${message}.`);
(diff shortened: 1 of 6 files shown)


test/integration/lint_test.js
```diff
@@ -46,12 +46,21 @@ suite('polymer lint', function() {
     return runCommand(binPath, ['lint', '--rules=polymer-2-hybrid'], {cwd});
   });
 
+  test('succeeds when lint warnings are found', () => {
+    const cwd = path.join(fixturePath, 'lint-with-warning');
+    const result = runCommand(binPath, ['lint'], {cwd});
+    return result.then((output) => {
+      assert.include(
+          output, '<style> tags should not be direct children of <dom-module>');
```
FredKSchott: can you assert that the lint found only the 1 expected warning as well?

src/lint/lint.ts
```diff
@@ -84,6 +84,6 @@ export async function lint(options: Options, config: ProjectConfig) {
       message += ` ${infos.length} ${chalk.green('info')} messages`;
     }
     console.log(`\n\nFound ${message}.`);
-    return new CommandResult(1);
+    return new CommandResult(errors.length > 0 ? 1 : 0);
```
FredKSchott: Add a comment to explain this a bit more. Or, better, yet, move it into a descriptive variable name:

```js
const shouldCommandFail = errors.length > 0;
return new CommandResult(shouldLintFail ? 1 : 0);
```

stramel:
@FredKSchott It may be good to create a new issue or extend the referenced issue to include after this, setting the `minimumSeverity`  (logging) and setting a new property that controls minimum severity for failing.

I adjusted this PR with your feedback and this prevents it from failing on warnings.

test/integration/lint_test.js
```diff
@@ -46,12 +46,25 @@ suite('polymer lint', function() {
...
+      assert.include(
+          output, 'Found  1 warnings.');
```
FredKSchott: blah, while you're here can you fix:
- the double-space
- `warnings` -> `warning(s)`
- `errors` -> `error(s)`

stramel: lol yeah, I noticed those and thought it was odd :+1:

stramel:
@rictic PTAL :)

justinfagnani:
We talked about this in the team meeting, and I think that the point of a linter is to fail if there are warnings.

Our linter happens to categorize things into Errors and Warnings because some things are more severe, like actual broken syntax - but in many languages this would not be the job of the linter, it would be the job of the compiler, the linter is there for stylistic and other rules not fully suitable for the compiler, and not Errors in the sense that they would break compilation or evaluation.

Only failing on Errors would make this linter more of a static checker that only failed if it thought your code couldn't run.

FredKSchott:
I disagree with the absolute that the point of a linter is to fail if there are warnings. There's are lots of useful cases where signaling an issue without erroring is preferred. For example, with JS linters I'll usually create a warning for `console.log` methods: I don't want to leave log messages scattered across my app and hard to find, but I also don't want them to break my tests/linting (I may have added logging specifically for my tests). Highlighting something not-so-serious in your IDE without breaking tests/linting is another use case.

I think the real problem that @stramel has highlighted here is that the linter, and not the user, is responsible for deciding what is a warning and what is an error. What I'd really like to see is the ability for the user to configure that warning vs. error. I thought that this was a good step towards that, but thinking through it now it might not actually help us get there.

stramel:
I agree with @FredKSchott. There are some like non-declared variables that really shouldn't break the builds. The case I'm thinking of is here is from the PSK, link to build (https://travis-ci.org/PolymerElements/polymer-starter-kit/jobs/228186636).

I think the ultimate goal is that the user can control their logging level and the failure level. Currently, this is tied to `minimumSeverity`. 

I was thinking this would be a decent stop-gap until the user has been given control by failing less often. However, it's up to you guys and I'm good with your decision.

PR #717, opened 2017-05-05
Use a single polymerproject for build, add support for build presets

### LINKED ISSUE #714: Build Presets
Add build preset support so that common build configurations are easy to use without having to specify every individual build option.

## Benefits

- Easier to use from the command line for getting started, demos, documentation, npm scripts, etc.
- Encodes some commonly used, and recommended combinations of options
- Presets can be changed over time so that new versions of the CLI bring new optimizations that are automatically applied.

## Built-in Presets

- `es5-bundled`: all optimizations, compilation, bundling
- `es6-unbundled`: all optimizations, no compilation, bundling

## `--preset` Flag

`--preset={name}`

- Additional build flags override preset
- Can be specified multiple times for multiple presets
- Additional build flags are not allowed with multiple presets
- The name of preset is used as the name of the build

## `polymer.json`

```json
{
  "builds": [
    {
      "preset": "{name}"
    }
  ]
}
```

- Each build can have one preset
- Additional options override the preset
- The name of the build defaults to the name of the preset.

FredKSchott:
~~*Note: Tests will fail until https://github.com/Polymer/polymer-project-config/pull/17 is merged and released. Checkout locally and `npm link` until then.*~~

- Add support for build configuration "presets" (Design: #714)
- Improves build performance by using a single PolymerProject instance to read sources/deps

### DIFF
--- a/src/commands/build.ts
+++ b/src/commands/build.ts
@@ -163,20 +163,20 @@ export class BuildCommand implements Command {
     const hasCliArgumentsPassed =
         this.args.some((arg) => typeof options[arg.name] !== 'undefined');
     if (hasCliArgumentsPassed) {
-      return build(commandOptionsToBuildOptions(options), polymerProject);
+      await build(commandOptionsToBuildOptions(options), polymerProject);
+      return;
     }
 
     // If no build flags were passed but 1+ polymer.json build configuration(s)
     // exist, generate a build for each configuration found.
     if (config.builds) {
-      return Promise
-          .all(config.builds.map((buildOptions: ProjectBuildOptions) => {
-            return build(buildOptions, polymerProject);
-          }))
-          .then(() => undefined);
+      await Promise.all(config.builds.map((buildOptions) => {
+        return build(buildOptions, polymerProject);
+      }));
+      return;
     }
 
     // If no builds were defined, just generate a default build.
-    return build({}, polymerProject);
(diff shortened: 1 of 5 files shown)


src/commands/build.ts
```diff
@@ -116,50 +156,27 @@ export class BuildCommand implements Command {
...
     }
 
-    // If a single build was defined or configured via the project config,
-    // generate a build for that configuration.
-    if (config.builds.length === 1) {
-      return build(config.builds[0], config);
+    // If no build flags were passed but 1+ polymer.json build configuration(s)
+    // exist, generate a build for each configuration found.
+    if (config.builds) {
+      return Promise
```
justinfagnani:
```ts
await Promise.all(...);
return;
```

src/commands/build.ts
```diff
@@ -116,50 +156,27 @@ export class BuildCommand implements Command {
...
+          .all(config.builds.map((buildOptions: ProjectBuildOptions) => {
```
justinfagnani: I don't think you need the argument type, it should be inferred from the type of `config.builds`. You might get down to:

```ts
await Promise.all(config.builds.map((options) => build(options, polymerProject));
```

FredKSchott: yup, you're right. Good catch

PR #720, opened 2017-05-08
De-duplicate babel helpers

rictic:


FredKSchott:
+1 for pump, what was you experience with it? Thinking about encouraging it for polymer-build pipelines / using it in the CLI build pipeline.

gulpfile.js
```diff
@@ -72,21 +80,83 @@ gulp.task(
...
+  'typeof', // Symbol support, for IE11
+  // 'jsx', // we don't support JSX
+  // 'asyncIterator', // async-iterators are not in ES2015
+  // 'asyncGenerator', // async-iterators are not in ES2015
+  // 'asyncGeneratorDelegate', // async-iterators are not in ES2015
+  // 'asyncToGenerator', // async functions are not in ES2015
+  'classCallCheck',
+  'createClass',
+  'defineEnumerableProperties',
+  'defaults', // TODO (justinfagnani): what's this for?
```
FredKSchott: can you respond to all these TODOs? Most of them are just looking for a confirmation/answer, we should have that before merging.

gulpfile.js
```diff
@@ -72,21 +80,83 @@ gulp.task(
...
+  'temporalUndefined',
+  'toArray',
+  'toConsumableArray',
+];
+
+gulp.task('gen-babel-helpers', () => {
+  return Promise.resolve().then(() => {
+    return streamToPromise(pump(
+      run(
+      `babel-external-helpers --whitelist ${babelHelperWhitelist.join(',')}`,
```
FredKSchott: There's a JavaScript interface as well, use that instead so that you don't need all this stream-to-promise and rename logic. 
https://babeljs.io/docs/plugins/external-helpers/#detail-injecting-the-external-helpers

Also we can drop babel-cli for the much smaller babel-core.
http://npm.anvaka.com/#/view/2d/babel-cli

gulpfile.js
```diff
@@ -72,21 +80,83 @@ gulp.task(
...
+      { silent: true })
+      .exec(),
+      rename('babel-helpers.js'),
+      gulp.dest('./lib/build/')
+    ));
+  }).then(() => {
+    return streamToPromise(pump(
+      run('uglifyjs ./lib/build/babel-helpers.js -m', { silent: true }).exec(), rename('babel-helpers.min.js'),
```
FredKSchott: Same here, uglify has a javascript API as well. Use that instead. We could get rid of `gulp-run` and `gulp-rename` completely that way.

gulpfile.js
justinfagnani: some weird clang-ness here

src/build/inject-babel-helpers.ts
```diff
@@ -0,0 +1,55 @@
...
+  constructor(entrypoint: string) {
+    super({objectMode: true});
+    this.entrypoint = entrypoint;
+  }
+
+  async _transform(
+      file: File,
+      _encoding: string,
+      callback: (error?: any, file?: File) => void) {
+    if (file.path !== this.entrypoint) {
```
FredKSchott: Isn't it more correct / less error-prone to inject these helpers into all compiled files? Injecting into the entrypoint only will break statically-served and/or non-app-shell apps.

src/build/inject-babel-helpers.ts
```diff
@@ -0,0 +1,55 @@
...
+      callback(null, file);
+      return;
+    }
+    const contents = await getFileContents(file);
+    const document = parse5.parse(contents);
+
+    const babelHelpersFragment = parse5.parseFragment('<script></script>\n');
+    dom5.setTextContent(
+        babelHelpersFragment.childNodes![0]!,
+        fs.readFileSync(path.join(__dirname, 'babel-helpers.min.js'), 'utf-8'));
```
FredKSchott: Don't block like this unnecessarily. Also, don't read the file contents every time (this could be called multiple times if there are multiple builds).

src/build/inject-babel-helpers.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    const firstScript =
+        dom5.nodeWalk(document, dom5.predicates.hasTagName('script'));
```
justinfagnani: I think this really should be first script or html import. It's a weird case: compiled script, but no polyfills, but it's possible.

src/build/inject-babel-helpers.ts
justinfagnani: move this read out of the `_transform()` path, up to module initialization, preferably to an async read: `const babelHelpers = await _getBabelHelpers();`

src/build/optimize-streams.ts
```diff
@@ -113,7 +113,10 @@ class JSBabelTransform extends GenericOptimizeTransform {
  */
 export class JSDefaultCompileTransform extends JSBabelTransform {
   constructor() {
-    super({presets: [babelPresetES2015NoModules]});
+    super({
+      presets: [babelPresetES2015NoModules],
+      plugins: require('babel-plugin-external-helpers'),
```
justinfagnani: I'd move to require out of the class

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- produce much smaller output when compiling many JS files to ES5 by inserting
```
justinfagnani: Capitalize the "p"

src/build/inject-babel-helpers.ts
```diff
@@ -0,0 +1,63 @@
...
+ * that they don't need to be included with each compiled file.
+ */
+export class BabelHelpersInjector extends AsyncTransformStream<File, File> {
+  constructor(private entrypoint: string) {
+    super({objectMode: true});
+  }
+
+  protected async * _transformIter(files: AsyncIterable<File>) {
+    for
+      await(const file of files) {
```
rictic: clang-format misformats `for await` pretty bad :/

I was able to keep the two uses of it very short here to compensate.

rictic:
What's a good test project to check for savings in? This only matters for projects that use a lot of ES6.

justinfagnani:
@rictic the 2.0-preview-class branch of Shop should be good

src/build/inject-babel-helpers.ts
```diff
@@ -0,0 +1,63 @@
...
+      return file;
+    }
+    const contents = await getFileContents(file);
+    const document = parse5.parse(contents);
+
+    const babelHelpersFragment =
+        parse5.parseFragment('\n\n<script></script>\n\n');
+    dom5.setTextContent(
+        babelHelpersFragment.childNodes![1]!,
+        fs.readFileSync(path.join(__dirname, 'babel-helpers.min.js'), 'utf-8'));
```
justinfagnani: but now you can async :)

src/build/streams.ts
```diff
@@ -46,3 +76,153 @@ export function pipeStreams(streams: PipeStream[]): NodeJS.ReadableStream {
...
+      this.blockedOn = undefined;
+    } else {
+      const deferred = new Deferred<void>();
+      this.backlog.push({value, deferred});
+      await deferred.promise;
+    }
+  }
+
+  /**
+   * Iterate over values in the queue. Not intended for multiple readers.
```
justinfagnani: Not sure the part about multiple readers is needed. Iterators hold state and are never intended for multiple readers. Normal use is that readers receive Iterables, which produce a new Iterator for each reader.

src/build/streams.ts
```diff
@@ -12,6 +12,36 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
+import {Deferred} from 'polymer-analyzer/lib/utils';
+import * as stream from 'stream';
+
+import File = require('vinyl');
+
+(Symbol as any).asyncIterator = Symbol.asyncIterator || Symbol('asyncIterator');
```
justinfagnani: why not define `Symbol.asyncIterator` if it doesn't exist?

src/build/streams.ts
```diff
@@ -46,3 +76,153 @@ export function pipeStreams(streams: PipeStream[]): NodeJS.ReadableStream {
...
+   *
+   * Passes input into this._inputs.
+   */
+  protected _transform(
+      input: In,
+      _encoding: string,
+      callback: (error?: any, value?: Out) => void) {
+    this._initializeOnce();
+    this._inputs.write(input).then(() => {
+      callback();
```
justinfagnani: I think you need to call the given callback instance for each file passed in, or the buffer will fill and the streams will halt. I think you can push `[input, callback]` onto the queue, then transform the queue to just inputs before passing it to `this._transformIter`

src/build/inject-babel-helpers.ts
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * When compiling to ES5 we need to inject Babel's helpers into a global so
+ * that they don't need to be included with each compiled file.
+ */
+export class BabelHelpersInjector extends AsyncTransformStream<File, File> {
+  constructor(private entrypoint: string) {
+    super({objectMode: true});
+  }
+
+  protected async * _transformIter(files: AsyncIterable<File>) {
```
justinfagnani: I don't quite see why this interface is preferable over something like `async _transformAsync(file: File): Promise<File>`

The base class can implement `_transform` as:

```ts
protected async _transform(
    input: In,
    _encoding: string,
    callback: (error?: any, value?: Out) => void) {
  try {
    callback(null, await this._transformAsync(file));
  } catch (e) {
    callback(e);
  }
}
```

src/build/streams.ts
justinfagnani: nevermind, you're always calling callback, so buffering won't be a problem

src/build/inject-babel-helpers.ts
rictic: oh, erp, done.

src/build/streams.ts
rictic: That's what this does?

src/build/streams.ts
rictic: Yeah, this is an iterable which has the listed constraints when multiple iterators are generated.

src/build/inject-babel-helpers.ts
rictic: This is true, but this strategy also allows doing more than just mapping over a stream. I hope to move our other transform streams over to use this abstraction.

Issue #687, opened 2017-04-24
Polymer Install shorthand like NPM and Bower


stramel:
It would be fantastic if the polymer cli could support the a shorthand of `polymer i` for `polymer install`.

FredKSchott:
+1, this would be nice to have!

If you (or anyone else) is interested in implementing I'd be available to review

stramel:
@FredKSchott I wouldn't mind adding it but I wasn't sure if there was something to already handle aliases or not. Was thinking about accepting a name in the constructor that would be able to provide a command name? Definitely open to any direction you can give.

FredKSchott:
I would add an `alias: string[]` property to each command, and then have `polymer-cli.ts` add commands by both name and alias. 

Finer points:
- The help command should automatically list aliases in the help output for any one command (not necessary for the general `$ polymer help` output).
- `polymer-cli.ts` should error if there's a duplicate name/alias conflict, so that if we ever mess up the tests won't pass.

stramel:
Done with #721

PR #721, opened 2017-05-09
New: Add aliases for commands

stramel:
Adds aliases to commands to accomplish things like `polymer i` for `polymer install`. Apologies if there is somethings wrong with the TypeScript. Haven't used it before.

### Example Implementation

```
$ polymer install -h

polymer install

  installs Bower dependencies, optionally installing "variants" 

Alias(es)

  i 

Command Options

  --variants    Whether to install variants 
  --offline     Don't hit the network       

Global Options

  --env type                      The environment to use to specialize certain commands, like build     
  --entrypoint                    The main HTML file that will be requested for all routes.             
  --shell string                  The app shell HTML import                                             
  --fragment string[]             HTML imports that are loaded on-demand.                               
  --root string                   The root directory of your project. Defaults to the current working   
                                  directory                                                             
  --sources string[]              Glob(s) that match your project source files. Defaults to `src/**/*`. 
  --extra-dependencies string[]   Glob(s) that match any additional dependencies not caught by the      
                                  analyzer to include with your build.                                  
  -v, --verbose                   turn on debugging output                                              
  -h, --help                      print out helpful usage information                                   
  -q, --quiet                     silence output                                                        
```

### Aliases

- `i` - install (npm supports `isntall` as an alias, not sure if that would be something we would want)

### Not Currently Implemented

In parens just random shots at what could be aliases

- analyze (`a`?)
- build (`b`?)
- help ( `h`?)
- init (`n`?)
- lint (`l`?)
- test (`t`?)
- serve (`s`? `start`?)

Addresses #687 

@FredKSchott @rictic 

### Tasks 

 - [x] Tests

Taking suggestions for Changelog entry...

### DIFF
--- a/src/commands/help.ts
+++ b/src/commands/help.ts
@@ -96,7 +96,7 @@ export class HelpCommand implements Command {
       {header: 'Global Options', optionList: globalArguments},
     ];
 
-    if (command.aliases.length) {
+    if (command.aliases.length > 0) {
       usageGroups.splice(1, 0, {header: 'Alias(es)', content: command.aliases});
     }
 
--- a/src/commands/help.ts
+++ b/src/commands/help.ts
@@ -47,7 +47,7 @@ ${b('\\  /\\')}  ${m('/\\  /')}   ${b('/\\  /')}
 
 export class HelpCommand implements Command {
   name = 'help';
-  aliases = ['h'];
+  aliases = [];
 
   description = 'Shows this help message, or help for a specific command';
 
--- a/src/commands/help.ts
+++ b/src/commands/help.ts
@@ -47,6 +47,7 @@ ${b('\\  /\\')}  ${m('/\\  /')}   ${b('/\\  /')}
 
 export class HelpCommand implements Command {
   name = 'help';
+  aliases = ['h'];
 
   description = 'Shows this help message, or help for a specific command';
 
@@ -70,7 +71,7 @@ export class HelpCommand implements Command {
       },
       {
         header: 'Available Commands',
-        content: Array.from(this.commands.values()).map((command) => {
+        content: Array.from(new Set(this.commands.values())).map((command) => {
           return {name: command.name, summary: command.description};
         }),
       },
@@ -94,6 +95,11 @@ export class HelpCommand implements Command {
       {header: 'Command Options', optionList: command.args},
       {header: 'Global Options', optionList: globalArguments},
     ];
+
+    if (command.aliases.length) {
+      usageGroups.splice(1, 0, {header: 'Alias(es)', content: command.aliases});
+    }
+
     return commandLineUsage(usageGroups.concat(extraUsageGroups));
   }
 
--- a/src/commands/lint.ts
+++ b/src/commands/lint.ts
@@ -32,8 +32,6 @@ export interface Options {
 }
 
 export class LintCommand implements Command {
-  // TODO(rictic): rename to 'lint' here and elsewhere, delete
-  // legacy-lint.ts. Also update the README.
   name = 'lint';
   aliases = [];
 
--- a/src/commands/lint.ts
+++ b/src/commands/lint.ts
@@ -35,6 +35,7 @@ export class LintCommand implements Command {
   // TODO(rictic): rename to 'lint' here and elsewhere, delete
   // legacy-lint.ts. Also update the README.
   name = 'lint';
+  aliases = [];
 
   description = 'Identifies potential errors in your code.';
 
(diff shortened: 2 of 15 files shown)


src/commands/lint.ts
```diff
@@ -35,6 +35,7 @@ export class LintCommand implements Command {
   // TODO(rictic): rename to 'lint' here and elsewhere, delete
```
stramel: @rictic Has this been done?

justinfagnani:
It'd be good to add the list of aliases implemented to the PR description.

Given that this only adds `i` and `h`, how much value is there here? Do other commands even have obvious aliases?

`install` is a bit of an odd command in that I don't expect it to be used as often as a normal `bower install` - it's really there for variants. Eventually I hope the command can be deprecated.

src/commands/lint.ts
rictic: oh, yeah. that comment is a holdover from when the new linter was `experimental-lint`

stramel:
I have added the list of aliases implemented to the PR description as well as the ones that were not and quick thoughts for aliases. 

I was under the impression this command, `install`, would be here to stay for future hybrid states but if not, then I'm not sure how helpful this PR is... I was primarily wanting the `polymer i` command but implementing a full aliases feature made more sense than just trying to make it a one-off.

jsilvermist:
I like the idea of a `polymer s -o` honestly.

src/commands/help.ts
```diff
@@ -94,6 +95,11 @@ export class HelpCommand implements Command {
       {header: 'Command Options', optionList: command.args},
       {header: 'Global Options', optionList: globalArguments},
     ];
+
+    if (command.aliases.length) {
```
FredKSchott: be explicit here: `command.aliases.length > 0`

src/commands/lint.ts
FredKSchott: @stramel this is safe to remove in this PR if you want, up to you

src/commands/help.ts
```diff
@@ -94,6 +95,11 @@ export class HelpCommand implements Command {
...
+      usageGroups.splice(1, 0, {header: 'Alias(es)', content: command.aliases});
```
FredKSchott: hard to tell from the PR, can you add to the PR description a sample of what this now looks like?

jsilvermist:
One more to add to the list of unimplemented potential aliases for future reference:
`serve` -> `start`

stramel:
@FredKSchott I adjusted the code based on your comments but I'm a little unsure what all i need to test

src/commands/help.ts
```diff
@@ -47,6 +47,7 @@ ${b('\\  /\\')}  ${m('/\\  /')}   ${b('/\\  /')}
 
 export class HelpCommand implements Command {
   name = 'help';
+  aliases = ['h'];
```
FredKSchott: lets actually remove this for now, the original request was just for an `install` alias. I'd rather tackle all others together when we decide on a rule for which aliases to support overall.

FredKSchott:
Check out the current `tests/unit/commands/`. You should create a new `tests/unit/commands/install_test.js` and add at least one test for both `polymer install` & `polymer i`.

That should actually be fine for now. I forgot how closely those command unit tests mirrored the actual command line syntax. Integration tests wouldn't give us much more in this case.

stramel:
WRT Windows tests, Some of these tests that are in there are pretty fragile for Windows. Mostly referring to `test/unit/init/init_test.js` where `isPlatformWin` really applies to `mingw` but not `powershell`. Also, there is a test that has the conditional wrapping it for windows but the counterpart is not wrapped so on windows running `mingw` that test fails. There is another that is in need of the wrapping conditional and counterpart but has neither.

test/unit/install/install_test.js
```diff
@@ -10,12 +10,60 @@
...
+const sinon = require('sinon');
 const assert = require('chai').assert;
+const ProjectConfig = require('polymer-project-config').ProjectConfig;
+const PolymerCli = require('../../../lib/polymer-cli').PolymerCli;
 
 const polymerInstall = require('../../../lib/install/install');
 
 suite('install', () => {
 
+  suite('runs command', () => {
```
FredKSchott: This is testing the install command, which is different from the install logic run by the command.

Move these to a new, separate file in `tests/unit/commands/install_test.js`. Look at `commands/help_test.js` as an example.

stramel: Oh, I had based my tests off of `commands/help_test.js` but put them in the wrong spot I guess lol Moved them to `commands/install_test.js`

stramel:
@FredKSchott Green Build PTAL :)

stramel:
@FredKSchott @rictic Ping :)

stramel:
:tada: Wooo!

PR #723, opened 2017-05-09
New: Control of minimumSeverity

### LINKED ISSUE #699: [lint] warnings exit in error

stramel:

I really think it would be good to add a way to differentiate erroring and logging.

Took a shot at allowing control over minimumSeverity. Will replace #709 to fix #699.

@FredKSchott @rictic PTAL :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- **Updated `lint` Command**: `polymer lint` now has finer control over what is output and causes the command to fail.
 - Update polyserve to 0.19.0 which adds HTTP compression and JS compilation for Mobile Safari and Vivaldi browsers.
 
 ## v0.18.1 [04-25-2017]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,10 @@
 # Changelog
 
-## Unreleased
+<!-- ## Unreleased -->
 
-- **Updated `lint` Command**: `polymer lint` now has finer control over what is output and causes the command to fail.
+<!-- Add new, unreleased changes here. -->
+
+- **Updated `lint` Command**: `polymer lint` now provides a way to control what is output and causes the command to fail. To customize this you can either include the CLI flag `--min-severity` (options include `error`, `warning`, and `info`) or custom polymer.json lint configuration using the `minSeverity` property. (Defaults to `warning`)
 - Update polyserve to 0.19.0 which adds HTTP compression and JS compilation for Mobile Safari and Vivaldi browsers.
 
 ## v0.18.1 [04-25-2017]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,10 @@
 # Changelog
 
-<!-- ## Unreleased -->
-
-<!-- Add new, unreleased changes here. -->
+## Unreleased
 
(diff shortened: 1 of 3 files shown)


src/lint/lint.ts
```diff
@@ -45,10 +45,9 @@ export async function lint(options: Options, config: ProjectConfig) {
   const rules = lintLib.registry.getRules(ruleCodes || lintOptions.rules);
   const filter = new WarningFilter({
     warningCodesToIgnore: new Set(lintOptions.ignoreWarnings || []),
-    minimumSeverity: Severity.WARNING
+    minimumSeverity: options.minSeverity || lintOptions.minSeverity || Severity.WARNING,
```
stramel: Would still need to add `minSeverity` to the `polymer-project-config` `lint` section.

Side Note: wasn't sure if you do trailing commas in this repo or not.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- **Updated `lint` Command**: `polymer lint` now has finer control over what is output and causes the command to fail.
```
rictic: Please be more explicit here. Mention the new flag and how to use it.

src/lint/lint.ts
rictic: This doesn't compile because project config doesn't have minSeverity. Options:

* don't include the reference
* reference it using a string literal `lintOptions['minSeverity']`

In either case we should have a TODO here to add it to project config

rictic: Hm, warning severity should really be a string enum rather than the numeric enum I used -.-

I guess there's still time for a breaking change there if we're very quick.

In any case, for now, you'll need to convert the string to the enum value, something like:

```typescript
function severityStringToValue(maybeSeverity: string): Severity|undefined {
  maybeSeverity = maybeSeverity.trim().toLowerCase();
  switch (maybeSeverity) {
    case 'error': return Severity.ERROR;
    case 'warning': return Severity.WARNING;
    case 'info': return Severity.INFO;
    default: return undefined;
  }
}
```

Then check for undefined and throw a useful error in before constructing the WarningFilter.

stramel: Up to you if you would like to switch it or I can take a stab at it.

Also, it seems the analyzer has an inverse of this Severity to String. Maybe it would make sense to expose those in the Warning model so this doesn't get duplicated somewhere else? Just a thought.

I will add this for now.

stramel:
There is a bit of conversation going on in https://github.com/Polymer/polymer-project-config/pull/20 in case anyone wants to follow.

FredKSchott:
Lets keep the discussion in https://github.com/Polymer/polymer-project-config/pull/20 since it's already started there. Closing this PR for now, to be re-opened based on what we decide in https://github.com/Polymer/polymer-project-config/pull/20

PR #727, opened 2017-05-11
fixes the cli preset flag to work properly

FredKSchott:
- [X] Changelog updated
- Test added
- Did some small refactoring to go from 1 build integration test to N integration tests


src/commands/build.ts
```diff
@@ -135,6 +105,22 @@ export class BuildCommand implements Command {
...
   ];
 
+  /**
+   * Converts command-line build arguments to the `ProjectBuildOptions` format
+   * that our build understands, applying the preset if one was given.
+   */
+  private commandOptionsToBuildOptions(options: CommandOptions):
+      ProjectBuildOptions {
+    const buildOptions: ProjectBuildOptions = {};
+    const validBuildOptions = this.args.map(({name}) => name);
```
justinfagnani:
```ts
const validBuildOptions = new Set(this.args.map(({name}) => name));
```

src/commands/build.ts
```diff
@@ -135,6 +105,22 @@ export class BuildCommand implements Command {
...
+    for (const buildOption in options) {
```
justinfagnani:
```ts
for (const buildOption in Object.keys(options)) {
```

and remove the hasOwnProperty check below

src/commands/build.ts
```diff
@@ -24,40 +24,10 @@ import {PolymerProject} from 'polymer-build';
...
-
 const logger = logging.getLogger('cli.command.build');
 
-/**
- * Converts command-line build arguments to the `ProjectBuildOptions` format
- * that our build understands, applying the preset if one was given.
- */
-function commandOptionsToBuildOptions(options: CommandOptions):
-    ProjectBuildOptions {
-  return applyBuildPreset({
```
justinfagnani: where was the actual bug here?

FredKSchott: By defining all options in the argument, `applyBuildPreset()` interpreted them all as overrides on the preset.

Issue #728, opened 2017-05-11
AsyncIterator throws error.


hvdb:
### Description

Error:
```
TypeError: Cannot assign to read only property 'asyncIterator' of function 'function Symbol() { [native code] }'
    at Object.<anonymous> (**/node_modules/polymer-cli/lib/build/streams.js:44:22)
```

The transpiled code looks like this:
```
Symbol.asyncIterator = Symbol.asyncIterator || Symbol('asyncIterator');
```

in 0.18.1 everything works (duh)

Commit causing the issue: 
https://github.com/Polymer/polymer-cli/commit/83c800e6493893fa75653febec741049194bca89

More information:

Running polymer-cli via commandline does not give this error.
Running it programatically does.

```
 const polymer_cli = new polymerCLI.PolymerCli(['build'], json);
    return polymer_cli.run(); 
```

### Versions & Environment
- Polymer CLI: 0.18.2
- node: 6.10.2 (or 7.10.2)
- Operating System: macOS

#### Steps to Reproduce
run polymer build.

#### Expected Results
No problems build available.

#### Actual Results
error, see above.

PR #729, opened 2017-05-11
fix(build): Check if asyncIterator is already assigned.

hvdb:
Check if the asynIterator is already defined, if not define it.
This fixes my issue in the case that it is already defined somewhere apparently.

closes: #728

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 - Fix the CLI preset flag.
+- Fix `build` as streams where not working anymore. 
 
 ## v0.18.2 [05-10-2017]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 - Fix the CLI preset flag.
-- Fix `build` as streams where not working anymore. 
+- Fix an issue where compiling JS would crash in versions of node with native async iterators.
 
 ## v0.18.2 [05-10-2017]
 
--- a/src/build/streams.ts
+++ b/src/build/streams.ts
@@ -17,7 +17,9 @@ import * as stream from 'stream';
 
 import File = require('vinyl');
 
-(Symbol as any).asyncIterator = Symbol.asyncIterator || Symbol('asyncIterator');
+if (typeof (Symbol as any).asyncIterator  === 'undefined') {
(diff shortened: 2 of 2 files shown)


justinfagnani:
TypeScript actually doesn't attempt to define `Symbol.asyncIterator`, this is left up to the program. See here: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-3.html

You can see in the Travis CI failure that we need to define `Symbol.asyncIterator`. It seems like your environment has another place that's defining it, probably with `Object.defineProperty` and `writable: false` as are the other built-in symbols.

The real fix would be to check if it's defined before defining it with `Object.defineProperty`.

hvdb:
I will have a look at the real fix.

but i am quite sure that we don't have it anywhere else as i just executed polymer cli.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 - Fix the CLI preset flag.
+- Fix `build` as streams where not working anymore.
```
rictic:
```
- Fix an issue where compiling JS would crash in versions of node with native async iterators.
```

src/build/streams.ts
```diff
@@ -17,7 +17,9 @@ import * as stream from 'stream';
 
 import File = require('vinyl');
 
-(Symbol as any).asyncIterator = Symbol.asyncIterator || Symbol('asyncIterator');
+if (typeof (Symbol as any).asyncIterator  === 'undefined') {
```
rictic: Prefer:

```typescript
if ((Symbol as any).asyncIterator !== 'undefined') {
  (Symbol as any).asyncIterator = Symbol('asyncIterator');
}
```

`undefined` isn't assignable in modern JS environments, doing the typeof trick is overly paranoid.

hvdb: Well i also think it is a bit overly paranoid, but found it in a place in the polymer-cli so thought it was the way to go.

will change it.

src/build/streams.ts
```diff
@@ -17,7 +17,9 @@ import * as stream from 'stream';
 
 import File = require('vinyl');
 
-(Symbol as any).asyncIterator = Symbol.asyncIterator || Symbol('asyncIterator');
+if ((Symbol as any).asyncIterator !== 'undefined') {
```
rictic: This isn't right, it's string undefined rather than the value undefined, I'll send over a PR

hvdb: Should haven seen that, sorry

PR #741, opened 2017-05-16
Bump polymer-build version to 1.3.0

aomarks:
This is failing in integration test. It looks like the test fragment file foo.html is appearing in the output, where it used to be omitted. Expected change?


justinfagnani:
Removing `foo.html` from the output was something that changed in the test last week. This sounds like a regression. @usergenic

usergenic:
Looking into this now.

PR #764, opened 2017-05-22
Update copyright year

jesselpalmer:


FredKSchott:
Thanks for contributing @jesselpalmer. Our current copyright strategy is to set the Copyright when the file is created, and (I think) update it only when you edit the file (if needed). So the copyright dates as-is are currently correct.

We can document this somewhere if it comes up again to avoid any future confusion.

jesselpalmer:
Ah, thanks @FredKSchott!

FredKSchott:
np, thanks again for taking the time & contributing!

jesselpalmer:
👍

justinfagnani:
Just to clarify for posterity - the policy is to _not_ update copyrights, even when edited. The copyright starts applying from when a file was created. Updating it falsely implies that copyright started later than it did.

jesselpalmer:
Makes perfect sense. Thanks, @justinfagnani!

PR #766, opened 2017-05-23
Don't set service worker or push manifest prefixes when basePath is set.

### LINKED ISSUE #761: Switch to relative URLs in generated service worker, simplify handling of `basePath`
Historically, the service workers generated by `polymer-build` have used absolute URLs from the server root to specify which resources to cache.

We propose to change this going forward, using relative URLs instead. Moving to relative URLs will make it considerably easier to support two emerging use cases:

* Multi-build deployments with differential serving based on user-agent
* Single-build deployments from directories other than the server root, a use case explicitly supported by the Polymer Starter Kit

This change should be transparent to most users; since all of our templates and standard file layouts place the service worker next to the app's entrypoint (`index.html`), the caching behavior should be identical. If necessary, we can introduce a new config option to handle edge cases -- an option to specify absolute URLs, or alternatively an option to locate the service worker elsewhere in the directory structure.

Making this change will simplify the implementation of the recently introduced `basePath` config option; currently, this option not only uses a `<base>` tag to set the entrypoint's base URL, but also rewrites paths in both the push manifest and the generated service worker. By moving to relative paths instead, we can drop the path rewriting.

aomarks:
Fixes #761 (see there for context).

Will fail on Travis until https://github.com/Polymer/polymer-build/pull/230 is merged.

Also updates integration test and CHANGELOG to account for https://github.com/Polymer/polymer-build/pull/229.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 - Updated dependency on latest polymer-project-config so that bundled presets include prefetch link generation.
+- `build` Entrypoints will now be properly cached by generated service workers, and assets will be fetched by service workers using relative URLs to support apps mounted at non-root host paths.
 <!-- Add new, unreleased items here. -->
 
 ## v1.0.2 [05-19-2017]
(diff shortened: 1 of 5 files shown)


aomarks:
Thanks Fred!

PR #782, opened 2017-06-02
Bump shop init template to 2.x and drop all 1.x templates.

aomarks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 <!-- Add new, unreleased items here. -->
 - Updated lint rule to `polymer-2` in the `polymer-2-element` template.
+- Drop 1.x init templates. Bump `shop` init template to latest version.
 
 ## v1.1.0 [05-23-2017]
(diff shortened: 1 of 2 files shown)


aomarks:
Yeah, good question. I'm also not exactly sure what the hybrid vs 2.0 status is across the init templates.

+ @graynorton @justinfagnani thoughts?

aomarks:
Confirmed offline with @justinfaganani last week that we're cool with dropping the 1.0 templates.

jsilvermist:
@aomarks Should the names keep the full `polymer-2` prefix? Is there any reason not to simplify this?

For example `element`, `application` and `starter-kit`?

aomarks:
@jsilvermist The `polymer` prefix does make it clear that you're getting boilerplate for a Polymer element (it's entirely possible we could have non-Polymer webcomponent boilerplate at some point). The `-2` prefix probably could go away, although it does help make it obvious given that a lot of people are still on Polymer 1.x that you'll be getting a 2.0 element, and not a 1.0 or hybrid.

PR #816, opened 2017-07-06
Use webcomponents-loader.js in the init element

abdonrd:
The Polymer v2.x init element should use the `webcomponents-loader.js`, right?

### DIFF
--- a/src/init/application/templates/polymer-2.x/test/_element/_element_test.html
+++ b/src/init/application/templates/polymer-2.x/test/_element/_element_test.html
@@ -6,7 +6,7 @@
 
     <title><%= elementName %> test</title>
 
-    <script src="../../../webcomponentsjs/webcomponents-lite.js"></script>
+    <script src="../../../webcomponentsjs/webcomponents-loader.js"></script>
     <script src="../../../web-component-tester/browser.js"></script>
 
     <link rel="import" href="../../src/<%= elementName %>/<%= elementName %>.html">
--- a/src/init/element/templates/polymer-2.x/demo/index.html
+++ b/src/init/element/templates/polymer-2.x/demo/index.html
@@ -6,7 +6,7 @@
 
     <title><%= name %> demo</title>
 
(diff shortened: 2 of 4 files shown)


FredKSchott:
If it works on all browsers & serve options, sure!

I'm trying to remember the exact support matrix for webcomponents-loader.js, does it only break now when elements are compiled and the shim isn't included by polymer-build? If so, polymer-serve auto-compiling might cause problems...

abdonrd:
Mmm okey!

Now I already try in the last versions of Chrome, Safari and Firefox using the last Polymer CLI version (`polymer serve --compile auto`). And it works.

FredKSchott:
Hmm, were you able to try in IE? That's the tricky one I think...

@usergenic or @justinfagnani may know for sure, again I haven't worked on this directly in a while 😅

JohnRiv:
It works in IE11... can this PR be merged? Seems like it'd be better to train new Polymer developers to use the webcomponents-loader.js instead of webcomponents-lite.js so they're only sending down the polyfills they need.

JohnRiv:
The /test/ files work in IE11 too if we switch to webcomponents-loader.js... so let's update it in all places

jsilvermist:
`polymer-cli/src/init/element/templates/polymer-2.x/test/_element_test.html` also has a link pointing to `lite` not `loader`.

abdonrd:
Rebased and updated! :)

justinfagnani:
@abdonrd looks good, but can you fix html files in /src/init/application/templates/polymer-2.x/ too?

I see -lite used here: https://github.com/Polymer/polymer-cli/blob/master/src/init/application/templates/polymer-2.x/test/_element/_element_test.html#L9

abdonrd:
Updated! Thanks @justinfagnani!

abdonrd:
🚀

JohnRiv:
Thanks folks!

PR #823, opened 2017-07-13
Update 2.0 element test to use ES6

stramel:


stramel:
@TimvdLippe The 2.0 polymer code uses es6 arrows, class, getters/setters etc. Just seems logical, also would make my life easier :P

PR #824, opened 2017-07-15
New: Check node version based on package.json

stramel:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Changelog
 
+<!-- Add new, unreleased items here. -->
 ## Unreleased
 - Added `--fix` option to `polymer lint`. When passed, some warnings with simple mechanical solutions will be fixed.
   - Also supports warnings which can be addressed with less-safe changes via
@@ -8,7 +9,7 @@
 - Added `--watch` (shorthand `-w`) option to `polymer lint`. When passed, we will watch the filesystem for changes and rerun the linter immediately afterwards.
   - Also works with `--fix` to automatically fix and report warnings as you work!
 - `build` Added a CLI argument for setting the `basePath` option: `--base-path`.
-<!-- Add new, unreleased items here. -->
+- Derives node version check from the package.json.
 
 ## v1.5.7 [10-11-2017]
 - Updated css-slam, bower and other dependencies.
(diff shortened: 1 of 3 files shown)


src/polymer-cli.ts
```diff
@@ -12,6 +12,9 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
+// Bail out if invalid node version
+require('please-upgrade-node')(require('../package.json'));
```
justinfagnani: We already have similar logic here: https://github.com/Polymer/polymer-cli/blob/master/bin/polymer.js#L27

Also, I'd be hesitant to put statements before import declarations, as that's not actually spec-compliant.

stramel: Doh! I definitely didn't look here.

I put this in the wrong place apparently. The require before import is to ensure that older versions that don't support `import`. By placing it in bin/polymer.js it is technically before all the other imports. 

I will modify this to do the bin/polymer.js

bin/polymer.js
```diff
@@ -22,11 +22,12 @@ process.title = 'polymer';
...
 // NOTE 04-21-2017: Confirmed "semver" supports Node versions as low as 0.10
 var semver = require('semver');
+var version = require('../package.json').engines.node;
 
 // Exit early if the user's node version is too low.
-if (!semver.satisfies(process.version, '>=6')) {
+if (!semver.satisfies(process.version, version)) {
   console.log(
-      'Polymer CLI requires at least Node v6. ' +
+      'Polymer CLI requires at least Node v' + verion.replace(/[^\d\.]*/, '') + '. '
```
abdonrd: Typo: `verion` > `version`.

justinfagnani: Missing `+` at the end of the line.

justinfagnani: Can you store the `version.replace` result in a variable and comment what it's doing?

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Drives node version check from the package.json
```
TimvdLippe: Derives 😉

stramel: Updated

stramel:
@justinfagnani @FredKSchott Following up on this, I have resolved the changelog conflicts

stramel:
Thanks @justinfagnani looks like the build failed on `lint` integration test. Not sure if this is a flake or a change someone else introduced.

stramel:
Any chance we can get this merged? I just rebased again. I also regenerated the package-lock.json since I had conflicts in there and none seemed to be any breaking changes.

justinfagnani:
@stramel merged!

PR #826, opened 2017-07-19
Upgrade polymer build to 2.0

usergenic:
- Upgraded to Polymer Build ^2.0.0 which uses Polymer Bundler ^3.0.0.
- When no specific option is set for Bundler's `rewriteUrlsInTemplates` the CLI attempts to get the version of Polymer for the project using `bower`.  When Polymer 2.x is discovered, `rewriteUrlsInTemplates` is defaulted to `false`.  In case of Polymer 1.x or where version can not be identified, it defaults to `true`.  Any user settings override these defaults.


justinfagnani:
cc/ @katejeffreys this needs docs. I'll file a docs issue.

PR #844, opened 2017-08-10
Added npm flag to the passed options.

usergenic:
- This flag passing is obnoxiously untestable.  I just tried several angles to verify it gets passed to polyserve in the server options, but I'd have to refactor the method altogether to test the flag passing and right now I just want to patch this.


usergenic:
Thanks for understanding. :)

PR #847, opened 2017-08-15
Upgrade babili preset to new package name

TimvdLippe:
Installing the polymer-cli showed the following message:
```
babili has been renamed to babel-minify. Please update to babel-preset-minifybabel-preset-minify
```

This PR upgrades `babel-preset-babili` to the new name and fix breaking changes from 0.1.0. The unsafe option was removed and I updated the options to reflect the equivalent behavior. One change that was apparent was the removal of parenthesis from arrow functions (see the changed test). I am not sure if this is a bug or change in behavior of babili.

### DIFF
--- a/src/build/optimize-streams.ts
+++ b/src/build/optimize-streams.ts
@@ -132,11 +132,7 @@ export class JSDefaultMinifyTransform extends JSBabelTransform {
   constructor() {
     super({
       presets:
-          [minifyPreset(null, {
-            flipComparisons: true,
-            guards: true,
-            typeConstructors: true
-          })]
+          [minifyPreset(null, {simplifyComparisons: false})]
     });
   }
 }
--- a/src/build/optimize-streams.ts
+++ b/src/build/optimize-streams.ts
@@ -21,7 +21,7 @@ import {Transform} from 'stream';
 
 
 const babelPresetES2015 = require('babel-preset-es2015');
-const babiliPreset = require('babel-preset-babili');
+const minifyPreset = require('babel-preset-minify');
 const babelPresetES2015NoModules =
     babelPresetES2015.buildPreset({}, {modules: false});
 const externalHelpersPlugin = require('babel-plugin-external-helpers');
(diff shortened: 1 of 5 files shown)


TimvdLippe:
Ping @FredKSchott This should make sure that minification works with ES6 and Safari. I hit that bug by using `<script type="module">` in production.

justinfagnani:
@TimvdLippe can you rebase?

TimvdLippe:
@justinfagnani Done :+1:

src/build/optimize-streams.ts
```diff
@@ -132,7 +132,11 @@ export class JSDefaultMinifyTransform extends JSBabelTransform {
   constructor() {
     super({
       presets:
-          [babiliPreset(null, {'unsafe': {'simplifyComparisons': false}})]
+          [minifyPreset(null, {
+            flipComparisons: true,
+            guards: true,
+            typeConstructors: true
```
usergenic: @TimvdLippe are these settings equivalent to the simplifyComparisons setting or are there any meaningful differences we should note in CHANGELOG etc?

TimvdLippe: Hm, maybe a plugin has to be included for that now: https://github.com/babel/minify/tree/master/packages/babel-plugin-transform-simplify-comparison-operators, but given that the plugin was not supposed to run, not including it has the same meaning? Not 100% on that.

usergenic: /derp I just noticed your description mentioning the paren-removal change identified by tests.

TimvdLippe:
@usergenic done!

src/build/optimize-streams.ts
usergenic: It looks like `simplifyComparisons` was hoisted out of `unsafe` into top level of options.  I'm able to get the tests to pass with the following...

```typescript
   constructor() {
     super({
       presets:
          [minifyPreset(null, {simplifyComparisons: false})]
     });
   }
```

Was that missed or do these options matter in other ways not covered:
```
            flipComparisons: true,
            guards: true,
            typeConstructors: true
```

TimvdLippe: Hm that is really weird, I couldn't find it in the options. If that fixes it, then yes please! Feel free to commit to this branch :smile:

usergenic: ok will push

jogibear9988:
as this change does not fix the problem with the broken js, could #878 be merged?

PR #870, opened 2017-08-31
Disable bower networking

nikitindv:
By default bower list command attempts to fetch latest package versions from github, this step is not required for polymer build and causes it hang because polymer is not forwarding bower's github api token request.

### DIFF
--- a/src/build/build.ts
+++ b/src/build/build.ts
@@ -70,7 +70,7 @@ export async function build(
   async function getPolymerVersion(): Promise<string> {
     return new Promise<string>(
         (resolve, _reject) =>
-            bower.commands.list()
+            bower.commands.list({offline: true})
                 .on('end',
                     (result: any) => {
                       if (result && result.dependencies &&


TimvdLippe:
This would require updating the typings of Bower to reflect the settings object: https://travis-ci.org/Polymer/polymer-cli/jobs/270339274#L538

FredKSchott:
@nikitindv can you fix the typing issue @TimvdLippe mentioned above? If you can, ping this PR again and I'll reopen it for review.

```diff
// custom_typings/bower.d.ts
- list(): NodeJS.EventEmitter;
+ list({offline?: boolean}): NodeJS.EventEmitter;
```

I think this only works for you because you've called it before and it's cached the results offline. I don't think it will pass on a clean Travis install.

justinfagnani:
This actually looks ok to me, since if the user hasn't run Bower install, they won't have a Polymer version to check.

Issue #875, opened 2017-09-05
Typings should be moved from "dependencies" to "devDependencies"


iainreid820:
A large number of typings are installed unnecessarily when installing Polymer CLI, these should be declared as "devDependencies" rather than "dependencies".

justinfagnani:
Which ones? We install typings along with the dependency that they type. Are any misplaced?

iainreid820:
As per the commit, all of the "@types" dependencies should sit in the "devDependencies" as they aren't required at runtime by a normal user.

iainreid820:
Could the PR related to this please be merged? This is unfortunately preventing us from using the CLI because of conflicting types with Request and our own types.

justinfagnani:
The general guidance on types is to install them to `dependencies` or `devDependencies` based on where the package is they describe is installed. If you're using the CLI as a top-level package, it shouldn't matter where types are installed, and if you're using the CLI as a library (which I'd be curious about), then you want the `dependency` types with dependencies. How are you getting conflicts?

PR #876, opened 2017-09-05
Move typings to "devDependencies" (#875)

iainreid820:
The majority of the typings used for TS development and transpilation
currently sit in the "dependencies" block when they should really be
declared in the "devDependencies" block. This commit resolves this
issue.


FredKSchott:
Thanks for the contribution @iainreid820. 

Lets follow up in https://github.com/Polymer/polymer-cli/issues/875 first before considering any PRs.

Issue #701, opened 2017-04-27
js minify breaks polymerfire


christophe-g:
### Description
building with `  "js": { "minify": true }` breaks polymerfire. Cannot read the database anymore and get error messages like: `FIREBASE INTERNAL ERROR: Server Error: ClientId[7309775]:ErrorId[966]: Error on incoming message `.

Just replacing the files in `build/min/bower_components/firebase/` by original files (firebase-app.js, firebase-database.js, firebas-auth.js ... which are already minified) makes the app work again. 

As an - unsuccessful - workaround, I tried to upgrade  babel-preset-babili to the latest version (0.0.12), and unset all babel-preset-options (`{boolean: false, buildIn, false, ...}` as in https://github.com/Polymer/polymer-cli/pull/698). 

Is there a way to prevent some files to be minified during the build process?

### Versions & Environment
- Polymer CLI: 0.18.1 
- node: 6.10.1
- Operating System: Linux

christophe-g:
Additional info: 
- the `FIREBASE INTERNAL ERROR` only appears if the user is logged-in to firebase.
- Only `bower_components/firebase/firebase-database.js` is not behaving as expected in the minified version 

C

christophe-g:
workaround (ugly!): add smth that breaks babili at the top of a file that should not be minified

For example adding the snippet below at the top of firebase-database.js makes my minified app work again (the build will complain with `warn:    .js: Unable to optimize /home/christophe/Programming/Polymer/preignition/bower_components/firebase/firebase-database.js 
{ err: 'unknown: Invalid regular expression: /^([A-Za-z0-9_-.])+@$/: Range out of order in character class' }
`) and this file will not be minified.

```js
(function(){
new RegExp("^([A-Za-z0-9_\-\.])+\@$");
})
```

Interested in better ways to do this of course ...
C.

FredKSchott:
Ugh, this is probably related to https://github.com/Polymer/polymer-cli/issues/689.

We've talked about adding a configurable "skip optimization" option, but atm our effort may be better spent on fixing the minify bug in babili for everyone.

FredKSchott:
Looks like we have a fix here: https://github.com/Polymer/polymer-cli/pull/698

christophe-g:
@FredKSchott that would be too easy, The fix  #698 does not work here - something else seem to be happening. I did not have time to trace down this issue and find the root cause.

Good luck !
C.

gdevacc12:
[...]

Issue #633, opened 2017-03-27
--js-compile --js-minify --html-minify make jQuery crash


baocang:
### Versions & Environment

- Polymer: 2.0.0-rc.3
- Polymer CLI: 0.18.0-pre.15
- node: v7.7.3
- webcomponentsjs: 1.0.0-rc.7
- Operating System: macOS Sierra Version 10.12.3

#### Steps to Reproduce

1. Create an application project: `polymer init polymer-2-application` 
2. Add script tag to src/app-shell/app-shell.html `<script src="jquery-3.2.0.js"></script>`
3. Update webcomponentsjs in bower.json to `#1.0.0-rc.7`
4. Update bower components: `bower update`
5. Build: `polymer build --entrypoint index.html --shell src/app-shell/app-shell.html --bundle --js-compile --js-minify --html-minify`
6. Serve: `polymer serve build/default`
7. Open in chrome: [http://127.0.0.1:8000](http://127.0.0.1:8000)

#### Expected Results

No error is throw

#### Actual Results

Error is thrown

<img width="765" alt="error" src="https://cloud.githubusercontent.com/assets/16043697/24356577/28e671e2-132d-11e7-8bc6-b44a5e84a20e.png">

baocang:
![image](https://cloud.githubusercontent.com/assets/16043697/24357117/2c2a6eec-132f-11e7-9603-49cf6bf9e995.png)

web-padawan:
Might be related to https://github.com/babel/babel/issues/340

BTW, why to use jQuery with Polymer? You [might not need it](http://youmightnotneedjquery.com/) at all, and take into account that in case of DOM manipulations, jQuery is not aware of Shadow DOM.

baocang:
I'm trying to build a SaaS application wtihout jQuery, But some thirdpart library not sure. such as: http://www.highcharts.com, https://www.mobiscroll.com 😪.

FredKSchott:
@baocang Can you try each of those flags individually, to see which is causing the issue? My guess is it's just `--js-compile` but it could also be `--js-minify`.

We have a task to add the ability to whitelist files to be skipped for processing. If some popular files are currently broken we should bump up the priority.

baocang:
@FredKSchott --bundle --js-minify.

baocang:
@FredKSchott  `polymer build --js-minify`  with no error in console. but the code \\x20 has already translated to unicode char.

FredKSchott:
[...]

Issue #623, opened 2017-03-22
Exceeding memory limit when running polymer lint on a large project


web-padawan:
### Description
When running `polymer lint --rules polymer-1` from my project root I see the following error:

```
<--- Last few GCs --->

   93118 ms: Mark-sweep 1365.7 (1407.3) -> 1365.6 (1411.3) MB, 1365.1 / 0.0 ms [allocation failure] [GC in old space requested].
   94569 ms: Mark-sweep 1365.6 (1411.3) -> 1365.6 (1411.3) MB, 1450.8 / 0.0 ms [allocation failure] [GC in old space requested].
   95967 ms: Mark-sweep 1365.6 (1411.3) -> 1374.8 (1407.3) MB, 1397.2 / 0.0 ms [last resort gc].
   97297 ms: Mark-sweep 1374.8 (1407.3) -> 1383.9 (1407.3) MB, 1329.0 / 0.0 ms [last resort gc].


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0x2eec362cfb51 <JS Object>
    2: applyScanners(aka applyScanners) [/Users/sergej/.nvm/versions/node/v6.9.3/lib/node_modules/polymer-cli/node_modules/polymer-analyzer/lib/javascript/javascript-document.js:44] [pc=0x636680398c2] (this=0x2eec36204381 <undefined>,callbackName=0x24b212b89ac1 <String[23]: enterFunctionExpression>,node=0x3398b20afdb9 <a Node with map 0x1e7570d2341>,parent=0x3398b20afd11 <a Node with map 0x1e75...

FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
 1: node::Abort() [/Users/sergej/.nvm/versions/node/v6.9.3/bin/node]
 2: node::FatalException(v8::Isolate*, v8::Local<v8::Value>, v8::Local<v8::Message>) [/Users/sergej/.nvm/versions/node/v6.9.3/bin/node]
 3: v8::internal::V8::FatalProcessOutOfMemory(char const*, bool) [/Users/sergej/.nvm/versions/node/v6.9.3/bin/node]
 4: v8::internal::Factory::NewFixedArray(int, v8::internal::PretenureFlag) [/Users/sergej/.nvm/versions/node/v6.9.3/bin/node]
[...]

PR #878, opened 2017-09-08
js-ignore option

christophe-g:
Adding a way to skip files in js minification process. 

Should help addressing issues like #701. 

Example (will skip all `*.min.js` files and `firebase-database.js`): 

```js
  "builds": [{
    "bundle": true,
    "js": {
        "minify": true ,
        "ignore": [
            "*.min.js",
            "firebase-database.js"
      ]}
  }],
```


jsilvermist:
Would a regex like that be maintainable for larger projects compared to something like an array which is used for other comparable project settings in polymer-cli? eg:
```json
"js": {
  "minify": true,
  "ignore": [
    "*.min.js",
    "firebase-database.js"
  ]
}
```
Also, is this something that should be set and repeated in each build or set globally for all builds?

christophe-g:
> Would a regex like that be maintainable for larger projects compared to something like an array which is used for other comparable project settings in polymer-cli?

I find regex to be more practical here - but open to change this to an Array if it gets more upvotes. 

>   is this something that should be set and repeated in each build or set globally for all builds?
Merge state

It is a per build config.

mgibas:
I really like wildcards/regex, specially for things like `*.min.js`.

eskan:
+1

christophe-g:
PR updated based on @jsilvermist suggestion

JoelCode:
In the mean time, my solution around it is to do an extra step to copy/overwrite from the source, to the build directory after the build process is completed.

christophe-g:
@usergenic - thanks for 1.5.5 release. 

Any chance this one gets reviews for the next one ?

jogibear9988:
ref: https://github.com/Polymer/polymer-cli/issues/886

jogibear9988:
can this be merged? as it seems to be the only usable fix for #886 and #633

web-padawan:
@jogibear9988 I was wondering if this should ever be fixed on the polymer-cli side? Wouldn't it be better to file corresponding upstream issues to `babili`?

Anyway, I must say I don't use CLI for app projects due to the lack of flexibility and extending its configuration options looks good to me.

jogibear9988:
don‘t know in which part of the cli this happens.

if it‘s a babili error, then it should be fixed there! maybe it‘s already fixed there? (don‘t know if polymer cli uses newest version) (newest name is babel minify)

web-padawan:
The original issue should be revised once #847 gets merged. /cc @TimvdLippe

christophe-g:
@usergenic - any chance to get this reviewed ? thanks

jogibear9988:
I've tested after the merge of #847, JQuery for example is still broken when minimized via polymer-cli

jogibear9988:
@usergenic @FredKSchott can this be reviewed? this is a really needed fix. without that many con not minify

jogibear9988:
@justinfagnani

christophe-g:
@usergenic @FredKSchott @justinfagnani 

Any chance to get this reviewed ? 
- without it, my app will not minify (it is not just Polymerfire, try to run a minify build against webassembly code) 
- 500% performance gain when excluding all already minified resources (`js-ignore: ["*.min.js"]`).

jogibear9988:
The Team was triggered to look at that issue, see: https://github.com/Polymer/polymer/issues/4931

jogibear9988:
@christophe-g can you rebase so it compiles again?
@TimvdLippe ping

christophe-g:
@jogibear9988  - sure, done.

jogibear9988:
@christophe-g  i think most of the issues why this is needed will be fixed in next version of babel minify, but I think this should be merged... @TimvdLippe said here https://github.com/Polymer/polymer/issues/4931 we should ping this PR if we get no response from the team ....

don't know how to ping...

christophe-g:
@jogibear9988 - thanks!

I am very much in doubt next version of babel minify will solve all issues I currently have building my app, without the need of excluding some of the sources. 

For instance, some modules are compiled versions of C/C++ libraries (http://kripken.github.io/emscripten-site) that make polymer-cli crash with a fatal error, similar to https://github.com/Polymer/polymer-cli/issues/623.

lpellegr:
@jogibear9988 ping
@TimvdLippe ping

lpellegr:
I encountered the same issue as @christophe-g . Building _polymer-cli_ from this PR and using the new JS ignore option was a solution to make building work with both already minified and non-already minified JS files.

hjr265:
@usergenic Any update regarding this? It would be nice to have this merged or something similar implemented in the CLI.

usergenic:
Hey everyone, sorry for the delay on this.  Reviewing this today.  Seems like a useful if not necessary feature.

jogibear9988:
We should also check to switch to babel minify nightly, because it has many bugs fixed

usergenic:
Okay, reviewing this, there are some test failures here and the proposed changes technically require an update to `polymer-project-config` which expresses the configuration file format.  I added this issue: https://github.com/Polymer/polymer-project-config/issues/50 and will work on getting in a PR for that today as well as a new PR for the CLI to operate on it.

Thanks for your patience, everyone.

usergenic:
Reimplemented as https://github.com/Polymer/polymer-cli/pull/952.  Note exclude paths are not glob-enabled but rather act as prefixes so you can exclude folders etc.

usergenic:
On second though, I have added back matcher for wildcards.  Now *.min.js works.

usergenic:
Also note that #952 does not have a command-line switch as part of the update.  Use polymer.json build config for that or we can file subsequent PR for an override flag.

jogibear9988:
Can you also look to switch to babel minify nightly, cause many of the issues because this is needed are fix in the nightly version. And it does not seem that they'll create a new release!

christophe-g:
closing as implemented by #952. @usergenic - thanks for lifting up the original idea!

PR #889, opened 2017-09-26
Exit with explicit status 0 instead of unspecified

fwielstra:
When a node process is just interrupted via a SIGINT, it will report an undefined exit code. [The FAQ of Concurrently](https://www.npmjs.com/package/concurrently#faq) explicitly states that in such a case, it will exit with a non-null status, causing npm's runner to exit with an error. This change fixes that issue.

(We have a use case where we use `concurrently` to start both polymer and a mock server, neither of which returned a zero exit code on exit, causing the error)

### DIFF
--- a/bin/polymer.js
+++ b/bin/polymer.js
@@ -35,3 +35,7 @@ if (!semver.satisfies(process.version, '>=6')) {
 
 // Ok, safe to load ES2015.
 require('../lib/run');
+
+process.on('SIGINT', function() {
+  process.exit(0);
+});


FredKSchott:
Thanks for the PR @fwielstra! I understand your use-case, but this breaks normal program expectations. A SIGINT is an unexpected interruption, and a program shouldn't exit with an "all good!" code if it was interrupted unexpectedly. The current behavior is expected.

For your specific use-case, take a look into handling this at the command running layer. Could you wrap your call to Polymer in some logic to handle/ignore the SIGINT call to polymer? [This search](https://www.google.com/search?q=ignore+sigint+exit+code+bash&oq=ignore+sigint+exit+code+bash&aqs=chrome..69i57.5749j0j1&sourceid=chrome&ie=UTF-8) might get you pointed in the right direction.

Good luck!

PR #809, opened 2017-06-24
Enhancement/better web component name validation

vilarfg:
By using Sindre Sorhus' [validate-element-name](https://www.npmjs.com/package/validate-element-name), we can make sure the element's name comply with the [specs](https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name).

 - [x] CHANGELOG.md has been updated

PR #903, opened 2017-10-04
 Enhancement/ Add Custom Element Name validation to init element

NiclasvanEyk:
This Pull Request improves the validation of the name provided to the init-element generator.
I wanted to try out polymer and named my element 'Test-Element'. 

polymer-upstream-test-element

The generator did not complain, but when I opened up the demo page in my Browser, an error was logged to the console because the name was not a valid custom element name, according to the [spec](https://www.w3.org/TR/custom-elements/#prod-potentialcustomelementname).

polymer-upstream-test-element-console

I used the [validate-element-name](https://github.com/sindresorhus/validate-element-name), which provides nice error-messages. I also wrote some typings for the package.
On top of validating the name according to the spec, the package also checks whether the name contains prefixes like 'polymer-', 'x-' or 'ng-'. If this is the case, the validation succeeds, but a warning is logged to the console.

polymer-fix-init
_Validation, according to the spec_

polymer-fix-warning
_Warning for prefixes, that should be avoided_

Later I realized that this is exactly was [this issue](https://github.com/Polymer/polymer-linter/issues/95) with polymer-linter is about and that there is an existing PR #809 that also fixes this issue.
But that issue is a few months old, so I thought I submit this anyway.
Let me know if this is not needed anymore.


NiclasvanEyk:
I signed it!

src/init/element/element.ts
```diff
@@ -53,13 +57,16 @@ export function createElementGenerator(templateName: string):
...
-            if (!nameContainsHyphen) {
-              this.log(
-                  '\nUh oh, custom elements must include a hyphen in ' +
-                  'their name. Please try again.');
+            const nameValidation = validateElementName(name);
+
+            if (!nameValidation.isValid) {
+              this.log(`\n${nameValidation.message}\nPlease try again.`);
+            } else if (nameValidation.isValid && nameValidation.message) {
+              this.log('');
```
TimvdLippe: Shouldn't this just be `this.log(nameValidation.message)`?

NiclasvanEyk: This would also be possible, but you'd have to prepend a `\n`, because the prompt for the element name doesn't automatically add a newline at the end. The call to basically `this.log('')` just acts as a newline.

If I would just call `this.log(nameValidation.message)` there would also be no yellow `warn:  `-prefix. I tested it first with `this.log` but that was, in my opinion, to subtle of a warning.
Also, if I'd call `logger.warn('\n' + nameValidation.message);`, the yellow `warn: `-prefix would be on the same line as the element-name prompt.

If you want to I could still change it though or at least add a comment indicating, that this is on purpose.

stramel:
Would love to see this get merged after conflicts are fixed. This would align with what I implemented in `polymer-linter`.

justinfagnani:
Thanks @NiclasvanEyk! Will merge when CI is done.

PR #912, opened 2017-10-13
Support lint --fix

rictic:

### DIFF
--- a/src/lint/lint.ts
+++ b/src/lint/lint.ts
@@ -20,7 +20,7 @@ import {Analysis, Analyzer, FSUrlLoader, PackageUrlResolver, Severity, Warning} 
 import {WarningFilter} from 'polymer-analyzer/lib/warning/warning-filter';
 import {WarningPrinter} from 'polymer-analyzer/lib/warning/warning-printer';
 import * as lintLib from 'polymer-linter';
-import {applyEdits, FixableWarning, makeParseLoader, Replacement} from 'polymer-linter/lib/warning';
+import {applyEdits, Edit, FixableWarning, makeParseLoader, Replacement} from 'polymer-linter/lib/warning';
 import {ProjectConfig} from 'polymer-project-config';
 
 import {CommandResult} from '../commands/command';
@@ -61,25 +61,37 @@ export async function lint(options: Options, config: ProjectConfig) {
   // of a lint run rather than doing this sort of parallel construction.
   // We want the analysis that the linter sees so that we can apply fixes
   // to the same version of the files that the fixes were constructed for.
+  // https://github.com/Polymer/polymer-linter/issues/110
   let warnings;
   let analysis;
   if (options.input) {
     [warnings, analysis] = await Promise.all(
         [linter.lint(options.input), analyzer.analyze(options.input)]);
+
+    const packageRelativeInputFilenames = new Set(options.input.map(
+        (i) => path.relative(config.root, path.resolve('.', i))));
+    warnings = warnings.filter(
+        (w) => packageRelativeInputFilenames.has(w.sourceRange.file));
   } else {
     [warnings, analysis] =
         await Promise.all([linter.lintPackage(), analyzer.analyzePackage()]);
   }
 
+
   const filtered = warnings.filter((w) => !filter.shouldIgnore(w));
 
   if (options.fix) {
-    return fix(filtered, options, config, analyzer, analysis);
+    return fix(filtered, config, analyzer, analysis);
   } else {
     return report(filtered);
   }
 }
 
+/**
+ * Report a friendly description of the given warnings to stdout.
+ *
+ * @param warnings Warnings to report.
+ */
 async function report(warnings: Warning[]) {
   const printer =
       new WarningPrinter(process.stdout, {verbosity: 'full', color: true});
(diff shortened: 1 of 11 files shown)


src/lint/lint.ts
```diff
@@ -48,37 +51,51 @@ export async function lint(options: Options, config: ProjectConfig) {
...
     urlLoader: new FSUrlLoader(config.root),
     urlResolver: new PackageUrlResolver(),
   });
-
   const linter = new lintLib.Linter(rules, analyzer);
 
+  // This is awkward, it would be nice to be able to get access to the analysis
+  // of a lint run rather than doing this sort of parallel construction.
+  // We want the analysis that the linter sees so that we can apply fixes
+  // to the same version of the files that the fixes were constructed for.
```
FredKSchott: This is very reasonable, can we file an issue to track this & then link it here?

src/lint/lint.ts
```diff
@@ -48,37 +51,51 @@ export async function lint(options: Options, config: ProjectConfig) {
...
+        await Promise.all([linter.lintPackage(), analyzer.analyzePackage()]);
   }
 
   const filtered = warnings.filter((w) => !filter.shouldIgnore(w));
 
-  const printer = new WarningPrinter(
-      process.stdout, {verbosity: 'full', color: true});
-  await printer.printWarnings(filtered);
+  if (options.fix) {
+    return fix(filtered, options, config, analyzer, analysis);
```
FredKSchott: Rename `fixAndReport()` (or similar) for clarity, I was about to ask if we really wanted to fix silently :)

src/lint/lint.ts
```diff
@@ -48,37 +51,51 @@ export async function lint(options: Options, config: ProjectConfig) {
...
+  await printer.printWarnings(warnings);
 
-  if (filtered.length > 0) {
+  if (warnings.length > 0) {
     let message = '';
-    const errors = filtered.filter((w) => w.severity === Severity.ERROR);
-    const warnings = filtered.filter((w) => w.severity === Severity.WARNING);
-    const infos = filtered.filter((w) => w.severity === Severity.INFO);
+    const errors = warnings.filter((w) => w.severity === Severity.ERROR);
+    const warningLevel =
```
FredKSchott: why the name change? What's a warning level?

src/lint/lint.ts
```diff
@@ -87,3 +104,64 @@ export async function lint(options: Options, config: ProjectConfig) {
...
+
+  for (const [file, count] of changeCountByFile) {
+    console.log(`  Made ${count} change${plural(count)} to ${file}`);
+  }
+
+  if (incompatibleEdits.length > 0) {
+    console.log(
+        `\nFixed ${appliedEdits.length} ` +
+        `warning${plural(appliedEdits.length)}, ` +
+        `${incompatibleEdits.length} had conflicts with other fixes. ` +
```
FredKSchott: As a user I should be told which files these occurred in, assuming that it is be possible for two fixes to conflict indefinitely (or for a fix to inadvertently cause a new error)

src/lint/lint.ts
```diff
@@ -48,37 +51,51 @@ export async function lint(options: Options, config: ProjectConfig) {
...
-      process.stdout, {verbosity: 'full', color: true});
-  await printer.printWarnings(filtered);
+  if (options.fix) {
+    return fix(filtered, options, config, analyzer, analysis);
+  } else {
+    return report(filtered);
+  }
+}
+
+async function report(warnings: Warning[]) {
```
FredKSchott: add jsdoc function description

src/lint/lint.ts
```diff
@@ -87,3 +104,64 @@ export async function lint(options: Options, config: ProjectConfig) {
     return new CommandResult(1);
   }
 }
+
+async function fix(
```
FredKSchott: add jsdoc function description

src/lint/lint.ts
```diff
@@ -87,3 +104,64 @@ export async function lint(options: Options, config: ProjectConfig) {
...
+    warnings: FixableWarning[],
+    options: Options,
+    config: ProjectConfig,
+    analyzer: Analyzer,
+    analysis: Analysis) {
+  if (options.input) {
```
FredKSchott: should this be handled outside of the function? If I gave a set of warnings to a `fix()` function to fix, I would expect them all to be fixed/reported. At the very least, document this to make expected behavior clear.

src/lint/lint.ts
```diff
@@ -87,3 +104,64 @@ export async function lint(options: Options, config: ProjectConfig) {
...
+    const packageRelativeInputFilenames = new Set(options.input.map(
+        (i) => path.relative(config.root, path.resolve('.', i))));
+    warnings = warnings.filter(
+        (w) => packageRelativeInputFilenames.has(w.sourceRange.file));
+  }
+
+  const fixes = warnings.map((w) => w.fix).filter((fix) => fix !== undefined) as
+      Replacement[][];
+  if (fixes.length === 0) {
```
FredKSchott: Consider adding: When I run this CLI command my intention is to lint for errors and then fix any that are found. With that understanding, it would be more clear to log "No errors found" if there were no warnings to begin with.

src/lint/lint.ts
rictic: +1, and better to be more consistent between reporting and fixing

src/lint/lint.ts
rictic: Done.

src/lint/lint.ts
rictic: It shadowed the other `warnings` variable. Clarified.

src/lint/lint.ts
rictic: Done

src/lint/lint.ts
rictic: We don't report in the same sense as `report()` does. Clarified in the jsdoc, WDYT?

src/lint/lint.ts
rictic: Done

src/lint/lint.ts
```diff
@@ -88,14 +100,14 @@ async function report(warnings: Warning[]) {
   if (warnings.length > 0) {
     let message = '';
     const errors = warnings.filter((w) => w.severity === Severity.ERROR);
-    const warningLevel =
+    const warningLevelWarnings =
```
FredKSchott: +10 clarity (side effect: +1 silliness)

src/lint/lint.ts
```diff
@@ -159,6 +171,18 @@ async function fix(
   }
 }
 
+function countEditsByFile(edits: Edit[]) {
```
FredKSchott: nit: description :)

rictic:
+1, I added the parameter args because I thought that's what you were asking for.

Agreed that they weren't adding much, removed.

Issue #849, opened 2017-08-19
Polymer 1 Bundled builds fail


milesje:
### Description
Polymer 1.X build with bundle: true fails.

### Versions & Environment
- Polymer CLI: 1.4.1
- node: 8.4
- Operating System: Windows 10

#### Steps to Reproduce
1. polymer build - with bundle: true (or use one of the bundled presets)

#### Expected Results
Build completes

#### Actual Results
build errors out. If you disable bundle or use an unbundled preset it builds fine.

milesje:
Here is a copy of my polymer.json
```
{
  "entrypoint": "index.html",
  "shell": "src/netd-auth.html",
  "fragments": [
    "src/add-user.html",
    "src/apps-view.html",
    "src/change-password.html",
    "src/create-role.html",
    "src/forgot-password.html",
    "src/login-view.html",
    "src/netd-app.html",
    "src/profile-view.html",
    "src/reset-password.html",
    "src/roles-view.html",
    "src/users-view.html"
  ],
  "sources": [
    "src/**/*",
    "images/**/*",
    "bower.json"
  ],
  "extraDependencies": [
    "manifest.json",
    "bower_components/webcomponentsjs/webcomponents-lite.min.js"
  ],
  "builds": [{
    "preset": "es5-bundled"
  },{
    "preset": "es6-unbundled"
  }],
  "lint": {
    "rules": [
      "polymer-1"
    ]
  }
}
```

milesje:
doing a unbundled build work fine.

web-padawan:
What kind of error do you see?
BTW, es6 preset is not needed for 1.x

milesje:
I know that es6 preset is not needed, I'm also working on coverting from 1.x to 2

milesje:
```
error:   Promise rejection: Error: Failed to execute "git ls-remote --tags --heads https://github.com/PolymerElements/polymer.git", exit code of #128
remote: Repository not found.
fatal: repository 'https://github.com/PolymerElements/polymer.git/' not found

error:   Error: Failed to execute "git ls-remote --tags --heads https://github.com/PolymerElements/polymer.git", exit code of #128
remote: Repository not found.
fatal: repository 'https://github.com/PolymerElements/polymer.git/' not found

    at createError (C:\Users\jmiles\AppData\Roaming\nvm\v8.4.0\node_modules\polymer-cli\node_modules\bower\lib\util\createError.js:4:15)
[...]

PR #916, opened 2017-10-19
Build: determine Polymer version via bower list using offline option

jstclair:
When building, polymer-cli uses 'bower list' to determine the current Polymer version in order to
set the 'rewriteUrlsInTemplates' option. However, 'bower list' by default will attempt to read
from the underlying package source (since it's typically used at the command-line to see if any
package has upgrades available). This behavior is undesirable (and unnecessary), particularly in CI.

I've also updated the TS custom_typing to match the actual bower list command signature

### DIFF
--- a/custom_typings/bower.d.ts
+++ b/custom_typings/bower.d.ts
@@ -8,12 +8,13 @@ declare module 'bower' {
     directory?: string;
     interactive?: boolean;
     save?: boolean;
+    offline?: boolean;
   }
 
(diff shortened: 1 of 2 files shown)


jstclair:
Just wanted to note that this may fix #849 - that's the same set of errors we would see at build time

jstclair:
Signed CLA

justinfagnani:
Thanks @jstclair!

PR #920, opened 2017-10-31
Add --auto-base-path flag to build.

aomarks:
This new flag sets the entrypoint `<base>` tag for all builds to match the name of that build. Unlike the other flags, it does not necessarily trigger a single one-off build. Instead, it switches the option on for *all* builds (including a one-off, if another flag triggered one).

The idea is to make building for prpl-server easier. See https://github.com/Polymer/prpl-server-node/issues/24 for motivation. Right now you need need to annotate each build in your `polymer.json` with `basePath: true` to make prpl-server work. The problem with this is that:

1) It's kind of unintuitive/odd ergonomically because you would almost always want this option all-or-nothing. Setting it in every build feels weird, especially with presets (we used to have this option on for the presets, but that broke the non-prpl-server use case by default).
2) It's something you would typically want to switch on/off for development vs deployment (since you can't directly serve a build when it's on), so having to update `polymer.json` makes this cumbersome.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## Unreleased
 - Added `--fix` option to `polymer lint`. When passed, some warnings with simple mechanical solutions will be fixed.
-- `build` Added a CLI argument for setting the `basePath` option: `--base-path`.
+- `build` Added `--base-path` flag. Sets the entrypoint `<base>` of the single one-off build to the specified value.
+- `build` Added `--auto-base-path` flag. Sets the entrypoint `<base>` tag for all builds to match the name of that build. Unlike other flags, does not necessarily trigger a single one-off build.
 <!-- Add new, unreleased items here. -->
 
 ## v1.5.7 [10-11-2017]
--- a/src/commands/build.ts
+++ b/src/commands/build.ts
@@ -29,6 +29,10 @@ import {Command, CommandOptions} from './command';
 
 const logger = logging.getLogger('cli.command.build');
 
+// This flag is special because it doesn't necessarily trigger a one-off single
+// build.
+const autoBasePath = 'auto-base-path';
+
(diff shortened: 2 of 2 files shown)


abdonrd:
I'm interested in the `prpl-server`. Any news? :)

aomarks:
> I'm interested in the prpl-server. Any news? :)

I put merging this PR on hold because of some concern that there should be a way to make setting the base tag more convenient without cluttering our flags. Maybe a top-level `autoBasePath` to the config file. Any thoughts @abdonrd? Do you find that you frequently want to jump between building with and without base path transformation, or would a config setting be just as useful?

abdonrd:
You mean something like this:

```json
{
  ...
  "autoBasePath": true,
  "builds": [
    {
      "preset": "es5-bundled"
    },
    {
      "preset": "es6-bundled"
    },
    {
      "preset": "es6-unbundled"
    }
  ]
}
```

Instead like this?

```json
{
  ...
  "builds": [
    {
      "preset": "es5-bundled",
      "basePath": true
    },
    {
      "preset": "es6-bundled",
      "basePath": true
    },
    {
      "preset": "es6-unbundled",
      "basePath": true
    }
  ]
}
```

<hr>

And if we always add `"basePath": true` as long as the `builds` array is greater than 1?
Whenever I have more than one build, it is to use the `prpl-server`.

aomarks:
> And if we always add "basePath": true as long as the builds array is greater than 1?
> Whenever I have more than one build, it is to use the prpl-server.

Exactly what I was thinking. Ok, I'll try to get a PR out to add that config options. Closing this one for now.

abdonrd:
Thanks @aomarks!

abdonrd:
@aomarks tell me if it works: https://github.com/Polymer/polymer-cli/pull/942 🙂

PR #925, opened 2017-11-14
Convert tests to typescript

rictic:
- [x] CHANGELOG.md not updated, internal change

This should be a completely noop change. I did convert our test code to async/await though.


src/init/init.ts
```diff
@@ -230,7 +230,7 @@ function createSelectPrompt(env: YeomanEnvironment): InquirerQuestion {
  * error will be thrown.
  */
 export async function runGenerator(
-    generatorName: string, options: {[name: string]: any}): Promise<void> {
+    generatorName: string, options?: {[name: string]: any}): Promise<void> {
```
justinfagnani: Can ignore for now, but I prefer doing `options: {[name: string]: any} = {}` and removing the next line.

src/test/integration/lint_test.ts
```diff
@@ -0,0 +1,126 @@
...
+const fixturePath =
+    path.join(__dirname, '../../../src/test/integration/fixtures/');
+
+suite('polymer lint', function() {
+
+  const binPath = path.join(__dirname, '../../../bin/polymer.js');
+
+  this.timeout(2 * 1000);
+
+  test.only('handles a simple correct case', async () => {
```
stramel: Should this test be focused?

rictic: Good catch! Thank you thank you thank you!

src/init/init.ts
rictic: +1, fixed since I needed to resolve the `.only` issue too

PR #928, opened 2017-11-16
Implement `polymer lint --watch`

rictic:

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -67,7 +67,7 @@ gulp.task(
               }))
               .pipe(tslint.report()));
 
-gulp.task('depcheck', () => {
+gulp.task('depcheck', ['build'], () => {
   return depcheck(__dirname, {
            // "@types/*" dependencies are type declarations that are
            // automatically
--- a/src/lint/lint.ts
+++ b/src/lint/lint.ts
@@ -79,7 +79,8 @@ async function runOnce(
     options: PrivateOptions,
     config: ProjectConfig,
     filter: WarningFilter,
-    editActionsToAlwaysApply = new Set(options.edits || [])) {
+    editActionsToAlwaysApply = new Set(options.edits || []),
+    watcher?: FilesystemChangeStream) {
   let warnings;
   if (options.input) {
     warnings = await linter.lint(options.input);
@@ -91,14 +92,22 @@ async function runOnce(
   const filtered = warnings.filter((w) => !filter.shouldIgnore(w));
 
   if (options.fix) {
-    const appliedFixes = await fix(
+    const changedFiles = await fix(
         filtered,
         options,
         config,
         analyzer,
         analysis,
         editActionsToAlwaysApply);
-    if (!appliedFixes && options.reportIfNoFix) {
+    if (watcher) {
+      // Some file watcher interfaces won't notice this change immediately after
+      // the one that initiated this lint run. Ensure that we notice these
+      // changes.
+      for (const changedFile of changedFiles) {
+        watcher.ensureChangeIsNoticed(changedFile);
+      }
+    }
+    if (changedFiles.size === 0 && options.reportIfNoFix) {
       await report(filtered);
     }
   } else {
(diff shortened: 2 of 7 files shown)


src/lint/lint.ts
```diff
@@ -57,6 +58,29 @@ export async function lint(options: Options, config: ProjectConfig) {
...
+
+interface PrivateOptions extends Options {
+  /**
+   * When running in --watch mode we want to report warnings if we're running
+   * with --fix but there weren't any warnings to fix.
+   */
+  reportIfNoFix?: boolean;
+}
+
+async function runOnce(
```
FredKSchott: +docs! also I'd prefer `run` instead of `runOnce`, I think run being singular is implied, especially if there's a second method that is explicitly "multiple/loop/watch/etc"

src/lint/lint.ts
```diff
@@ -68,12 +92,133 @@ export async function lint(options: Options, config: ProjectConfig) {
...
+  const lintActionsToAlwaysApply = new Set(options.edits || []);
+
+  const paths =
+      new Set([...analysis.getFeatures({kind: 'document'})].map((d) => d.url));
+  const watcher = new FilesystemChangeStream(
+      chokidar.watch([...paths], {persistent: true}));
+  for await(const changeBatch of watcher) {
+    await analyzer.filesChanged([...changeBatch]);
+
+    await runOnce(
```
FredKSchott: oh, wait, do you pass a `watcher` to `runOnce()`? Can you document this?

src/lint/lint.ts
```diff
@@ -68,12 +92,133 @@ export async function lint(options: Options, config: ProjectConfig) {
...
+  constructor(watcher: chokidar.FSWatcher) {
+    watcher.on('change', (path: string) => {
+      this.noticeChange(path);
+    });
+    watcher.on('unlink', (path: string) => {
+      this.noticeChange(path);
+    });
+  }
+
+  private noticeChange(path: string) {
```
FredKSchott: +docs!

src/lint/lint.ts
```diff
@@ -68,12 +92,133 @@ export async function lint(options: Options, config: ProjectConfig) {
...
+        for (const path of notices) {
+          this.noticeChange(path);
+        }
+        this.outOfBandNotices = undefined;
+      }, 100);
+    }
+    this.outOfBandNotices.add(path);
+  }
+
+  async * [Symbol.asyncIterator](): AsyncIterator<Set<string>> {
```
FredKSchott: +docs!

rictic:
All comments addressed!

rictic:
[Justin's out this week, so passing this review over to Fred]

gulpfile.js
```diff
@@ -67,7 +67,7 @@ gulp.task(
               }))
               .pipe(tslint.report()));
 
-gulp.task('depcheck', () => {
+gulp.task('depcheck', ['build'], () => {
```
FredKSchott: why build first? it should be analyzing our typescript directly

src/lint/lint.ts
```diff
@@ -57,6 +58,36 @@ export async function lint(options: Options, config: ProjectConfig) {
...
+   * with --fix but there weren't any warnings to fix.
+   */
+  reportIfNoFix?: boolean;
+}
+
+/**
+ * Run a single pass of the linter, and then report the results or fix warnings
+ * as requested by `options`.
+ *
+ * In a normal run this is called once and then it's done. When running with
```
FredKSchott: super clear 👍

src/lint/lint.ts
```diff
@@ -68,12 +99,149 @@ export async function lint(options: Options, config: ProjectConfig) {
...
+    analysis = await analyzer.analyzePackage();
+  }
+  /** Remember the user's preferences across runs. */
+  const lintActionsToAlwaysApply = new Set(options.edits || []);
+
+  const paths =
+      new Set([...analysis.getFeatures({kind: 'document'})].map((d) => d.url));
+  const watcher = new FilesystemChangeStream(
+      chokidar.watch([...paths], {persistent: true}));
+  for await (const changeBatch of watcher) {
```
FredKSchott: man, `for await` is wild.

So the intention is that this iterator never completes, until the user kills the process, correct? That would be worth a comment, just to make sure future readers don't follow through past this.

src/test/integration/lint_test.ts
```diff
@@ -183,6 +184,77 @@ Fixed 4 warnings.
 </dom-module>
 `);
   });
+
+  suite('--watch', function() {
```
FredKSchott: clever test 👍

gulpfile.js
rictic: Without this change it crashes saying that `chokidar` is unused.

Adding some console.log statements around makes me think that it's unable to handle `lint/lint.ts`, giving a warning in the babylon parser, a little bit after an `await` statement. It seems like it's not handling typescript perfectly. Fortunately by ensuring that `build` happens first, it will also run over our generated JS, which it does handle correctly.

src/lint/lint.ts
rictic: I know right!? It's the best.

I added a comment mentioning that this is a forever loop.

src/test/integration/lint_test.ts
rictic: Thanks!

PR #929, opened 2017-11-17
Regenerate package-lock.json, delete yarn.lock, update lint/lint.ts usage

rictic:
e.g. uses the upstreamed Warning implementation in the analyzer.

 - [x] CHANGELOG.md not updated, internal change


FredKSchott:
> The recommended use-case for npm-shrinkwrap.json is applications deployed through the publishing process on the registry: for example, daemons and command-line tools intended as global installs or devDependencies. It's strongly discouraged for library authors to publish this file, since that would prevent end users from having control over transitive dependency updates.

https://docs.npmjs.com/files/shrinkwrap.json

You'll need to run `npm shrinkwrap`, which takes a `package-lock.json` and outputs a `npm-shrinkwrap.json`.

rictic:
Let's do that on release rather than maintaining it with each PR. WDYT?

rictic:
This is just so that our tests are matching the latest public version.

Maybe we should delete our lockfiles and shrinkwraps outside of release commits?

Issue #613, opened 2017-03-17
Feature Request: Ability to select a single build


coreyfarrell:
When using `polymer build` all targets found in polymer.json "builds" array are always built.  It would be useful if this could be controlled by config and command-line.

Example polymer.json builds section:
```
{
  "builds": [
    {
      "name": "default",
      "bundle": true,
      "js": {"minify": true},
      "css": {"minify": true},
      "html": {"minify": true}
    },
    {
      "name": "debug",
      "bundle": false,
      "autobuild": false
    }
  ]
}
```

The goal is that running `polymer build` should only build "default", but the "debug" settings can be left in polymer.json.  For compatibility the "autobuild" setting has to be true by default.

Second part of this request is the ability to select one or more build(s) from command-line.  This way if my code is broken I could run `polymer build --target debug` to build "debug" only instead of trying to troubleshoot in a giant bundled minified source.  The command-line option should cause all "autobuild" options to be ignored and only build the requested targets.

FredKSchott:
+1 I agree, this would be a really useful feature for troubleshooting single builds. Also for faster iteration during development (1 target builds faster than N targets)

timweightman:
I think this is a great idea, I'm definitely missing the feature right now.
Rather than `autobuild`, I would prefer env-specific config similar to #408 proposal.

Example:
```
{
  "environments": {
    "DEV": {
      "builds": [
        {
          "name": "debug",
          "bundle": false,
        }
      ]
    },
    "PROD": {
      "builds": [
        {
          "name": "default",
          "bundle": true,
          "js": {"minify": true},
          "css": {"minify": true},
          "html": {"minify": true}
        }
      ]
    }
  }
}
```

If there is a root `"builds"` property in polymer.json, it would always be built, and based on environment config, there may be additional builds for DEV, PROD etc.

justinfagnani:
I think this is a great issue for a contribution, if anyone wants to take a crack.

davidmaxwaterman:
How is this different to the `--name` option? Not that `--name` actually works, but hey.

jsilvermist:
@davidmaxwaterman AFAIK `--name` is for setting the name of a build, aka the folder it's placed into.

I believe it currently doesn't take into account whether or not there's a entry in the `polymer.json`, so for a bundled build it would be:
```sh
polymer build --name bundled --bundle
```
even if you already have an entry in your `polymer.json` such as:
```json
{
  "name": "bundled",
  "bundle": true
}
[...]

PR #930, opened 2017-11-17
Fix #827. Merge commandline options with named build configuration options

### LINKED ISSUE #827: compile="true" configuration setting ignored when building 'by name'
### Versions & Environment
- Polymer CLI: 1.3.1
- node: v6.10.1
- Operating System: Windows 10

#### Steps to Reproduce
- Start with a command prompt in a new empty directory
- Execute a polymer init, choosing polymer-2-application (any application will do)
- Open polymer.json, and add a build configuration, to have in the end this result:
```
{
 "lint": {
   "rules": [
     "polymer-2"
   ]
 },
 "builds": [
    {
      "name": "test",
      "bundle": true,
      "js": {
        "minify": true,
        "compile": true
      },
      "css": {
        "minify": true
      },
      "html": {
        "minify": true
      }
    }
 ]
}
```

- Now, from the command line:
`polymer build`
Will produce the correct result (application compiled from ES6 to ES5, see build/test directory)
`polymer build --name test`
Will produce a wrong result (application NOT compiled from ES6 to ES5, see build/test directory)

alber70g:
This fixes #827 
When building from command line with `--name <buildName>`, it now takes the options from the command line, merges them with the project's options and continues the build.

It's probably fixing #613 as well.

### DIFF
--- a/src/build/build.ts
+++ b/src/build/build.ts
@@ -38,8 +38,27 @@ export async function build(
     options: ProjectBuildOptions,
     polymerProject: PolymerProject): Promise<void> {
   const buildName = options.name || 'default';
-  const optimizeOptions:
-      OptimizeOptions = {css: options.css, js: options.js, html: options.html};
+
+  // If a config for the current buildName is present
+  // Merge the projects optimizeOptions with the options
+  // Commandline options will override the projects options
+  const optimizeOptions: OptimizeOptions = { css: options.css, js: options.js, html: options.html };
+
+  // Remove undefined properties to prevent overriding the projects options
+  (!optimizeOptions.css) && delete optimizeOptions.css;
+  (!optimizeOptions.js) && delete optimizeOptions.js;
+  (!optimizeOptions.html) && delete optimizeOptions.html;
+
+  if (polymerProject.config.builds) {
+    const projectOptions = polymerProject.config.builds.find((buildConfig) => buildConfig.name === buildName);
+    if (projectOptions) {
+      const projectOptimizeOptions = { css: projectOptions.css, js: projectOptions.js, html: projectOptions.html };
+      // Assign each property separately, as it's not a deep assign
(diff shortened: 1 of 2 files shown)


src/build/build.ts
```diff
@@ -38,8 +38,18 @@ export async function build(
     options: ProjectBuildOptions,
     polymerProject: PolymerProject): Promise<void> {
   const buildName = options.name || 'default';
-  const optimizeOptions:
-      OptimizeOptions = {css: options.css, js: options.js, html: options.html};
+
+  // If a config for the current buildName is present
+  // Merge the projects optimizeOptions with the options
+  // Commandline options will override the projects options
+  const projectOptions = polymerProject.config.builds.find(buildConfig => buildConfig.name === buildName);
```
TimvdLippe: You can destructure css, js and html directly. Then line 49 instead of projectOptimizeOptions you can do `{css, js, html}`

TimvdLippe (commented):
Would be great to see a regression test for this as well, if you are able to figure that out (not sure if we have existing testing infrastructure for that)

src/build/build.ts
Alber70g: But what if one of the options is not available, then the option where it should be merged to will be overridden with `undefined`

Alber70g: Also, haven't seen any regression tests for the build/cli options. So that's why I didn't add one

Alber70g:
Please take another look. I've changed it quite a bit. Should be better now

- Command line options weren't taken into account previously.
- When there are no command line options, all further options were overridden by 'undefined'

Can you tell me where to put the tests? Since it's quite a bit more complex than thought at the start

Alber70g:
There are no tests at this moment for this. Please provide an infrastructure for it, or merge this MR because it fixes the `--name <buildName>` issue

aomarks:
Closing this pull request because this repo is now archived. If this PR is still active, please submit a new one at https://github.com/Polymer/tools.

PR #940, opened 2017-12-24
Reduce the amount of noise in test output.

rictic:
This captures all non-plylog output and lets us assert on it rather than logging it from the test runner.

It does not intercept plylog, as plylog is really hard to intercept. For the editor service, when it was a hard requirement, the only way I was able to do it was by importing a special log interceptor module before any plylog-using modules.

 - [x] CHANGELOG.md not been updated, internal change


rictic:
Hm, looks like the appveyor failure might be a flake due to some on-install behavior of a dependency.

rictic:
Bump!

PR #941, opened 2017-12-25
Use strict type checking.

rictic:
- [x] CHANGELOG.md not updated, noop change

Stubbed my toe on an implicitly `any` `this`. This will opt us into all the strictness.


rictic:
Yeah, comments are legal in tsconfig files

stramel:
Hmm seems strange but okay 👍🏼 Thanks for the knowledge

PR #942, opened 2017-12-27
If there are multiple builds, add the `basePath` flag to each build

abdonrd:
If there are multiple builds, add the `basePath` flag to each build for the `prpl-server` support.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
   - Also works with `--fix` to automatically fix and report warnings as you work!
 - `build` Added a CLI argument for setting the `basePath` option: `--base-path`.
 - Derives node version check from the package.json.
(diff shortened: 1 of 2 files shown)


aomarks:
I think we should only do this if the config file has `autoBasePath: true`, otherwise this will be a pretty big breaking change. Sorry I wasn't very clear in the other PR discussion.

Also I think this logic should actually live in `polymer-build`, probably in the `PolymerProject` constructor: https://github.com/Polymer/polymer-build/blob/master/src/polymer-project.ts#L30

Thanks for the PR!

abdonrd:
> otherwise this will be a pretty big breaking change

Oh, totally right! 🤦‍♂️ 

New PR here: https://github.com/Polymer/polymer-build/pull/279

Forgive the possible failures, it's my first time with TypeScript and with the Polymer Build project. 😅

PR #948, opened 2018-01-11
Adapt to analyzer, build, linter v3

valdrinkoshi:


src/lint/lint.ts
```diff
@@ -320,8 +320,9 @@ async function fix(
       await applyEdits(edits, makeParseLoader(analyzer, analysis));
 
   for (const [newPath, newContents] of editedFiles) {
+    const filePath = newPath.replace('file://', '').replace(config.root, '');
```
valdrinkoshi: this is ugly, I'm probably missing something, but makes the --fix work

justinfagnani: We need a common utility for this, but something like:

```ts
    const url = new URL(newPath);
    let documentPath = decodeURIComponent(documentURL.pathname);
    if (isWindows() && documentPath.startsWith('/')) {
      documentPath = documentPath.substring(1);
    }
```

justinfagnani:
@valdrinkoshi can you rebase and try to upgrade to the latest Analyzer 3 preview?

src/lint/lint.ts
valdrinkoshi: I moved this url resolution in `util`, method `resolvePath`. I'm using https://www.npmjs.com/package/url-js for getting the `URL` object, as it's a minimal package with no deps.

valdrinkoshi:
@justinfagnani I've rebased and updated deps to latest analyzer and linter (forgot the linter previously!)

I'm still seeing some issues with path resolution: when invoking `polymer lint` I get warnings with the full path, and when invoking `polymer lint --fix` the warning doesn't get resolved:
```sh
$ echo "<dom-module name='foo-elem'></dom-module>" > test.html
$ polymer lint test.html -r polymer-2-hybrid


<dom-module name='foo-elem'></dom-module>
            ~~~~

file:///Users/valdrin/dev/polymer-cli/test.html(1,13) warning [dom-module-invalid-attrs] - Use the "id" attribute rather than "name" to associate the tagName of an element with its dom-module.

Found 1 warning.

$ polymer lint test.html -r polymer-2-hybrid --fix
No fixes to apply.
```

@rictic any ideas on this one? I haven't been following the progresses on `polymer-linter` lately

stramel:
@valdrinkoshi That fix wasn't part of the lastest release. As of the latest release that should just warn and not be fixable.

src/lint/lint.ts
```diff
@@ -320,8 +320,9 @@ async function fix(
       await applyEdits(edits, makeParseLoader(analyzer, analysis));
 
   for (const [newPath, newContents] of editedFiles) {
+    const documentPath = resolvePath(newPath);
```
rictic: I forgot! We added a method on FsUrlLoader to deal with this: `FsUrlLoader#getFilePath()`

That might mean that we'd have to pass down an FsUrlLoader into this function (I think this collection of functions has enough shared state that it should be a class, but that's a different issue)

valdrinkoshi: oh nice! that makes the whole thing much simpler, we don't need to pass the config around but rather the urlLoader 👌 Fixed!

valdrinkoshi:
@stramel good catch, I spent some time on that one w/o realizing i took a test from master :x
Anyways, fixed and tested with: 
```sh
$ echo "<dom-module id='foo-elem'><style></style><template></template></dom-module>" > test.html

$ polymer lint test.html -r polymer-2

<dom-module id='foo-elem'><style></style><template></template></dom-module>
                          ~~~~~~~

file:///Users/valdrin/dev/polymer-cli/test.html(1,27) warning [style-into-template] - Style tags should not be directchildren of <dom-module>, they should be moved into the <template>


Found 1 warning. 1 can be automatically fixed with --fix.
```
And
```sh
$ polymer lint test.html -r polymer-2 --fix
  Made 2 changes to file:///Users/valdrin/dev/polymer-cli/test.html

Fixed 1 warning.
```

The PR fails because many tests still expect relative paths printed (instead of absolute ones).
I think it would be nicer to have relative paths printed on console, no? @rictic WDYT? should we update tests to expect absolute paths here or fix this in the linter?

src/lint/lint.ts
```diff
@@ -320,8 +320,12 @@ async function fix(
       await applyEdits(edits, makeParseLoader(analyzer, analysis));
 
   for (const [newPath, newContents] of editedFiles) {
-    await fs.writeFile(
-        path.join(config.root, newPath), newContents, {encoding: 'utf8'});
+    const documentPath = urlLoader.getFilePath(newPath);
```
valdrinkoshi: `newPath` is already an absolute path, and `urlLoader.getFilePath` returns the same absolute path...I'm a bit confused now: either we use urlLoader to get the relative path here, or we don't use it at all and always print absolute paths. What should be done here?

rictic: Hm. newPath should be a URL, documentPath should be a FS path. Is that not what you're seeing?

src/lint/lint.ts
```diff
@@ -52,18 +51,19 @@ export async function lint(options: Options, config: ProjectConfig) {
     minimumSeverity: Severity.WARNING
```
stramel: @valdrinkoshi Do you mind linking up the new `filesToIgnore` option for the `WarningFilter`?

stramel:
Can't wait for this so I can test upgrading my element's using all the new rules and fixes in linter! 🎉

package.json
```diff
@@ -15,7 +15,8 @@
     "test": "gulp test lint",
     "test:smoke": "gulp test",
     "test:integration": "gulp test:integration",
-    "format": "find src gulpfile.js \\( -iname '*.ts' -o -iname '*.js' \\) -not -path '*fixtures*' | xargs clang-format --style=file -i"
+    "format": "find src gulpfile.js \\( -iname '*.ts' -o -iname '*.js' \\) -not -path '*fixtures*' | xargs clang-format --style=file -i",
```
stramel: You can clean this command up a bit by doing,

```
"format": "find src gulpfile.js -iname '*.[jt]s' -not -path '*fixtures*' | xargs clang-format --style=file -i"
```

package.json
```diff
@@ -15,7 +15,8 @@
...
+    "test:watch": "tsc-then -- mocha -c --ui tdd lib/test/integration/*_test.js lib/test/unit/*_test.js lib/test/unit/*/*_test.js"
```
stramel: You could shorten this line by using a more complex glob selector if you want,

```
"test:watch": "tsc-then -- mocha -c --ui tdd 'lib/test/@\\(integration\\|unit\\)/{,!\\(fixtures\\)/**/}/*_test.js'"
```

rictic: The gulp command does too much, it compiles which we're using tsc-then to do quickly and incrementally. 

Gulp also doesn't seem to do deduplication, so since both `test` and `test:integration` both depend on `build` it will end up compiling _twice_.

I often feel like I'm holding gulp wrong though, there's got to be a way to use it so that it doesn't do this… I would think.

rictic:
Blocked by https://github.com/Polymer/polymer-build/issues/297

package.json
stramel: I definitely thought `gulp test test:integration` would only run build once, maybe I'm wrong

rictic:
Linking in locally, it looks like https://github.com/Polymer/polymer-build/pull/314 is all that's needed for the build integration tests to pass. Once https://github.com/Polymer/polymer-build/pull/326 is merged and 3.0.0-pre.1 of build is released this may be ready to merge.

src/lint/lint.ts
```diff
@@ -322,22 +339,49 @@ async function fix(
...
       await applyEdits(edits, makeParseLoader(analyzer, analysis));
 
-  for (const [newPath, newContents] of editedFiles) {
-    await fs.writeFile(
-        path.join(config.root, newPath), newContents, {encoding: 'utf8'});
+  const pathToFileMap = new Map<string, string>();
+  for (const [url, newContents] of editedFiles) {
+    const conversionResult = urlLoader.getFilePath(url);
+    if (conversionResult.successful === false) {
+      console.log(
```
valdrinkoshi: should we use `logger` here and below?

package.json
```diff
@@ -57,10 +58,10 @@
...
     "mz": "^2.6.0",
     "plylog": "^0.4.0",
-    "polymer-analyzer": "^2.6.0",
-    "polymer-build": "^2.5.0",
-    "polymer-linter": "^2.2.0",
-    "polymer-project-config": "^3.10.0",
+    "polymer-analyzer": "^3.0.0-pre.14",
+    "polymer-build": "^3.0.0-pre.1",
+    "polymer-linter": "^3.0.0-pre.3",
+    "polymer-project-config": "^3.9.0",
```
valdrinkoshi: probably a merge issue, should keep 3.10.0 here

src/lint/lint.ts
rictic: +1 to this fatal error. The rest of the output shouldn't go through the logger I don't think, it should be logged directly to stdout as it's the primary purpose of running the command to log this stuff to stdout.

package.json
rictic: +1, done

valdrinkoshi:
LGTM :shipit:

PR #954, opened 2018-02-02
Updated CHANGELOG for 1.6.0 release and updated package-lock.json

usergenic:
## v1.6.0 [2-2-2018]
 - Added support for `exclude` option in `polymer.json` for the build command's `minify` and `compile` options for `css`, `js` and `html` files.
 - Added `--fix` option to `polymer lint`. When passed, some warnings with simple mechanical solutions will be fixed.
   - Also supports warnings which can be addressed with less-safe changes via
     an interactive prompt and the `--edits` flag. See `polymer lint --help` for
     more info.
 - Added `--watch` (shorthand `-w`) option to `polymer lint`. When passed, we will watch the filesystem for changes and rerun the linter immediately afterwards.
   - Also works with `--fix` to automatically fix and report warnings as you work!
 - `build` Added a CLI argument for setting the `basePath` option: `--base-path`.
 - Derives node version check from the package.json.
 - The polymer.json project config can now specify paths to exclude when minifying files and compiling JavaScript.  See https://github.com/Polymer/polymer-project-config/issues/50 for more.


abdonrd:
I see the `package-lock.json` removed.

stramel:
@abdonrd I am seeing the same.

FredKSchott:
Also for the CLI, we generate a shrinkwrap file for releases (lockfile alone doesn't lock dependencies for packages installed globally)

usergenic:
Ugh. I was using my Node 6 vm to verify it compiled on older version and I guess its npm was before package-lock.json.  Fixing.

usergenic:
Added back `package-lock.json`.

@FredKSchott I don't see npm-shrinkwrap.json in the repo.  Can you define the process of adding shrinkwrap to CLI?

usergenic:
Okay we are switching from package-lock.json to npm-shrinkwrap.json for CLI because trying to keep dependencies golden for users.

stramel:
Finally don't have to build it to use `--fix` 🎉

Issue #722, opened 2017-05-09
Support for Proxy


a1626:
### Description
Disclaimer: This might be similar to #201. 
Is there support in cli to proxy calls to api.github.com? If not, is there any other way to proxy same. Or is there any near future plan to add the same, as a lot of polymer users will be behind some sort of firewall. IMHO cloning the template manually and then working on it beats the purpose of cli.

### Versions & Environment
- Polymer CLI: 0.18.1
- node: 6.9.5
- Operating System: Sierra (10.12.3)

#### Actual Results
__Stack trace for shop template from terminal__: 
info:    Running template shop...
info:    Finding latest ^1.0.0 release of Polymer/shop
error:   Uncaught exception: getaddrinfo ENOTFOUND api.github.com api.github.com:443
error:   Error: getaddrinfo ENOTFOUND api.github.com api.github.com:443
    at __dirname.sendError (/usr/local/lib/node_modules/polymer-cli/node_modules/github/lib/index.js:838:19)
    at /usr/local/lib/node_modules/polymer-cli/node_modules/github/lib/index.js:849:29
    at callCallback (/usr/local/lib/node_modules/polymer-cli/node_modules/github/lib/index.js:713:17)
    at ClientRequest.<anonymous> (/usr/local/lib/node_modules/polymer-cli/node_modules/github/lib/index.js:791:17)
    at emitOne (events.js:96:13)
    at ClientRequest.emit (events.js:188:7)
    at TLSSocket.socketErrorListener (_http_client.js:310:9)
    at emitOne (events.js:96:13)
    at TLSSocket.emit (events.js:188:7)
    at connectErrorNT (net.js:1022:8)

__Result of api.github.com from browser__:
{
  "current_user_url": "https://api.github.com/user",
  "current_user_authorizations_html_url": "https://github.com/settings/connections/applications{/client_id}",
  "authorizations_url": "https://api.github.com/authorizations",
  "code_search_url": "https://api.github.com/search/code?q={query}{&page,per_page,sort,order}",
  "commit_search_url": "https://api.github.com/search/commits?q={query}{&page,per_page,sort,order}",
  "emails_url": "https://api.github.com/user/emails",
  "emojis_url": "https://api.github.com/emojis",
  "events_url": "https://api.github.com/events",
  "feeds_url": "https://api.github.com/feeds",
  "followers_url": "https://api.github.com/user/followers",
  "following_url": "https://api.github.com/user/following{/target}",
[...]

PR #955, opened 2018-02-03
Support to proxy github api calls (copy of #722)

a1626:


FredKSchott:
Looks like a ton of commits got pulled into this. Are you sure this was rebased on the latest master?

a1626:
Ya actually after rebasing a new version got released and i ended up pulling and merging from master again.

src/github/github.ts
```diff
@@ -51,6 +51,7 @@ export class GithubResponseError extends Error {
 export interface GithubOpts {
   owner: string;
   repo: string;
+  proxy: string;
```
FredKSchott: `proxy?: string`, since proxy is optional

src/github/github.ts
```diff
@@ -64,6 +65,7 @@ export class Github {
       .RequestAPI<request.Request, request.CoreOptions, request.RequiredUriUrl>;
   private _owner: string;
   private _repo: string;
+  private _proxy: string;
```
FredKSchott: `private _proxy?: string`

src/github/github.ts
```diff
@@ -77,6 +79,10 @@ export class Github {
     this._token = opts.githubToken || Github.tokenFromFile('token');
     this._owner = opts.owner;
     this._repo = opts.repo;
+    this._proxy = opts.proxy;
+    // TODO: this is a temporary solution and needs to be fixed once
+    // https://github.com/mikedeboer/node-github/issues/539 gets fixed
+    this._proxy ? process.env.HTTPS_PROXY = this._proxy : null;
```
FredKSchott: with the changes above, you shouldn't need this anymore. If you still do, can you explain why / what you're trying to do here? (https://github.com/octokit/rest.js/issues/539 is fixed btw)

src/init/github.ts
```diff
@@ -23,21 +23,23 @@ export interface GithubGeneratorOptions {
   owner: string;
   repo: string;
   semverRange?: string;
+  proxy: string;
```
FredKSchott: This is optional right? If so, `proxy?: string`

src/init/init.ts
```diff
@@ -193,6 +198,41 @@ function createSelectPrompt(env: YeomanEnvironment) {
   };
 }
 
+/**
+ * Checks and returns proxy settings from gitconfig. This function
+ * assumes that gitconfig will either be located locally (project's config file)
+ * or globally (user home)
+ */
+function getProxyConfig() {
+  let pathToConfig = './.git/config';
```
FredKSchott: Instead of using `let` and reassigning variables within a function scope, we prefer to create new, clean variables with `const`. Can you rename to `pathToLocalConfig`, `localMeta`, `userMeta`, etc. (fwiw, the first `pathToConfig` value is only used once and doesn't need its own variable)

FredKSchott: Also, use `path.join()` when joining paths for windows support.

FredKSchott (changes requested):
Thanks for resubmitting! Left some comments, also existing tests are failing :( Lets get those running and then talk new tests.

src/github/github.ts
a1626: It is still required. This change was not because of `cli`'s code but because of `github` module code. The problem was that the module did not had parameter `proxy` in it's `config` object because of which passing it as param to `GitHubApi` was not working. On checking the module's code i found that if `proxy` is not in `config` object they read it from `process.env`.
Even though the PR has been closed it looks like the project has also been migrated from `mikedeboer/node-github` to `octokit/rest.js`, which is not changed in `package.json`.

a1626:
Test cases are passing now, but what i noticed is that for `init_test.ts` even though test cases are passing cli is throwing errors.

image

justinfagnani:
Let's make sure to give PRs meaningful names. Naming them just after the issue number requires too much searching around.

aomarks:
Closing this pull request because this repo is now archived. If this PR is still active, please submit a new one at https://github.com/Polymer/tools.

PR #962, opened 2018-02-19
Fix es5 template literals

### LINKED ISSUE #950: Tagged template literals incorrectly transpiled to es5
Tagged template literals are transpiled incorrectly to es5. _templateObject variables are placed outside script closures and end up on the global scope. This causes multiple tagged template literals to override each other. 

I've investigated this, and it seems that babel is expecting to be in a module scope rather than a global scope. I am not sure if this is due to how babel is configured inside the polymer cli, or a bug within the babel transformer.

The same transpilation done by the typescript compiler provides valid output, it does not place variables outside closures.

This affects polyserve as well as polymer build. It happens on .js and .html files.

A minimal reproduction is like this:

Given two files:

html-file-a.html
```
<link rel="import" href="../bower_components/polymer/lib/utils/html-tag.html">
<link rel="import" href="./html-file-b.html">

<script>
  (function () {
    const a = window.Polymer.html`<div>A</div>`;
  })();
</script>
```

html-file-b.html
```
<script>
  (function () {
    const b = window.Polymer.html`<div>B</div>`;
  })();
</script>
```

And the following polymer.json:
```
{
  "entrypoint": "index.html",
  "shell": "src/html-file-a.html",
  "fragments": [
  ],
  "sources": [
    "src/**/*"
  ],
  "extraDependencies": [
    "bower_components/webcomponentsjs/*"
  ],
  "builds": [
    {
      "preset": "es5-bundled"
    }
  ]
}
```

Produces the following output:
```
<html><head></head>
<body>
  <div hidden="" by-polymer-bundler="">
    <script>... insertions by polymer omitted for brevity...</script>
    <script>var _templateObject = babelHelpers.taggedTemplateLiteral(["<div>B</div>"], ["<div>B</div>"]); (function () { window.Polymer.html(_templateObject) })();</script>
    <script>var _templateObject = babelHelpers.taggedTemplateLiteral(["<div>A</div>"], ["<div>A</div>"]); (function () { window.Polymer.html(_templateObject) })();</script>
  </div>
</body>
</html>
```

Versions:
Polymer CLI v1.5.7
Polymer Build v2.1.1
Polymer Bundler v3.1.1

tony19:
Babel's variable naming convention for tagged template literals is `"_templateObject<INDEX>"`, where `<INDEX>` is appended to subsequent names when there is more than one. For example, given 3 tagged template literals in a JavaScript file, Babel creates `"_templateObject"`, `"_templateObject2"`, and `"_templateObject3"`. (see [[1]](https://github.com/babel/babel/blob/454933e/packages/babel-core/src/transformation/file/index.js#L333) and [[2]](https://github.com/babel/babel/blob/4d164bd/packages/babel-traverse/src/scope/index.js#L231-L254))

`polymer-cli` feeds each script to Babel individually, which allows for a potential naming collision for the `_templateObject` variables when multiple scripts are bundled. The effect can be observed in this repro, where `Polymer.html` is used and bundling is enabled: https://github.com/LarsDenBakker/polymer-cli-tagged-template

This PR attempts to workaround the problem by ensuring all `_templateObject` variables are unique (by appending UUIDs), as suggested in [#950 comment](https://github.com/Polymer/polymer-cli/issues/950#issuecomment-363327475).

Fixes #950 


src/build/optimize-streams.ts
```diff
@@ -104,10 +104,24 @@ export class GenericOptimizeTransform extends Transform {
...
+   * Transforms given Babel-transpiled code so that the variable names
+   * of template literals are each unique
+   * @param code Transpiled code to evaluate
+   * @returns The code with all `_templateObject` variable
+   * names modified to be unique.
+   */
+  _replaceTemplateObjectNames(code: string): string {
+    // e.g., _templateObject -> _templateObject_200817b1154811e887be8b38cea68555
+    const uniqueId = uuid().replace(/-/g, '');
+    return code.replace(/_templateObject(?!_[A-Fa-f0-9]+)/g, `_templateObject_${uniqueId}`);
```
43081j: not that it makes any functional difference but you could probably just check a word boundary here (`_templateObject\b`)

tony19: Good call. Done

justinfagnani:
Awesome, thanks for this fix!

tony19:
@justinfagnani Ok, I've added the integration test.

LarsDenBakker:
Friendly bump for a review. This MR is highly anticipated :)

Issue #938, opened 2017-12-18
Add --npm flag


justinfagnani:
This global flag should feed into the build, lint, serve and test commands to set up the component directory and local package name correctly.

bicknellr:
#964 is merged, closing.

PR #964, opened 2018-02-23
Add `--npm` flag to build, lint, serve, and test commands.

bicknellr:
parent: https://github.com/Polymer/polymer-cli/issues/938
 - [x] https://github.com/Polymer/polymer-build/pull/289
 - [x] ~~Add `npm` flag documentation to `build`, `lint`, `serve`, and `test` commands~~.
 - [x] Add `npm` and `component-dir` flag documentation to global args.
- [ ] Tests

### DIFF
--- a/src/commands/build.ts
+++ b/src/commands/build.ts
@@ -25,6 +25,8 @@ import {PolymerProject} from 'polymer-build';
 import {applyBuildPreset, ProjectBuildOptions, ProjectConfig} from 'polymer-project-config';
 
 import * as buildLibTypeOnly from '../build/build';
+import {dashToCamelCase} from '../util';
+
 import {Command, CommandOptions} from './command';
 
 const logger = logging.getLogger('cli.command.build');
@@ -113,13 +115,15 @@ export class BuildCommand implements Command {
     {
       name: 'npm',
       type: Boolean,
-      description: 'Look for dependencies in "node_modules/".'
+      description: 'Sets npm mode: component directory is "node_modules/" ' +
+          'and the package name is read from package.json.'
+    },
+    {
+      name: 'component-dir',
+      type: String,
+      description: 'Sets the component directory.'
     },
   ];
-
-  private dashToCamelCase(text: string): string {
-    return text.replace(/-([a-z])/g, (v) => v[1].toUpperCase());
-  }
 
   /**
    * Converts command-line build arguments to the `ProjectBuildOptions` format
@@ -136,7 +140,7 @@ export class BuildCommand implements Command {
           (<any>buildOptions)[prefix] = (<any>buildOptions)[prefix] || {};
           (<any>buildOptions)[prefix][option] = options[buildOption];
         } else {
-          (<any>buildOptions)[this.dashToCamelCase(buildOption)] =
+          (<any>buildOptions)[dashToCamelCase(buildOption)] =
               options[buildOption];
         }
       }
--- a/src/commands/build.ts
+++ b/src/commands/build.ts
@@ -110,6 +110,11 @@ export class BuildCommand implements Command {
       type: String,
       description: 'Updates the <base> tag if found in the entrypoint document.'
     },
+    {
+      name: 'npm',
+      type: Boolean,
+      description: 'Look for dependencies in "node_modules/".'
+    },
   ];
 
   private dashToCamelCase(text: string): string {
--- a/src/commands/build.ts
+++ b/src/commands/build.ts
@@ -112,17 +112,6 @@ export class BuildCommand implements Command {
       type: String,
       description: 'Updates the <base> tag if found in the entrypoint document.'
     },
-    {
-      name: 'npm',
-      type: Boolean,
-      description: 'Sets npm mode: component directory is "node_modules/" ' +
-          'and the package name is read from package.json.'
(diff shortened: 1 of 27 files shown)


src/commands/test.ts
```diff
@@ -145,10 +145,20 @@ export class TestCommand implements Command {
     },
   ];
 
-  run(_options: CommandOptions, _config: ProjectConfig): Promise<void> {
+  run(_options: CommandOptions, config: ProjectConfig): Promise<void> {
```
bicknellr: I need to check if `_options` already been merged into `config`. If not, then there needs to be something below to overwrite settings from `config` with anything in `_options`.

src/commands/build.ts
```diff
@@ -110,6 +110,11 @@ export class BuildCommand implements Command {
       type: String,
       description: 'Updates the <base> tag if found in the entrypoint document.'
     },
+    {
+      name: 'npm',
+      type: Boolean,
+      description: 'Look for dependencies in "node_modules/".'
```
justinfagnani: polyserve is documented like this: Sets npm mode: component directory is "node_modules" and the package name is read from package.json

Since this also effects package name in `polymer serve`, I'd go with this description.

bicknellr: Not really sure if this is the right way to put this. `--npm` for `build` just sets `componentDir: "node_modules/"` but the idea is that if there were other configurable things that end up in polymer.json that are npm related, then it should set those too. Also, there's an implicit `--component-dir` flag, given that change to polymer-project-config. Maybe I should just document both of these for all the commands.

src/commands/serve.ts
```diff
@@ -70,7 +70,8 @@ export class ServeCommand implements Command {
       browser: options['browser'],
       openPath: options['open-path'],
       npm: options['npm'],
```
justinfagnani: this should pull from config as well, right?

bicknellr: yeah; will update

src/commands/build.ts
bicknellr: Ok, does anything need to read the package name in build?

bicknellr:
I think most of my switches for `options` vs `config` for `npm` and `componentDir` isn't really necessary anymore given the camel-casing thing that I added in 'polymer-cli.ts'.

bicknellr:
Yeah, those switches weren't necessary and were removed.

bicknellr:
The tests for `lint` and `build` are just "run over something trivial that shouldn't explode if the flags are read correctly". I don't have tests for `test` and `serve` yet. Although, I think `test` working probably implies that `serve` does also since `test` uses WCT, so I might be able to just add tests for `test` and call that good enough?

bicknellr:
Also, these tests only check that the cli flag works, not explicitly the polymer.json option. They should use the same path but maybe I should make separate fixtures to test that situation?

bicknellr:
@justinfagnani, are you sure I'm not meant to be committing npm-shrinkwrap.json? I don't think I'll be able to make this build in CI without it since it's already in here but has references to versions of polymer-build and polymer-project-config that aren't recent enough for this branch.

justinfagnani:
@bicknellr oh, if it's already checked in, then update it.

src/polymer-cli.ts
```diff
@@ -67,6 +68,17 @@ function parseCLIArgs(commandOptions: any): {[name: string]: string} {
   return parsedOptions;
 }
 
+/**
+ * Shallowly copies an object, converting keys from dash-case to camelCase.
+ */
+function objectDashToCamelCase(input: Object): Object {
+  const output = {};
+  for (const key of Object.keys(input)) {
+    (<any>output)[dashToCamelCase(key)] = (<any>input)[key];
```
justinfagnani: 1) when needing casts, prefer `as`: (output as any)[dashToCamelCase(key)] = (input as any)[key];
2) You shouldn't need the casts if you just type things as `any`

src/polymer-cli.ts
```diff
@@ -183,8 +195,11 @@ export class PolymerCli {
     const commandOptions = parseCLIArgs(commandOptionsRaw);
     logger.debug(`command options parsed from args:`, commandOptions);
 
-    const mergedConfigOptions =
-        Object.assign({}, this.defaultConfigOptions, commandOptions);
+    const mergedConfigOptions = Object.assign(
```
justinfagnani: Use Object spread syntax:

```ts
const mergedConfigOptions = {
  ...this.defaultConfigOptions,
  ...objectDashToCamelCase(commandOptions),
};
```

In fact, I think (but not 100% sure) object spread support iterables of entries:

```ts
const mergedConfigOptions = {
  ...this.defaultConfigOptions,
  ...Object.entries(commandOptions).map(([k, v]) => [dashToCamelCase(k), v]),
};
```

src/polymer-cli.ts
bicknellr: Ok, `input` and `output` are now typed as `any`.

src/polymer-cli.ts
bicknellr: I changed this to the first option because the second ends up spreading the indexes of the map output and adding the reduce seemed to make this line too verbose.

PR #966, opened 2018-03-07
Get latest shop parsing in compile/minify without errors

aomarks:
- Add babel syntax plugins for object-rest-spread and dynamic-import.
- Add babel transform plugin for object-rest-spread.
- Bump babel-preset-minify version and drop unused babili dependency.
- Make it more clear which optimize step warnings are coming from.
- Clang format everything.
- Analyzer still needs a bump, so that is still crashing (because of object rest spread).


src/build/optimize-streams.ts
```diff
@@ -116,9 +137,13 @@ class JSBabelTransform extends GenericOptimizeTransform {
  */
 export class JSDefaultCompileTransform extends JSBabelTransform {
   constructor() {
-    super({
+    super('babel-compile', {
       presets: [babelPresetES2015NoModules],
-      plugins: [externalHelpersPlugin],
+      plugins: [
+        externalHelpersPlugin,
+        babelObjectRestSpreadPlugin,
```
justinfagnani: did you look into babel-preset-env?

aomarks: just read about it. seems like we should switch but haven't tried it out yet. actually the preset plugins now warn you to move to it.

PR #967, opened 2018-03-08
New: Run custom builds from associated commands

stramel:

Wanted to get an okay before getting too far. What I have currently functions as expected.

@FredKSchott had brought up a while ago that it would be nice for someone to add an NPM script and be able to run that from the associated Polymer CLI command. For example,

**package.json**
```json
{
  "scripts": {
    "polymer:build": "gulp build"
  }
}
```

Then running `polymer build` in the console would result in the custom `gulp build` command being run instead of the default build process. If that NPM script is not present, it will run the default process.

I have set it up to where all commands except `help` follow the syntax `polymer:<command>`.


kito99:
I'd like to see the ability to bypass this if necessary. For example, I'd like for `polymer build` to be equivalent to this: `tsc && polymer build`. As-is, this would cause an infinite loop, so we'd need support for something like this: `tsc && polymer build -bypass-custom-commands`.

stramel:
@kito99 Ah, I see what you mean. I think we could easily add a `--bypass-custom-command`, `--ignore-custom-command`, or whatever naming the Polymer team would want for a flag to bypass.

stramel:
Ping @justinfagnani I can't remember what you wanted done here.

PR #968, opened 2018-03-09
New: Allow globs to be used for input arg

stramel:

Wanting to make sure this is the approach you guys want to take on this before adding tests.

Would fix https://github.com/Polymer/polymer-linter/issues/87


src/util.ts
```diff
@@ -35,6 +36,15 @@ function checkIsMinGW(): boolean {
   }
 }
 
+export async function resolveGlobs(globs: string[]) {
+  let paths: string[] = [];
+  for (const glob of globs) {
```
stramel: Had trouble getting a reduce to work for Array.

tony19: Isn't `map` more appropriate here than `reduce`?

```javascript
return globs.map(async (glob) => await globby(glob, {gitignore: true} as any));
```

stramel: Wouldn't that return an Array of Arrays?

tony19: Oops, you're right. Actually, I don't think either is needed, as `globby` takes a string array as the first argument, so you could do:

```javascript
return await globby(globs, {gitignore: true} as any);
```

stramel: Oh, you are correct 😅 will update when I get to my computer.

stramel:
Maybe it makes more sense to have this in analyzer instead? @usergenic wdyt?

src/util.ts
```diff
@@ -35,6 +36,14 @@ function checkIsMinGW(): boolean {
   }
 }
 
+/**
+ * Wrapper around globby that defaults the option to use gitignore
+ */
+export async function resolveGlobs(globs: string[]) {
+  // TODO(stramel): globby options typing is incorrect
+  return await globby(globs, {gitignore: true} as any);
```
FredKSchott: this gitignore feels a bit unexpected, I could easily see users confused why files weren't being matched. Can we remove it?

stramel: Yeah, we can and at that point, probably no need for the wrapper function?

FredKSchott: I could go either way on adding this sort of extra functionality to the analyzer. Agreed that it's useful, but it's also extra complexity/dependencies/maintenance/potential for bugs, for a feature that most people won't need.

Either way, this is really good, lets not block it here.

src/init/element/element.ts
```diff
@@ -62,7 +62,7 @@ export function createElementGenerator(templateName: string):
             if (!nameValidation.isValid) {
               this.log(`\n${nameValidation.message}\nPlease try again.`);
             } else if (nameValidation.message) {
-              this.log(''); // 'empty' log inserts a line break
+              this.log('');  // 'empty' log inserts a line break
```
FredKSchott: nit: only 1 space needed

stramel: Hmmm, this is what `npm run format` gave as output

FredKSchott: Really? I didn't think npm format touched spacing like that. In that case, this is fine :)

PR #970, opened 2018-03-13
Update to babel-preset-env to support async/await

Westbrook:

This was a pretty straightforward change which makes me think that there must be something that I'm missing here. If anyone has any insight on how this could be an issue, I'm happy to make further changes to support this.


justinfagnani:
We're starting to move this code to polymer-build. Can you open up a PR there?

PR #976, opened 2018-03-22
Speed up Travis builds from 13 minutes to 5 minutes

samuelli:
This change improves Travis build times on pull requests. It does this by:
 - caching `node_modules/`. `npm install` is still run and used, it just starts from the last builds cached `node_modules`.
 - uses parallel builds to run smoke tests and integration tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,15 @@ addons:
       - google-chrome
     packages:
       - google-chrome-stable
+cache:
+  directories:
+    - node_modules/
+env:
+  - TEST_SUITE=smoke-lint
+  - TEST_SUITE=integration
 language: node_js
 node_js:
   - "node"
   - "6"
 script:
-  - xvfb-run npm test
-  - xvfb-run npm run test:integration
+  - xvfb-run npm run test:$TEST_SUITE
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "build": "gulp build",
     "test": "gulp test lint",
     "test:smoke": "gulp test",
+    "test:smoke-lint": "gulp test lint",
     "test:integration": "gulp test:integration",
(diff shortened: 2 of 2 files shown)


rictic:
Hm. I'm really skeptical about caching node_modules, as `npm install` behaves differently and in unexpected ways with an existing node_modules directory.

TimvdLippe:
We have been using caching on polymer/polymer for a great while now without issues. I think this is a safe change.

samuelli:
> Hm. I'm really skeptical about caching node_modules, as npm install behaves differently and in unexpected ways with an existing node_modules directory.

It does, I haven't really had a problem and its a trade off I've usually been willing to take. Travis lets you clear the caches out pretty easily.

rictic:
Does polymer core have tight requirements on its node dependencies?

samuelli:
FWIW, on this repo, without caching its about ~45 seconds for `npm install` and ~12 seconds with caching.

rictic:
An example of a super subtle dependency issue that having hermetic, reproducible builds helped a lot with: https://github.com/Polymer/web-component-tester/issues/648

samuelli:
> An example of a super subtle dependency issue that having hermetic, reproducible builds helped a lot with: Polymer/web-component-tester#648

I don't understand how uncached Travis builds helped here. If anything it seems that this kind of upgrade would've been picked up on a cached Travis build before it affected any users.

rictic:
Background: The issue was that a release of wct-local was broken, and required corresponding changes in wct. wct-local doesn't have tests, and depends on testing in wct, and the breaking change was super subtle and not picked up in code review, so the release of wct-local went out.

IIRC we caught that failure in CLI's tests of `polymer test` because we got the latest version of wct-local and wct, which were incompatible. If we had gotten a cached version of wct-local (which would have still fulfilled our semver requirements) then CLI's tests would not have caught this issue.

TimvdLippe:
If I understand Travis caching correctly, it willjust copy the folder and
then do the check. This would mean that npm install will check and
potentially update the package (and thus catch such errors)

PR #984, opened 2018-03-29
Integrate AMD + module bundling into CLI.

### LINKED ISSUE #983: bundling and AMD transform probably don't work together
We currently run "optimize" before bundling in CLI, which means, if enabled, AMD transformation will happen before bundling. This will break Analyzer's understanding of bundling, so the two won't work together.

I think we'll need to do bundling first.

cc @usergenic

aomarks:
- Adds AMD transform step to build pipeline, and a new `--js-transform-modules-to-amd` flag. When bundling is happening, we do the AMD step in a separate pass, after bundling (fixes #983). Otherwise it happens in the original "optimize" pass.

- Fixes magic flag parsing logic, which mapped `--js-foo-bar-baz` to `js.foo` (taking only the first two parts). (I'm thinking we should write proper types for command-line-args that use the new mapped type feature in TS 2.8, and just map them through manually instead of doing this auto-mapping, going forward).

- Adds integration tests for a module based project in various configurations: unbundled-es, unbundled-amd, bundled-es, bundled-amd.


justinfagnani:
I think `--js-transform-es-modules-to-amd` should be named just `--js-transform-modules-to-amd`

Also, an option for import.meta would be great in this PR.

src/test/integration/run-command.ts
```diff
@@ -71,6 +71,9 @@ async function pipesToString(
   let str = '';
   const promises = [];
   for (const stream of [stdout, stderr]) {
+    if (!stream) {
```
rictic: Huh, when would this be the case?

aomarks: Don't know why, but if you set `silent=true`, it will be undefined and throw.

rictic: I guess the typings for node commands aren't super great about nulls, so this sort of defensiveness isn't unreasonable.

src/build/build.ts
```diff
@@ -13,14 +13,16 @@
...
 import * as path from 'path';
 import * as logging from 'plylog';
 import {dest} from 'vinyl-fs';
 
 import mergeStream = require('merge-stream');
 import {forkStream, PolymerProject, addServiceWorker, SWConfig, HtmlSplitter} from 'polymer-build';
 
-import {OptimizeOptions, getOptimizeStreams} from 'polymer-build';
+import {getOptimizeStreams} from 'polymer-build';
+import {JsTransform, HtmlTransform, matchesExt} from 'polymer-build/lib/optimize-streams';
```
rictic: Maybe we should re-export these from polymer-build

aomarks:
> I think --js-transform-es-modules-to-amd should be named just --js-transform-modules-to-amd

Done.

aomarks:
> I think --js-transform-es-modules-to-amd should be named just --js-transform-modules-to-amd

Done.

> Also, an option for import.meta would be great in this PR.

Done.

src/commands/build.ts
```diff
@@ -124,10 +135,12 @@ export class BuildCommand implements Command {
     const validBuildOptions = new Set(this.args.map(({name}) => name));
     for (const buildOption of Object.keys(options)) {
       if (validBuildOptions.has(buildOption)) {
-        const [prefix, option] = buildOption.split('-', 2);
+        const [prefix, ...rest] = buildOption.split('-');
         if (['css', 'html', 'js'].indexOf(prefix) !== -1) {
+          const option = dashToCamelCase(rest.join('-'));
           (<any>buildOptions)[prefix] = (<any>buildOptions)[prefix] || {};
```
rictic: I guess this is tricky because we can't represent functions over string literals well in the type system. But this is fragile and will fail silently in all sorts of ways.

aomarks: Yeah, I mentioned this in the PR description. I feel like we should just directly map flags into the config object rather than trying to magic the CLI (especially once we make cool mapped types for the cli lib).

src/build/build.ts
aomarks: I'm not all that excited about adding them to our public API surface.

PR #986, opened 2018-04-04
Fix missing gulp-if dependency.

aomarks:


aomarks:
> Do we need to update the shrinkwrap?
> 
> I assumed that running npm install --save newdep would update it if it needed to. Not the case?

I ran `npm install` afterwards, and it updated the shrinkwrap. That seems like a bug.

aomarks:
PTAL.

PR #989, opened 2018-04-05
Do minify and other things after bundling instead of before.

aomarks:


src/test/integration/build_test.ts
```diff
@@ -201,8 +201,8 @@ function copyDir(fromDir: string, toDir: string) {
 
 function assertDirsEqual(actual: string, expected: string, basedir = actual) {
   // uncomment to update the golden:
-  // fsExtra.emptyDirSync(expected);
-  // fsExtra.copySync(actual, expected);
+  fsExtra.emptyDirSync(expected);
+  fsExtra.copySync(actual, expected);
```
rictic: re-comment?

Running in golden-update mode should fail the tests, as otherwise it would be easy to accidentally reduce the tests to no-ops

src/test/integration/fixtures/build-modules/expected/bundled-amd-minified/index.html
```diff
@@ -0,0 +1 @@
...t=dataMain,cfg.baseUrl||-1!==mainScript.indexOf("!")||(src=mainScript.split("/"),mainScript=src.pop(),subPath=src.length?src.join("/")+"/":"./",cfg.baseUrl=subPath),mainScript=mainScript.replace(jsSuffixRegExp,""),req.jsExtRegExp.test(mainScript)&&(mainScript=dataMain),cfg.deps=cfg.deps?cfg.deps.concat(mainScript):[mainScript],!0}),define=function(e,r,i){var n,t;"string"!==typeof e&&(i=r,r=e,e=null),isArray(r)||(i=r,r=null),!r&&isFunction(i)&&(r=[],i.length&&(i.toString().replace(commentRegExp,commentReplace).replace(cjsRequireRegExp,function(e,i){r.push(i)}),r=(1===i.length?["require"]:["require","exports","module"]).concat(r))),useInteractive&&(n=currentlyAddingScript||getInteractiveScript(),n&&(!e&&(e=n.getAttribute("data-requiremodule")),t=contexts[n.getAttribute("data-requirecontext")])),t?(t.defQueue.push([e,r,i]),t.defQueueMap[e]=!0):globalDefQueue.push([e,r,i])},define.amd={jQuery:!0},req.exec=function(text){return eval(text)},req(cfg)}})(this,"undefined"===typeof setTimeout?void 0:setTimeout);</script><script>define('polymer-build-generated-module-0', ['shared_bundle_1.js']);</script><script>require(['polymer-build-generated-module-0']);</script></head><body></body></html>
```
rictic: Can we use the code from the build test infra to replace common scripts with comments? This is super hard to read/review without it

src/build/build.ts
```diff
@@ -117,25 +94,29 @@ export async function build(
...
-        gulpif(matchesExt('.html'), new HtmlTransform({
-                 entrypointPath: polymerProject.config.entrypoint,
-                 js: {transformModulesToAmd: true},
-                 rootDir: polymerProject.config.root,
-               })),
-      ]);
-    }
   }
 
+  const htmlSplitter = new HtmlSplitter();
```
rictic: Do we have a test to ensure that we're inserting the ES5 shim properly?

aomarks: The golden below shows that we did not change how we inject the es5 adapter script tag (the HTML is just now better minimized). As for whether we compile it or not, this is tested here: https://github.com/Polymer/polymer-build/blob/master/src/test/optimize-streams_test.ts#L74.

src/test/integration/fixtures/build-with-preset/expected/es5-bundled/shared_bundle_1.html
```diff
@@ -1 +1 @@
-<template>I'm baz.html!</template>
\ No newline at end of file
+<html><head><template>I'm baz.html!</template></head><body></body></html>
```
aomarks: Not sure why these phantom dom5 tags have re-appeared, since I thought we were quite careful to remove them both in bundler and in the build HTML transform. From offline conversation, I think we're going to leave these for the upcoming pre release and file an issue to investigate.

src/test/integration/fixtures/build-modules/expected/bundled-amd-minified/index.html
aomarks: I would like to, agree these are very hard to read. It's a little more work than I originally thought when we talked, because it makes re-generating the goldens more complicated (which I am doing by just dumping out the whole dir), since I will need to transform the files in the temp dir or copy them one by one, rather than just copying the dirs over wholesale. Not a huge deal, but ok if I file an issue for now?

(I also removed this particular new test case anyway, because it wasn't actually testing anything already covered by other ones.)

src/test/integration/fixtures/build-with-preset/expected/es5-bundled/shared_bundle_1.html
aomarks: https://github.com/Polymer/polymer-cli/issues/991

